Issue #515, opened 2026-07-05
bug(factory-health): factory artifact red-gate-log-S-12.08.md tracked at repo ROOT on develop — outside .factory/, invisible to path-prefix leak checks


arcavenai:
## Summary

A factory artifact — `red-gate-log-S-12.08.md` — is tracked at the **repository root** on `develop`. It is a red-gate log (frontmatter: `document_type: red-gate-log`, `story_id: S-12.08`, `cycle: v1.0-feature-engine-discipline-pass-1`, `status: RED_GATE_VERIFIED`) that entered product history via a normal story PR and still ships in every clone of the code repo today.

This is the same theme as #341 (factory artifact tracked on the product branch) but a **different mechanism and a blind spot in #341's proposed fix**: the file lives at the root, not under `.factory/`, so

- the `.gitignore` `.factory/` rule never applied to it at any point, and
- the `git ls-files '.factory/*'` leak check proposed in #341 will never see it.

## Evidence

- `git ls-files` on `develop` includes `red-gate-log-S-12.08.md` at the root (verified on a fresh clone at rc.22).
- Introduced by commit `99d24315` — "[S-12.08] Convergence hook wave_context migration — closes F-P2-001 (CRITICAL PATH TERMINUS) (v1.2) (#123)" (2026-05-10).
- `git ls-tree -r --name-only factory-artifacts | grep red-gate-log-S-12.08` → no match. Unlike #341's split-brain case, this artifact exists **only** in product history — the factory side has no copy, so the factory's own record of that gate is the product repo's root file.

## Impact

- A process artifact rides along in every checkout, release tarball, and code search of the product repo.
- Because it's not split-brain (no live counterpart on `factory-artifacts`), nothing ever diverges or shows as modified — so it's *quieter* than #341 and will never self-surface.
- Precedent risk: story PRs can deposit red-gate/decision logs anywhere in the tree; path-prefix checks won't catch them.

## Proposed fix

1. Remove/relocate the file: `git rm --cached red-gate-log-S-12.08.md` on `develop` (and, if the factory wants the record, commit it to its proper home under `.factory/cycles/v1.0-feature-engine-discipline-pass-1/...` on `factory-artifacts`).
[...]

Issue #510, opened 2026-07-05
External code contributions from ArcavenAE — proposed flow for changes with factory-artifact implications


arcavenai:
Hi @DrBothen @Zious11 — following the issue reports we've been filing (recently
#497–#504 and the architectural-mismatch family #410–#417), we'd like to
start offering fixes and small features as PRs, not just reports. We've
forked to `ArcavenAE/vsdd-factory` (contribution-only fork — no divergent
development) and can run your full gate locally (fmt/clippy/test, both bats
suites, semgrep) before offering anything.

Code-only PRs are straightforward: `feature/*` branch → PR to `develop`,
conventional-commit title, green CI, squash-merge — per CONTRIBUTING.md.
We'll start there.

The open question is **changes with factory-artifact implications.** Since
`.factory/` mounts from `factory-artifacts`, a PR to `develop` can't carry
artifact updates, and your factory writes that branch continuously
(STATE.md especially).

**Our proposal: each code PR links an issue carrying the structured
artifact content** (spec deltas, BC/VP suggestions, story framing) **for
your factory to apply/crystallize on your side.** Your factory keeps sole
authorship of `factory-artifacts`; you never take external writes to the
branch your automation mutates daily; nothing we offer collides with
STATE.md.

We reached that by simulating a wave of ~5 parallel contributions before
asking (happy to share the model + numbers as a reference issue if
useful). The alternatives came out clearly worse:

- **Paired PRs** (code → `develop` + artifacts → `factory-artifacts`)
  averaged ~2 rebases per artifact PR against your factory's commit
  cadence, doubled your review load per feature, and a code-PR rejection
  after its artifact merged leaves phantom specs to revert. If you'd
  rather receive artifacts as PRs anyway, we'd do it with discipline: at
  most one artifact PR open at a time, artifact merged before its code PR,
  STATE.md untouched unless you want us registering there.
- **Code-only PRs with artifact implications described in prose** is the
  lowest-ceremony fallback, but leaves you reconstructing artifact intent
  from PR bodies and opens a spec/code divergence window until your
  pipeline regenerates.

Two mechanics to make the linked-issue flow dependable:

1. **Sequencing.** Where a code change depends on new artifacts (its
   hooks/tests reference them, and CI mounts your `factory-artifacts`),
   we'll file the artifact issue first and keep the code PR in **draft**
   until the content lands — no review time spent on a PR that CI must
   fail. One ask: when your factory applies content from one of our
[...]

Issue #341, opened 2026-06-29
bug(factory-health + repo-initialization): factory artifact tracked on product branch — .gitignore added without git rm --cached leaves split-brain copy, undetected


arcavenai:
## Summary

A factory artifact (`.factory/cycles/v1.0.0-greenfield/implementation/red-gate-log.md`) is tracked on the PRODUCT branch (`develop`) even though `.factory/` is listed in `.gitignore`. This produces a split-brain: the same path is tracked on BOTH `develop` (stale, 182 lines) and `factory-artifacts` (live, 317 lines — the copy agents actually write to). The `develop`-tracked copy shows as perpetually "modified" in the main worktree, blocked a clean fast-forward merge, and was accidentally swept (+74 lines) into an unrelated product PR merge.

`.gitignore` rules never untrack an already-tracked path. So adding `.factory/` to `.gitignore` without a paired `git rm --cached` is a silent no-op for any file already committed — and there is no health check that detects the leak.

## Environment

vsdd-factory plugin 1.0.0-rc.21, Claude Code, greenfield+wave mode. `.factory/` mounted as a git worktree on the `factory-artifacts` orphan branch; product code on `develop` (gitflow). Repo: private.

## Root cause (reconstructed from git history)

1. An early feature PR (commit `aeb442d`, "feat(S-W3.04): full daemon assembly") committed `red-gate-log.md` into `.factory/cycles/...` on `develop` — BEFORE any ignore rule existed for `.factory/`. A factory artifact entered product history through a normal feature commit. (This also points to a gap: code-delivery PRs should never include `.factory/` paths; a pre-PR guard could have caught it.)
2. A later cleanup PR (commit `36c5e98`, "chore: remove leftover kos process scaffolding") ADDED `.factory/` (and `.factory`) to `.gitignore` — but did NOT run `git rm --cached` on the already-tracked file. Git ignore rules do not untrack already-tracked paths, so it remained tracked.
3. Subsequent factory writes updated the `factory-artifacts` copy; the `develop` copy stayed frozen and divergent → split-brain + perpetual dirty status.

## Evidence

- `git ls-files '.factory/*'` on `develop` returns exactly one path: `.factory/cycles/v1.0.0-greenfield/implementation/red-gate-log.md`
- Same path is tracked on `factory-artifacts` too; contents differ (`develop` 182 lines vs `factory-artifacts` 317 lines).
[...]

PR #524, opened 2026-07-06
fix(factory-health): remove factory artifact red-gate-log-S-12.08.md from repo root

arcavenai:
## What

Removes `red-gate-log-S-12.08.md` from the repository root on `develop` — a factory artifact (frontmatter `document_type: red-gate-log`, `story_id: S-12.08`) that entered product history via PR #123 and ships in every clone.

Refs: #515

## Why

Per the analysis in #515: the file lives outside `.factory/`, so the `.gitignore` rule never applied to it and the `git ls-files '.factory/*'` leak check proposed in #341 can never see it. It has no counterpart on `factory-artifacts` and nothing in the code tree references it (verified: `grep -rn "red-gate-log-S-12.08" .` excluding `.git`/`.factory` → no matches).

This PR is deliberately **code-side removal only** (Class 0 in the terms of #510 — no artifact references, no `.factory` dependency). The content remains recoverable from git history (`99d24315`) if the factory wants to relocate the record to its proper home under `.factory/cycles/v1.0-feature-engine-discipline-pass-1/` on `factory-artifacts` — happy to route that content via an artifact issue per the #510 proposal once there's a ruling there.

Governance note: no BC/spec surface — this is factory hygiene restoring the artifact/product branch separation, not a behavior change. If you'd like story ceremony on external PRs (asked in #510), glad to adopt it going forward.

## Test evidence (local, at `f5242bef` + this change, rust 1.95.0)

```
cargo fmt --check --all                                  → clean
cargo clippy --workspace --all-targets -- -D warnings    → clean (Finished, no warnings)
cargo test --workspace --all-targets                     → 179 result lines, 0 failures
bats plugins/vsdd-factory/tests/hooks.bats               → 45/45 ok
bats plugins/vsdd-factory/tests/bin.bats                 → 13/13 ok
semgrep ci --config=auto (diff vs develop)               → Findings: 0 (0 blocking), 1074 rules
```

Commit is signed; diff is a single file deletion (67 lines).


Zious11:
Wouldnt it be best to move this into the .factory branch rather then delete it?

arcaven:
“The content remains recoverable from git history (99d24315) if the factory wants to relocate the record to its proper home under .factory/cycles/v1.0-feature-engine-discipline-pass-1/ on factory-artifacts — happy to route that content via an artifact issue per the https://github.com/drbothen/vsdd-factory/issues/510 proposal once there's a ruling there.”

there is no .factory branch, and we need to address how/if we will create PRs (it would be a second PR, we can’t to it in one step) to merge this to the orphan-branch factory-artifacts where it was supposed to go. 

This is a minor issue, it’s missing there, but the slightly more interesting risk is that future sessions may copy writing these files to the default branch. 

either way, it does not belong here, under the orphaned branch strategy, and deleting it is safe, and adding it to factory-artifacts must be done with a second PR.

PR #691, opened 2026-07-19
fix(test-writer): correct Go nil Stdin guidance — nil reads from null device, os.Stdin wiring is the trap

Zious11:
## Summary

Corrects two factual imprecisions in the subprocess-stdin discipline bullet introduced by PR #531 (dc110e03).

**Commit 1 — Go correction (b1ff2457):** PR #531 said \`cmd.Stdin = nil\` INHERITS — that is the trap. This is wrong. Per the Go \`os/exec\` documentation, a nil \`Cmd.Stdin\` means the child reads from the null device (\`os.DevNull\`); nil is the safe default. The actual trap is the opposite: explicitly wiring \`cmd.Stdin = os.Stdin\`, which causes the subprocess to compete with the test runner for terminal input.

**Commit 2 — Node precision (1aafa74b):** The original text implied the Node hazard was stdin inheritance. The actual hazard is that Node's default \`stdio\` option is \`'pipe'\` — a \`'pipe'\` stdin hangs a stdin-reading child process if the pipe is never written to or closed. The fix (\`stdio: ['ignore', ...]\`) is the same; this commit makes the diagnosis accurate so readers understand *why* the fix is needed.

The Rust guidance is also clarified: \`spawn\`/\`status\` inherit stdin by default, so \`.stdin(Stdio::null())\` is the explicit safe form.

No logic changes — doc-only corrections to \`plugins/vsdd-factory/agents/test-writer.md\`. References PR #531.

## Files changed

- \`plugins/vsdd-factory/agents/test-writer.md\` — subprocess-stdin discipline bullet, two clauses corrected

## Test plan

- [ ] Doc-only change; no runtime surface to drive
- [ ] Go clause: nil Cmd.Stdin reads from null device per \`os/exec\` docs
- [ ] Node clause: default \`'pipe'\` hangs unclosed; \`'ignore'\` is the fix

### DIFF
--- a/plugins/vsdd-factory/agents/test-writer.md
+++ b/plugins/vsdd-factory/agents/test-writer.md
@@ -91,13 +91,14 @@ implementation code exists. You enforce the Red Gate.
 - For any test that spawns a subprocess capable of interactive input (a
   shell, a REPL, a TUI, or any command that may prompt), ALWAYS set stdin
   explicitly to the null/closed equivalent and set a per-command timeout.
-  NEVER inherit stdin from the test runner: Rust
-  `.stdin(Stdio::null())`; Node `stdio: ['ignore', ...]`; Go leave
-  `cmd.Stdin` nil (nil already reads from the null device — the trap is
-  explicitly wiring `cmd.Stdin = os.Stdin`). Inherited stdin is
-  invisible on CI (no TTY, stdin closed) but hangs the suite locally in
-  a terminal, and the hang is routinely misdiagnosed as a product
-  regression.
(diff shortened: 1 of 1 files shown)


arcavenai:
Thanks @Zious11 for the precise correction — I had the Go trap exactly backwards (nil `Stdin` already reads from the null device; the real hazard is wiring `os.Stdin`), and the Node hazard misdiagnosed (unclosed `'pipe'`, not inheritance). Rust was the case we actually observed; the other two were extrapolated without verification. Lesson taken on our side: per-language behavior claims in guidance PRs now get checked against the docs or a minimal repro before we file.

Issue #637, opened 2026-07-13
bug(hooks): validate-input-hash + bash compute-input-hash use $(cat) command substitution (strips trailing newlines) — hashes diverge from raw-byte readers and the PostToolUse hook hard-blocks (exit 2) on false-positive drift


Zious11:
## Summary

The plugin's \`validate-input-hash\` PostToolUse hook and the bash \`bin/compute-input-hash\` it calls both compute the input hash using bash command substitution: \`CONCAT="${CONCAT}$(cat "$RESOLVED")"\`. Bash \`$()\` command substitution strips **all trailing newlines** from each file's content before concatenation, so the hash is computed over trailing-newline-stripped bytes. Any raw-byte reader (e.g., Python \`open(f, "rb").read()\`, or \`md5sum <file\`) is sensitive to trailing newlines and produces a different digest. The hook then hard-blocks (exit 2 via \`block_pre\`) on the resulting false-positive drift, treating an algorithm discrepancy as if the story's inputs had genuinely changed.

## Evidence

- **Hook registration:** \`hooks-registry.toml:293,302\` registers \`validate-input-hash\` → \`hooks/validate-input-hash.sh\`.
- **Hard-block path:** \`hooks/validate-input-hash.sh:76,107–114\` — on any hash mismatch the hook emits "input-hash drift ... stored X != computed Y" and exits with \`block_pre\` (exit 2), which halts the PostToolUse pipeline.
- **Lossy idiom:** \`bin/compute-input-hash:348\` — \`CONCAT="${CONCAT}$(cat "$RESOLVED")"\`; also \`:334\`. The \`$(cat file)\` subshell strips all trailing newlines from each file's content before the string is appended to \`CONCAT\`. The final \`md5sum\`/\`md5\` call at \`:386–390\` therefore operates over bytes with trailing newlines stripped.
- **Root cause (external confirmation):** bash \`$()\` command substitution strips all trailing newlines by POSIX spec. An MD5 of \`$(cat file)\` is insensitive to trailing newlines; \`md5sum file\` (reading raw bytes) is not. These two approaches produce different digests for any file whose content ends with one or more newlines — which is the normal case for text files.
- **Concrete divergences documented in wirerust (PG-HASH-HOOK-DIVERGENCE, wave-71):**
  - STORY-156: Python canonical \`ce96d86\` vs hook \`7b7dc6b\`
  - STORY-150: Python canonical \`c5acbe4\` vs hook \`26416e1\`
[...]

Issue #623, opened 2026-07-12
compute-input-hash --update silently no-ops when the target file has no pre-existing input-hash field — reports success, writes nothing


arcavenai:
## Summary

`compute-input-hash --update` silently no-ops when the target file has no pre-existing `input-hash` field: it reports success but performs only in-place substitution on an existing field, so a file that never had the field never gains it. There is no error, warning, or nonzero exit — the tool's own success report is indistinguishable between "field updated" and "field never existed and still doesn't."

## Reproduction

1. Run `compute-input-hash --update` against a target artifact whose frontmatter has no `input-hash:` key at all (as opposed to an empty or stale one).
2. The command exits successfully and prints its normal success output.
3. Inspecting the file afterward shows no `input-hash` field was added — the frontmatter is byte-identical to before the run except for whatever else was already there.
4. The gap was only discovered by grepping the file for the field after the fact; the agent then added the field manually to work around it.

## Why this matters

This is the same "reports success while writing nothing" shape as #599 (agent-authored lint gates that silently discard file args and under-scan while printing a clean "0 violations" summary), but at a different tool tier — #599 is about agent-authored gate scripts, this is about a shipped plugin binary that pipeline agents trust as ground truth for drift tracking. A tool that can report success on a no-op invites exactly the vacuous-verification chain #599 describes: any downstream step that trusts "`--update` succeeded, therefore the field is now populated" is silently wrong for every file that started without the field.

This is also a sibling defect to #314 in the same binary (`compute-input-hash`) — #314 is about the hash VALUE being wrong (frontmatter included in the hash computation, causing spurious drift); this is about the WRITE itself silently failing to happen for a specific starting condition (field absent). Independent bugs, same tool.

## Suggested fix

`--update` should either (a) always add the field if absent (find-or-create instead of find-and-replace), or (b) if find-and-replace is intentional, fail loudly (nonzero exit, explicit stderr message) when no existing field is found to update, rather than exiting 0. Silent success on a no-op should not be possible for a tool whose entire purpose is bookkeeping-field maintenance.

## Environment

- Helper: `bin/compute-input-hash` (`--scan`, `--check`, `--update`) — same binary as #314.

## Cross-refs

- #314 — sibling defect, same tool, different mechanism (wrong hash value vs. silent no-write).
- #599 — same silent-success-on-no-op pattern family, different tool tier (shipped plugin bin vs. agent-authored lint gate).

PR #715, opened 2026-07-20
fix(bin): preserve trailing newlines in compute-input-hash accumulation

arcavenai:
## Why

`compute-input-hash` accumulates each input file's bytes with `CONCAT="${CONCAT}$(cat "$RESOLVED")"`. Bash command substitution strips *all* trailing newlines from `$(cat file)`, so the hash is computed over newline-stripped bytes — but every raw-byte reader (`md5sum <file`, Python `open(f,"rb").read()`) hashes the real bytes. For any text file ending in a newline (the normal case), the two disagree. That divergence is exactly what makes the `validate-input-hash` PostToolUse hook hard-block (exit 2) on false-positive drift, halting the pipeline on artifacts whose inputs never actually changed. This restores byte fidelity so the hook's block can be trusted again, matching the minimal fix laid out in #637.

## What changed

- `plugins/vsdd-factory/bin/compute-input-hash`: replaced the two `CONCAT="${CONCAT}$(cat …)"` accumulations (single-file and glob-expansion paths) with `cat "$file" >> "$HASH_INPUT"`, where `$HASH_INPUT` is a `mktemp` file cleaned up by an `EXIT` trap. Redirection does not go through a subshell, so all bytes — including trailing newlines — are preserved. The final MD5 now reads from that temp file (`md5sum < "$HASH_INPUT"` / `md5 < "$HASH_INPUT"`) instead of `echo -n "$CONCAT"`. The empty-input guard changed from `[[ -z "$CONCAT" ]]` to `[[ ! -s "$HASH_INPUT" ]]` (temp file has zero size when no inputs resolved), preserving the existing `--resolve` / `--check` / default behavior for the no-input case.
- `plugins/vsdd-factory/tests/input-hash.bats`: two regression tests asserting the tool's hash equals a byte-accurate reference (`md5sum` over the raw file bytes / raw concatenation), one single-input and one multi-input, both using files with trailing newlines.

Scope is deliberately the bash `bin/compute-input-hash` accumulation only (issue proposal step 1). The hook calls this same binary, so fixing the binary fixes the hook's comparison hash without touching hook logic.

## Test evidence

Empirical red (unfixed binary), captured against the pristine `upstream/develop` copy of the script:

```
=== tool-computed hash (current buggy $(cat)) ===
TOOL_HASH=8af1b23
=== byte-accurate reference (md5 of raw file bytes) ===
REF_HASH=a9a7d79
cat-substitution bytes:       24
raw file bytes:               27
=== VERDICT ===
RED: tool hash (8af1b23) != byte-accurate reference (a9a7d79) — trailing newlines stripped
```

The two new bats tests, run against the unfixed script, fail as expected:

```
1..2
not ok 1 compute-input-hash: hash matches byte-accurate reference for input with trailing newlines (#637)
#   `[ "$tool_hash" = "$ref_hash" ]' failed
not ok 2 compute-input-hash: multi-input hash matches byte-accurate concatenation reference (#637)
#   `[ "$tool_hash" = "$ref_hash" ]' failed
```

Green after the fix — the same scenario now matches the byte-accurate reference for single and multi-input:

```
=== tool-computed hash (AFTER fix) ===
TOOL_HASH=a9a7d79
REF_HASH=a9a7d79
=== VERDICT ===
GREEN: tool hash (a9a7d79) == byte-accurate reference (a9a7d79) — trailing newlines preserved
multi tool=ddc11c6  ref=ddc11c6
GREEN multi
```

Full suite green (40/40, was 38 before the two new tests):

```
ok 37 compute-input-hash: repo-root-relative input (plugins/…) resolves and contributes to hash
ok 38 compute-input-hash: repo-root-relative input changes hash when SKILL.md changes
ok 39 compute-input-hash: truly-missing repo-root-relative path surfaces MISSING, not silently skipped
ok 40 compute-input-hash: path-traversal input (..) is rejected as MISSING, not resolved
```

Pairs with the sibling PR fixing #623 — independent hunks in the same file; either merges cleanly first.

## Migration note (artifact re-hash at merge)

This re-bases the tool's output for every trailing-newline input: existing
artifacts on `factory-artifacts` carry stored input-hashes computed under the
old accumulation (~2.3k files), and each would report drift from
`validate-input-hash` on its next write until re-hashed. A one-pass
`compute-input-hash --scan .factory --update` sweep at merge time clears the
backlog in a single commit; without it the re-hashes trickle in per-file as
writes occur.

Refs: #637

### DIFF
--- a/plugins/vsdd-factory/bin/compute-input-hash
+++ b/plugins/vsdd-factory/bin/compute-input-hash
@@ -215,7 +215,13 @@ case "$FILE" in
     ;;
 esac
 
-CONCAT=""
+# Accumulate raw input bytes into a temp file rather than a shell variable.
+# Command substitution ($(cat file)) strips all trailing newlines, so the hash
+# diverges from any raw-byte reader (md5sum <file, Python rb.read()) — see #637.
+# Appending with `cat file >> tmpfile` does not go through a subshell and
+# preserves every byte, including trailing newlines.
+HASH_INPUT=$(mktemp)
+trap 'rm -f "$HASH_INPUT"' EXIT
 MISSING=0
 MISSING_FILES=""
 
@@ -331,7 +337,7 @@ while IFS= read -r input_file; do
     fi
     while IFS= read -r expanded_file; do
       [[ -z "$expanded_file" ]] && continue
-      CONCAT="${CONCAT}$(cat "$expanded_file")"
+      cat "$expanded_file" >> "$HASH_INPUT"
     done <<< "$EXPANDED"
     continue
   fi
@@ -345,10 +351,10 @@ while IFS= read -r input_file; do
     continue
   fi
 
-  CONCAT="${CONCAT}$(cat "$RESOLVED")"
+  cat "$RESOLVED" >> "$HASH_INPUT"
 done <<< "$INPUTS"
 
-if [[ -z "$CONCAT" ]]; then
+if [[ ! -s "$HASH_INPUT" ]]; then
   if [[ "$MODE" == "--resolve" ]]; then
     echo "compute-input-hash: $(basename "$FILE"): $MISSING MISSING — $MISSING_FILES" >&2
     exit 1
@@ -383,11 +389,13 @@ if [[ "$MISSING" -gt 0 ]]; then
 fi
 
 # --- Compute short MD5 hash (7 chars) ---
+# Hash the accumulated bytes directly from the temp file so trailing newlines
+# are preserved and the digest matches any raw-byte reader (md5sum <file) — #637.
 if command -v md5sum &>/dev/null; then
-  HASH=$(echo -n "$CONCAT" | md5sum | cut -c1-7)
+  HASH=$(md5sum < "$HASH_INPUT" | cut -c1-7)
 elif command -v md5 &>/dev/null; then
   # macOS
-  HASH=$(echo -n "$CONCAT" | md5 | cut -c1-7)
+  HASH=$(md5 < "$HASH_INPUT" | cut -c1-7)
 else
   die "neither md5sum nor md5 available"
 fi
--- a/plugins/vsdd-factory/tests/input-hash.bats
+++ b/plugins/vsdd-factory/tests/input-hash.bats
@@ -295,6 +295,61 @@ EOF
   [[ "$output" == *"nonexistent.md"* ]]
 }
 
+# ===== regression: trailing-newline byte fidelity (#637) =====
+
+@test "compute-input-hash: hash matches byte-accurate reference for input with trailing newlines (#637)" {
+  # Regression for the $(cat file) command-substitution bug: bash strips ALL
+  # trailing newlines from command-substitution output, so the computed hash
+  # diverged from any raw-byte reader (md5sum <file, Python rb.read()). Files
+  # ending in one or more newlines (the normal case for text) MUST hash to the
+  # same digest the byte-accurate reference produces.
+  printf '# Product Brief\nLine two\n\n\n' > "$WORK/.factory/specs/product-brief.md"
+  cat > "$WORK/.factory/specs/prd.md" << 'EOF'
+---
+inputs: [product-brief.md]
+input-hash: "[md5]"
+---
+EOF
+
+  tool_hash=$("$BIN" "$WORK/.factory/specs/prd.md")
+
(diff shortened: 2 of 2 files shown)


arcaven:
Nicely scoped — this matches #637's step-1 fix, and the two regression tests reproduce the divergence cleanly (verified the digests locally: `8af1b23` → `a9a7d79` on a trailing-newline file). One line may be worth adding to the body: since this re-bases the tool's output for every trailing-newline input, existing artifacts on `factory-artifacts` (~2.3k carry stored input-hashes computed under the old accumulation) would each report drift from `validate-input-hash` on their next write until re-hashed — a `compute-input-hash --scan --update` sweep at merge time would clear that in one pass.

arcavenai:
Added to the body as a migration note: ~2.3k stored input-hashes computed under the old accumulation would each report drift on next write, and a one-pass `compute-input-hash --scan .factory --update` at merge time clears the backlog in a single commit.

Zious11 (approved):
# PR Review — #715 fix(bin): preserve trailing newlines in compute-input-hash accumulation

**Verdict:** APPROVE
**Date:** 2026-07-22
**Closes:** #637

---

## Summary

This PR fixes a byte-fidelity bug in `plugins/vsdd-factory/bin/compute-input-hash`. The accumulation loop used `CONCAT="${CONCAT}$(cat file)"` — Bash command substitution strips all trailing newlines, causing the computed hash to diverge from any raw-byte reader (`md5sum <file`, Python `open(f,"rb").read()`). The pipeline's `validate-input-hash` hook then hard-blocked on false-positive drift for any text file ending in a newline (the normal case), making the hook's block untrustworthy.

---

## Changes reviewed

**`plugins/vsdd-factory/bin/compute-input-hash`** — 4 hunks

- Hunk 1 (line 215): Introduces `HASH_INPUT=$(mktemp)` + `trap 'rm -f "$HASH_INPUT"' EXIT`. Replaces the shell-variable accumulation with a temp file. The EXIT trap ensures cleanup on error paths.
- Hunk 2 (line 337): Glob-expansion path — `CONCAT="${CONCAT}$(cat "$expanded_file")"` → `cat "$expanded_file" >> "$HASH_INPUT"`. Redirection does not go through a subshell; all bytes including trailing newlines are preserved.
- Hunk 3 (line 345): Resolved-file path — same replacement. Also changes the empty-input guard from `[[ -z "$CONCAT" ]]` (checks empty string) to `[[ ! -s "$HASH_INPUT" ]]` (checks zero-size temp file). Semantically equivalent for the no-inputs case.
- Hunk 4 (line 383): Hash computation — `echo -n "$CONCAT" | md5sum` → `md5sum < "$HASH_INPUT"` (and `md5 < "$HASH_INPUT"` on macOS). Stdin redirection preserves all bytes.

**`plugins/vsdd-factory/tests/input-hash.bats`** — 2 new tests

- Single-input: tool hash must equal `md5sum < file` reference for a file with trailing newlines.
- Multi-input: tool hash must equal `md5sum < concatenated-files` reference.

---

## Findings

No findings. The fix is minimal, mechanically correct, and directly targets the constraint that command substitution cannot be used when trailing-newline byte fidelity is required. The bats tests provide byte-accurate regression coverage for both the single-input and multi-input paths.

The migration note (existing artifacts on `factory-artifacts` carry stored hashes computed under the old accumulation; a one-pass `--scan --update` sweep at merge clears the backlog) is accurate and appropriately documented in the PR body.

---

## Cross-PR note (#715 / #718 same file)

Both #715 and #718 touch `plugins/vsdd-factory/bin/compute-input-hash` and `plugins/vsdd-factory/tests/input-hash.bats`. Their hunks are non-overlapping:

- #715 modifies the accumulation loop (lines ~215, ~337, ~345) and hash computation (~383).
- #718 modifies the `--update` case branch (~399+).
- The gap between #715's last changed line (~393) and #718's first changed line (~399) is sufficient; git context windows do not overlap.
- In `input-hash.bats`: #715 inserts at line ~295; #718 inserts at line ~84. Non-overlapping.
- Either PR can merge first. Merge #715 first is the lower-risk order (lower line-number insertions; #718's higher-line-number changes are unaffected by the line-count delta from #715).

---

## Verdict rationale

The fix is production-grade: it addresses the root cause (command substitution strips newlines), uses a well-established shell idiom (temp file + EXIT trap), does not introduce new failure modes, and ships with adequate bats regression coverage.

Issue #690, opened 2026-07-19
bug(hooks/validate-count-propagation): epic/ID label "E-11 stories" mis-parsed as count "11 stories" — ID-like substring false-positive forces STATE.md rewording


Zious11:
## Summary

The `validate-count-propagation` hook's count extractor mis-parses an **epic/identifier
label** as a numeric count. The token "E-11 stories" (referring to an epic whose ID is
`E-11`) was read as the count "11 stories," firing a count-propagation drift false-positive
and forcing an operator to reword STATE.md to avoid the pattern rather than fixing a real
inconsistency.

## Environment

- vsdd-factory plugin; hook `validate-count-propagation`.
- Observed on a downstream VSDD-factory-managed project during a STATE.md edit.

## Reproduction

1. Write a STATE.md line referencing an epic by ID next to the word "stories", e.g.
   "… → 5 E-11 stories …".
2. The count extractor matches `11` out of the `E-11` epic identifier and treats it as a
   claimed count of "11 stories".
3. It compares that phantom `11` against the real story count and reports
   count-propagation drift, even though every genuine count site is correct and mutually
   consistent.
4. Workaround applied: reword the STATE.md prose so the `E-11` token is not adjacent to a
   countable noun — a prose contortion to satisfy a false-positive.

## Root cause (inferred)

The count scanner extracts digit runs adjacent to countable nouns without excluding
digit runs that are part of an identifier token (`E-<n>`, epic/story IDs). `E-11` is an
identifier, not a quantity; the trailing `11` is not a count.

## Why this is distinct from existing issues

- **#567** — same hook (`validate-count-propagation`), but that is the *historical-vs-
  current* count-comparison false-positive (frozen changelog counts flagged as drift).
  This is a *tokenizer* false-positive: an identifier's digits read as a count. Different
  mechanism, same hook.
- **#641** — same *class* (an ID-like substring matched by a too-permissive pattern:
  "RED-4", "TD-001"), but a *different hook* (decision-chain-citation freshness). This is
  that class instantiated in `validate-count-propagation`.

## Proposed fix

Anchor the count extractor to require the digit run be a standalone quantity, not part of
an identifier: exclude matches immediately preceded by an identifier prefix (`E-`, `S-`,
`STORY-`, `BC-`, `TD-`, `D-`, etc. — i.e., `[A-Za-z]+-` immediately before the digits), or
require a word boundary that is not a hyphen preceded by a letter. Equivalently, tokenize
[...]

Issue #641, opened 2026-07-14
process-gap(hooks): decision-chain-citation freshness check false-positives on non-citation substrings (e.g. "RED-4", "TD-001") that merely resemble its ID pattern


arcavenai:
## What

A PostToolUse hook that checks a "decision-chain citation" freshness invariant (looking
for `D-\d+`-style decision-log references and verifying the highest cited value is current)
false-positived on a routine state-document write by matching unrelated substrings that
happen to contain a similar shape — e.g. "RED-4", "GREEN-4", and a debt-register ID like
"TD-001" were treated as if they were decision-chain citations. The hook then reported a
stale "max cited value" derived entirely from these false matches, with no genuine
decision-log reference anywhere in the actual write.

## Why it matters

This is a distinct root cause from the delta-vs-whole-file scoping problem already tracked
in #616 — the regex itself is too permissive, matching on a bare pattern shape (a letter,
a dash, digits) rather than requiring the specific decision-log ID prefix the hook is
actually meant to validate. It's part of a small family of advisory-only hook false
positives we've been tracking across one working session (the other four members are
already filed: undocumented STATE.md schema demands — #572; filename-based template
matching with no scope override — #617; whole-file-vs-delta evaluation — #616). All five
share the same operational shape: `block_intent=true`, the write always lands anyway, and
the disposition is always "confirm content landed, proceed" — which is a safe response but
means five separate hook-behavior defects are currently living entirely on operator
judgment rather than being fixed upstream.

## Proposed remedy

Anchor the decision-chain-citation regex to the actual ID prefix/format the framework's
own decision log uses (rather than a bare `[A-Z]+-\d+` shape), and/or require a minimum
context window (e.g. the literal word "Decision" or a known decision-log section header
nearby) before treating a match as a genuine citation.

## Cross-refs

#572, #616, #617 (sibling false-positive types in the same advisory-hook family,
already filed).

Framework-pattern detail only; from a private project running the factory.

arcaven:
One possible sibling for the batch: #690 (epic-label 'E-11 stories' mis-parsed as the count '11') looks like the same root class from a different hook — an ID/number extractor matching a bare shape rather than the specific token it's meant to read. If that read holds, a single principle might cover both: anchor each extractor to the actual prefix/format its target uses (here the decision-log ID, there the count field) and require a minimum surrounding context before treating a shape-match as the real thing, rather than fixing each hook's regex in isolation.

Issue #567, opened 2026-07-08
bug(hooks/validate-count-propagation): count_propagation_drift false positive — compares historical changelog counts against current counts, flagging accurate historical records as drift


arcavenai:
## Summary

The `validate-count-propagation` hook's `count_propagation_drift` check fires a false positive on every edit to BC-INDEX.md by comparing **historical** count mentions (old changelog entries, STATE.md phase-history rows) against each other or against the current count. Immutable historical records are flagged as drift even though the current-count sites are all correct and mutually consistent.

## Environment

- vsdd-factory plugin 1.0.0-rc.22 (claude-mp marketplace), greenfield run, Phase-3 wave-9 spec burst
- Hook: `validate-count-propagation`, code `count_propagation_drift`

## Reproduction (observed)

1. BC-INDEX.md changelog retains its v1.0 entry: "Lists all 36 BCs" (accurate for v1.0, 2026-06-27).
2. STATE.md phase-1 history row reads "PRD (38 BCs)" (accurate at phase-1 close).
3. Current truth: 41 BCs — correctly reflected in BC-INDEX frontmatter `bc_count: 41` and the §Count Verification section.
4. Any edit to BC-INDEX.md (in our case a v3.5→v3.6 bump adding a story-column entry, no count change) fires `count_propagation_drift` against the 36/38 historical mentions.

## Root cause (inferred)

The count-mention scanner does not distinguish current-count sites (frontmatter field, Count Verification section, live summary tables) from historical narrative (changelog entries, phase-history rows). Historical text is by definition frozen and SHOULD disagree with the current count.

## Impact

- Fires on every BC-INDEX edit for the life of the project once any count has ever changed — permanent alert fatigue on a check that exists to catch #277-class real drift.
- Perverse incentive: an agent could "fix" the warning by rewriting historical changelog entries, destroying the audit trail the changelog exists to provide.

## Suggested fix

Restrict the site inventory to designated current-count locations (frontmatter `bc_count`, Count Verification section, any table explicitly labeled as live) and exclude `changelog:` blocks and STATE.md historical rows. Alternatively, only compare sites whose surrounding context matches a "current" pattern.

## Related

Complements #277 (real derived-count staleness — that issue wants MORE propagation on live sites; this one wants historical sites EXCLUDED). Also adjacent to #563's principle: set membership should be re-derived from the generating tool, and historical prose is not part of the set.

arcaven:
[...]

PR #716, opened 2026-07-20
fix(hooks): stop validate-count-propagation reading epic/ID digits as counts

arcavenai:
## Why

`validate-count-propagation` mis-parses an epic/identifier label as a numeric count. In a STATE.md line like "5 E-11 stories", the count extractor's Pattern-A regex `([0-9][0-9,]+)[[:space:]]+(stories|...)` matches the `11` inside the `E-11` epic id and treats it as a claimed count of "11 stories" — firing a spurious `count_propagation_drift` block and forcing operators to reword STATE.md prose to dodge a false positive (#690). This is the ID-substring class from #641 instantiated in this hook (distinct from #567's historical-vs-current comparison bug). The fix follows the maintainer's stated remedy: tokenize and drop identifier tokens before counting, so an id's digits are never read as a quantity while genuine counts on the same line survive.

## What changed

- `plugins/vsdd-factory/hooks/validate-count-propagation.sh`
  - In `_extract_counts`, drop identifier tokens (`<letters>-<digits>` with optional dotted segments: `E-11`, `S-3`, `SS-01`, `TD-001`, `BC-2.1.001`) from each line before the Pattern A/B/C/D matchers run, via a pure-bash extglob substitution (`shopt -s extglob`; no new process/dependency). A genuine count elsewhere on the line is untouched.
  - Hardened the "no counts found" guard: `${#SOURCE_COUNTS[@]}` on an *empty associative array* is itself an unbound-variable error under `set -u` in bash 4+ — the original hook already crashed (exit 1) on any counts-free file, and dropping id tokens makes that path reachable from #690's exact input class ("5 E-11 stories" as the only count token). Switched to `[[ -z "${SOURCE_COUNTS[*]:-}" ]]` so the no-count path exits 0 cleanly instead of crashing — which is what delivering the maintainer's "no false positive, no reword" intent actually requires here.

- `plugins/vsdd-factory/tests/hooks.bats` — added a `validate-count-propagation` regression section (4 cases): epic-id token not parsed as a count (#690 repro → exit 0); genuine count drift still blocks (exit 2); mixed line keeps the genuine count after dropping the id (exit 2); all-phantom line exits clean with no output. The hook uses `declare -A` (bash 4+); the cases carry a skip guard keyed on the `/bin/bash` shebang-interpreter version, so they run on the ubuntu `validate` CI job (bash 5.x) and skip on macOS `/bin/bash` 3.2 rather than fail spuriously — the hook's own pre-existing requirement, not a limitation introduced here.

## Test evidence

Regex-level reproduction (the exact BASH_REMATCH miss):

```
INPUT LINE: Phase 3 complete → 5 E-11 stories delivered
=== Pattern A (current) ===
MATCHED. count=11 keyword=stories
  -> phantom count: 11 stories
```

RED — unfixed hook, STATE.md "5 E-11 stories" with a genuine, consistent 42 in STORY-INDEX.md:

```
exit=2
BLOCKED by validate-count-propagation: Count propagation drift detected in STATE.md: COUNT DRIFT DETECTED: '11 stories' in STATE.md but '42 stories' in STORY-INDEX.md.;   Run: grep -r "stories" .factory/specs/ .factory/STATE.md to reconcile. Fix: Update the lagging index to match the source-of-truth count: BC-INDEX total_bcs, ARCH-INDEX subsystem counts, or STORY-INDEX bcs count — see the specific drift cited above. Code: count_propagation_drift.
```

GREEN — fixed hook, same corpus (phantom `11` gone, no drift):

```
exit=0
[]
```

Regression, fixed hook under CI-equivalent bash 5.x:

```
[PASS] T46 epic-id not a count           (exit 0)
[PASS] T47 genuine drift blocks          (exit 2, "COUNT DRIFT DETECTED")
[PASS] T48 mixed line keeps genuine 13   (exit 2, "13 stories")
[PASS] T49 all-phantom line exits clean  (exit 0, empty output)
```

`bats tests/hooks.bats` locally (macOS /bin/bash 3.2): 49 ok, 4 skipped (the count-propagation cases, per the bash-4+ guard), 0 failures. `bash -n` clean under both bash 3.2 and 5.x. `check-bats-orphans.sh`: all hook references resolve.

Note: the four new cases run in the ubuntu-latest `validate` job (which runs `hooks.bats` under bash 5.x). They cannot execute on the macOS bash-3.2 legs, which run different suites; the skip guard makes that explicit rather than a red X on dev machines.

Refs: #690

### DIFF
--- a/plugins/vsdd-factory/hooks/validate-count-propagation.sh
+++ b/plugins/vsdd-factory/hooks/validate-count-propagation.sh
@@ -96,9 +96,15 @@ _extract_counts() {
     # Drop identifier tokens (E-11, S-3, BC-2.1.001, TD-001, SS-01) before
     # count extraction: the digits inside an ID are not a quantity. Without
     # this, "5 E-11 stories" mis-parses "11 stories" as a phantom count and
-    # fires a false count-propagation drift (#690). Matches <letters>-<digits>
-    # with optional dotted segments so multi-part BC/VP ids are dropped whole.
-    line="${line//+([A-Za-z])-+([0-9])?(*(.+([0-9])))/}"
+    # fires a false count-propagation drift (#690). Matches <letters>-<digits-
+    # and-dots> so multi-part BC/VP ids are dropped whole. Deliberately flat:
+    # a nested extglob (`?(*(.+([0-9])))`) matches the same ID tokens but
+    # backtracks super-linearly on long dotted-id lines (33s on a 600-char
+    # line of BC-N.NN.NN tokens vs 0.5s flat), and this runs per line on
+    # repo-controlled content. The flat class additionally eats dots glued to
+    # an ID (e.g. a sentence period in "delivered E-11."), which is harmless
+    # here — a whitespace-separated count token can never contain them.
+    line="${line//+([A-Za-z])-+([0-9.])/}"
     # Pattern A: count before keyword
     if [[ "$line" =~ ([0-9][0-9,]+)[[:space:]]+(BCs|VPs|stories|capabilities|subsystems) ]]; then
       keyword="${BASH_REMATCH[2]}"
--- a/plugins/vsdd-factory/hooks/validate-count-propagation.sh
+++ b/plugins/vsdd-factory/hooks/validate-count-propagation.sh
@@ -17,6 +17,7 @@
 # S-7.02 / BC-7.05.001, BC-7.05.002
 
 set -euo pipefail
+shopt -s extglob
 
 # Source canonical block-message helper if available (provides block_pre).
 _SELF_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
@@ -92,6 +93,12 @@ _extract_counts() {
   local path="$1"
   while IFS= read -r line; do
     local count keyword
+    # Drop identifier tokens (E-11, S-3, BC-2.1.001, TD-001, SS-01) before
+    # count extraction: the digits inside an ID are not a quantity. Without
+    # this, "5 E-11 stories" mis-parses "11 stories" as a phantom count and
+    # fires a false count-propagation drift (#690). Matches <letters>-<digits>
+    # with optional dotted segments so multi-part BC/VP ids are dropped whole.
+    line="${line//+([A-Za-z])-+([0-9])?(*(.+([0-9])))/}"
     # Pattern A: count before keyword
     if [[ "$line" =~ ([0-9][0-9,]+)[[:space:]]+(BCs|VPs|stories|capabilities|subsystems) ]]; then
       keyword="${BASH_REMATCH[2]}"
@@ -126,8 +133,13 @@ while IFS=: read -r kw cnt; do
   fi
 done < <(_extract_counts "$FILE_PATH")
 
-# Nothing to compare if no count patterns found
-if [[ ${#SOURCE_COUNTS[@]} -eq 0 ]]; then
+# Nothing to compare if no count patterns found.
+# Guard via ${arr[*]:-} rather than ${#arr[@]}: under `set -u`, expanding the
+# element count of an *empty associative array* is itself an unbound-variable
+# error in bash 4+. Dropping identifier tokens above makes an all-phantom line
+# (e.g. "5 E-11 stories" with no genuine count) leave this array empty, so this
+# no-count path is now reachable and must exit 0 cleanly, not crash.
+if [[ -z "${SOURCE_COUNTS[*]:-}" ]]; then
   exit 0
 fi
 
--- a/plugins/vsdd-factory/tests/hooks.bats
+++ b/plugins/vsdd-factory/tests/hooks.bats
@@ -364,3 +364,17 @@ require_bash4_hook_interp() {
   [ "$status" -eq 0 ]
   [ -z "$output" ]
 }
+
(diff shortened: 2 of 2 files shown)


arcaven:
The id-drop and the `set -u` empty-array guard both check out against a local repro — the #690 line exits 0, genuine drift still blocks with exit 2, and the old `${#arr[@]}` guard does crash under `set -u` on a counts-free file, so the switch is a real fix. One soft concern worth a look before merge: the nested extglob `+([A-Za-z])-+([0-9])?(*(.+([0-9])))` backtracks super-linearly on long dotted-id lines (a ~600-char line of `BC-1.11.11…`-shaped tokens ran ~1s locally, and longer lines grow much worse), and it runs per-line under `set -euo pipefail` on repo-controlled STATE.md content. A flatter `+([A-Za-z])-+([0-9.])` appears to drop the same id tokens in linear time with identical output on the cases I tried.

arcavenai:
Adopted — and the backtracking is worse than measured: on this machine the nested extglob took **33.4s** on a 550-char line of `BC-1.11.11` tokens, and 2× the length didn't finish in two minutes; the flat `+([A-Za-z])-+([0-9.])` runs the same line in ~0.5s. Output equivalence checked across the existing cases plus adversarial strings — the only differences are dots glued to an ID (sentence period after `E-11.", `X-.`), which can never be part of a whitespace-separated count token. Pushed in 8290156f with a long-line regression case that also asserts a genuine trailing count still extracts.

Zious11 (approved):
# PR #716 — Fresh-Eyes Review

**Title:** fix(hooks): stop validate-count-propagation reading epic/ID digits as counts
**Branch:** `fix/count-propagation-epic-id` → `develop`
**Verdict:** APPROVE

---

## Summary

Fixes #690: `validate-count-propagation.sh` mis-parses the digits inside an
identifier token (e.g. the `11` in `E-11`) as a numeric count, firing a
spurious `count_propagation_drift` block. The remedy tokenizes and drops
identifier tokens (`<letters>-<digits/dots>`) from each line before the
count matchers run, via a pure-bash extglob substitution, and hardens the
"no counts found" guard against empty-associative-array expansion under
`set -u`. Adds a 4-case regression section to `hooks.bats`.

I verified the core logic by hand against the diff:
- `"5 E-11 stories"` → `E-11` dropped → `"5  stories"` → single-digit `5`
  fails Pattern A's two-digit requirement (`[0-9][0-9,]+`) → empty
  `SOURCE_COUNTS` → new `[[ -z "${SOURCE_COUNTS[*]:-}" ]]` guard exits 0
  (the #690 repro, matches GREEN evidence).
- `"E-11 delivered 13 stories"` → id dropped, genuine `13` survives and
  still drifts against `42` → exit 2 (matches T48).
- Multi-part ids (`BC-2.1.001`) consumed whole by `+([0-9.])`, closing the
  `001`-as-phantom subclass.
- Commas in real counts (`1,234`) are not in the drop set, so genuine
  comma-grouped counts survive.
- The flat extglob pattern is deliberately chosen over a nested one to
  avoid super-linear backtracking — correct security/perf call for
  per-line repo-controlled content.

Diff is coherent, single-purpose, well under the 500-line threshold.
Conventional commit subject with `Refs: #690`. Class 0 (develop-tracked
files only); no story AC, so demo-evidence is N/A.

## Findings

| 1 | ADVISORY | description | Empty-array crash claim is overstated. PR body says `${#arr[@]}` on an empty associative array errors under `set -u` in "bash 4+" and that the hook "already crashed on any counts-free file." This was fixed in bash 4.4; on the ubuntu `validate` CI interpreter (bash 5.x) the old guard returns `0` and does not crash. Holds only for bash ≤4.3. | Correct the rationale to "bash ≤4.3" so future readers don't chase a non-reproducing crash. Fix itself is correct/safe on all versions. |
| 2 | ADVISORY | coherence | `shopt -s extglob` is enabled process-wide, not scoped to `_extract_counts`. It changes glob semantics for any later `[[ == pattern ]]`, `case`, or param-substitution glob in the file. Cannot confirm no downstream effect from the diff alone. | Confirm no other glob/case in the file relies on extglob being off; consider scoping via subshell or `shopt -u` after the loop. |
| 3 | MINOR | coherence | The id-drop pattern `+([A-Za-z])-+([0-9.])` also deletes legitimate `word-NN` prose tokens (e.g. `top-10 stories`), which could mask a genuine count on such a line (false negative). Documented, deliberate tradeoff; low risk on repo-controlled content. | Accept consciously; the ReDoS-safe flat pattern is the right call. |
| 4 | MINOR | coverage | Pre-existing (not introduced): Pattern A requires two digits, so single-digit counts are never detected. Out of scope; noted for completeness. | None required this PR. |

No blocking findings.

## Checklist

1. Diff coherence — PASS (single-purpose #690 fix).
2. Description accuracy — PASS with finding #1 (imprecise crash rationale).
3. Test coverage — PASS (4 regression cases; RED/GREEN evidence provided).
4. Demo evidence — N/A (Class 0 hook fix, no story AC).
5. Commit quality — PASS (conventional, `Refs: #690`).
6. Diff size — PASS (well under 500 lines).
7. Missing changes — PASS.
8. Dependency status — PASS (no upstream deps).

## Verdict

**APPROVE.** The fix is sound and matches the maintainer's stated remedy.
All findings are non-blocking; correcting the empty-array rationale (#1)
and confirming the global `extglob` has no downstream effect (#2) would
tighten it but are not merge blockers.

Issue #473, opened 2026-07-04
bug(phase-4 workflow): scenario-rotation step tasks the write-denied orchestrator with writing a file, to a non-canonical directory


arcavenai:
## Two defects in one step

`workflows/phases/phase-4-holdout-evaluation.lobster` (and the inlined copy in `workflows/greenfield.lobster:~1014`):

```yaml
- name: scenario-rotation
  type: agent
  agent: orchestrator
  task: >
    Perform scenario rotation for this pipeline run. Randomly select 80%
    of holdout scenarios. Write selected scenario IDs to
    .factory/holdout-evaluation/scenario-selection.json.
```

### Defect 1 — tool contract violation

The step is assigned to `agent: orchestrator` and its task is *"Write selected scenario IDs to ... scenario-selection.json"*. But the orchestrator's own agent definition is explicit and non-negotiable:

> - You NEVER write ANY files — you delegate all writing to specialist agents
> - Tool Access ... Denied: `write`, `edit`, `apply_patch`, `exec`, `process`

As written, the step cannot be executed by its declared agent without violating the agent's constitution. In practice the orchestrator either (a) violates its write prohibition, (b) spawns state-manager for a one-line JSON write (undocumented, improvised), or (c) skips the artifact. All three are divergence; none is specified.

**Fix:** either route the write through state-manager in the workflow (`agent: state-manager` with the selection provided by the orchestrator), or make selection an input to the holdout-eval skill step rather than a persisted artifact requiring a write.

### Defect 2 — non-canonical output directory

The step writes to `.factory/holdout-evaluation/scenario-selection.json`, but the canonical `.factory/` structure (enforced/created by `/factory-health` check 5) has **no `holdout-evaluation/` directory**. The canonical tree is:

```
holdout-scenarios/
holdout-scenarios/wave-scenarios/
holdout-scenarios/evaluations/
```

and `skills/holdout-eval/SKILL.md` writes results to `.factory/holdout-scenarios/evaluations/wave-<N>/`. The rotation artifact lands in a sibling directory that no health check knows about, no template references, and `relocate-artifact`'s path registry presumably doesn't cover. It will read as orphan clutter on the orphan branch (compare #230).

**Fix:** `.factory/holdout-scenarios/evaluations/<run>/scenario-selection.json` (or wherever the evaluation report template expects it), and register the path in the artifact-path registry.

## Environment / evidence

[...]

Issue #300, opened 2026-06-27
bug(artifact-path-registry): no entry for L1 product-brief — create-brief output falls outside template/drift/register governance


arcavenai:
## Summary

`config/artifact-path-registry.yaml` has **no entry for the L1 product brief**, even though `skills/create-brief` writes the canonical artifact to `.factory/specs/product-brief.md`. The registry has 194 entries; a grep for `brief` returns zero matches.

## Evidence

```
$ grep -in "brief" config/artifact-path-registry.yaml
(no matches)
$ grep -c ":" config/artifact-path-registry.yaml
194
```

`skills/create-brief/SKILL.md` documents the output path as `.factory/specs/product-brief.md`, but that path is not registered.

## Impact

The registry is the source of truth for the registry-driven cross-cutting skills:

- `/vsdd-factory:register-artifact`
- `/vsdd-factory:validate-template-compliance`
- `/vsdd-factory:conform-to-template`
- `/vsdd-factory:check-input-drift`

With no `product-brief` entry, the **L1 brief — the document every downstream artifact (L2 domain spec, L3 PRD, behavioral contracts, architecture) traces back to** — silently falls outside these governance checks. It cannot be template-validated, cannot be drift-checked, and `register-artifact` has no canonical path/index to slot it into.

This is a coverage/governance gap, not data loss. But it is load-bearing: the brief is the root of the traceability chain (`traces_to`), and it is precisely the artifact you'd most want drift-checked when inputs change at the Phase 1 gate.

## Suggested fix

Add a `product-brief` entry to `config/artifact-path-registry.yaml`:
- path: `.factory/specs/product-brief.md`
- level: L1
- producer: `product-owner`
- index/template: as appropriate (the brief has a stable frontmatter + section structure that `validate-template-compliance` could check).

## Environment

- vsdd-factory `1.0.0-rc.21`
- Observed during a greenfield Phase 1 run (create-brief -> L2 -> L3 -> architecture).

## Cross-refs

Adjacent but distinct from registry issues #219 (subsystem-registry), #213 (decision-registry), #242 (resolver load_registry). This is specifically the artifact-path-registry missing the brief.

arcaven:
Confirmed on `develop`: the L2 `domain-spec` and L3 `prd` artifact types are both registered, but there's no L1 `product-brief` entry, and `.factory/specs/product-brief.md` matches none of the existing patterns.

[...]

PR #717, opened 2026-07-20
fix(config): register holdout scenario-selection path in artifact-path-registry

arcavenai:
## Why

The phase-4 scenario-rotation step writes `.factory/holdout-scenarios/scenario-selection.json`, but the artifact-path registry has no entry that matches it. Since `validate-artifact-path` hard-blocks any `.factory/` write with no matching registry entry (`ARTIFACT_PATH_UNREGISTERED`), the canonical write is blocked at dispatch time — the exact remaining gap you kept #473 open for after #527 corrected the workflow's agent and directory. This adds the missing registry pattern so the write the workflow already asks for is actually permitted, and locks it in with a test that loads the real shipped registry.

## What changed

- `plugins/vsdd-factory/config/artifact-path-registry.yaml`: added a `holdout-scenario-selection` entry under the "Holdout Evaluations" section, mirroring the neighboring entries' schema exactly (`artifact_type` / `canonical_path_pattern` / `description` / `enforcement_level: block`). Pattern is the literal canonical path both phase-4 workflow files write: `.factory/holdout-scenarios/scenario-selection.json` (phase-4-holdout-evaluation.lobster:21 and the inlined greenfield.lobster:1014).
- `crates/hook-plugins/validate-artifact-path/src/tests.rs`: two tests that load the **actual production registry** (walking up from `CARGO_MANIFEST_DIR` — see note below) and assert the selection path returns `MatchResult::Block`, plus an end-to-end `hook_logic` check that the write is allowed (`Continue`). A helper hard-fails if the registry can't be located, so the tests can never pass vacuously.

Note: the existing EC-007 benchmark tries to load the production registry via a hard-coded `../../../../` path that resolves one level too high, so it silently falls back to a fixture. The new tests use an ancestor-walk instead; happy to file that as a separate observation if useful.

## Test evidence

RED (before the registry entry, tests already in place):

```
running 2 tests
test tests::test_issue_473_scenario_selection_hook_allows_write ... FAILED
test tests::test_issue_473_scenario_selection_matches_production_registry ... FAILED

---- test_issue_473_scenario_selection_matches_production_registry ----
assertion `left == right` failed: #473: canonical scenario-rotation output
'.factory/holdout-scenarios/scenario-selection.json' must match a block entry
in the production registry (currently NoMatch).
  left: NoMatch
 right: Block

---- test_issue_473_scenario_selection_hook_allows_write ----
#473: hook_logic must Continue (allow) the write ...; got Block { reason:
"... Write to '.factory/holdout-scenarios/scenario-selection.json' under
.factory/ has no matching entry ... Code: ARTIFACT_PATH_UNREGISTERED." }

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 48 filtered out
```

GREEN (after adding the registry entry):

```
running 2 tests
test tests::test_issue_473_scenario_selection_matches_production_registry ... ok
test tests::test_issue_473_scenario_selection_hook_allows_write ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 48 filtered out
```

Full crate suite (no regressions):

```
test result: ok. 50 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out   (lib)
test result: ok. 4 passed; 0 failed; ...                                       (kani_path_matching)
test result: ok. 6 passed; 0 failed; ...                                       (proptest_registry_load)
```

Pairs with the sibling registry PR for #300 — separate entries in different sections; either merges cleanly first (verified: both patches apply to develop tip in either order, and the combined tree compiles + passes).

Refs: #473

### DIFF
--- a/crates/hook-plugins/validate-artifact-path/src/tests.rs
+++ b/crates/hook-plugins/validate-artifact-path/src/tests.rs
@@ -1583,3 +1583,87 @@ fn test_F_P18_001_partial_match_rejected_leading_slash_discipline() {
         partial_match
     );
 }
+
+// -----------------------------------------------------------------------
+// Issue #473: scenario-rotation writes .factory/holdout-scenarios/
+// scenario-selection.json, but that path had no registry entry — so
+// validate-artifact-path blocked the state-manager's canonical write with
+// ARTIFACT_PATH_UNREGISTERED.
+//
+// PR #527 corrected the workflow (agent: state-manager, canonical
+// holdout-scenarios/ directory) but the registry still lacked the matching
+// pattern. These tests load the ACTUAL production registry (same harness as
+// EC-007) and assert the canonical selection path now matches with block
+// enforcement, and that the hook allows the write end-to-end.
+//
+// Refs: #473, PR #527.
+// -----------------------------------------------------------------------
+
+/// Locate the shipped production registry by walking up from
+/// CARGO_MANIFEST_DIR until `plugins/vsdd-factory/config/artifact-path-registry.yaml`
+/// is found. Layout-independent (unlike a hard-coded `../` count) so it can
+/// never silently fall back to a fixture and pass vacuously.
+fn production_registry_path_i473() -> std::path::PathBuf {
+    let manifest_dir =
+        std::env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR must be set during test");
+    let mut dir = std::path::PathBuf::from(&manifest_dir);
+    loop {
+        let candidate = dir.join("plugins/vsdd-factory/config/artifact-path-registry.yaml");
+        if candidate.exists() {
+            return candidate;
+        }
+        if !dir.pop() {
+            panic!(
+                "could not locate plugins/vsdd-factory/config/artifact-path-registry.yaml \
+                 walking up from CARGO_MANIFEST_DIR ({manifest_dir})"
+            );
+        }
+    }
+}
+
+/// Read the shipped production registry as YAML text.
+fn production_registry_yaml_i473() -> String {
+    let path = production_registry_path_i473();
+    std::fs::read_to_string(&path)
+        .unwrap_or_else(|e| panic!("failed to read registry at {}: {}", path.display(), e))
+}
+
+#[test]
+fn test_issue_473_scenario_selection_matches_production_registry() {
+    // #473: .factory/holdout-scenarios/scenario-selection.json is the canonical
+    // write target for the phase-4 scenario-rotation step (both the phase file
+    // and the inlined greenfield copy). It MUST match a block entry in the
+    // shipped registry, otherwise state-manager's write is blocked with
+    // ARTIFACT_PATH_UNREGISTERED.
+    let registry =
+        load_registry(&production_registry_yaml_i473()).expect("production registry must load");
+    let path = ".factory/holdout-scenarios/scenario-selection.json";
+    let result = matches_canonical(path, &registry);
+    assert_eq!(
+        result,
+        MatchResult::Block,
+        "#473: canonical scenario-rotation output '{}' must match a block entry in the \
+         production registry (currently {:?}). Without the entry, validate-artifact-path \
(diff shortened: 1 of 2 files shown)


Zious11 (approved):
# PR #717 Fresh-Eyes Review — fix(config): register holdout scenario-selection path in artifact-path-registry

**Branch:** `fix/register-scenario-selection` → `develop`
**Refs:** #473 (pairs with sibling registry PR for #300)

## VERDICT: APPROVE

No blocking findings. Clean, tightly-scoped fix: one registry entry mirroring the neighboring
Holdout Evaluations schema exactly, plus a proper RED→GREEN test pair that loads the real shipped
registry (never vacuous — helper hard-panics if the registry can't be located).

## Checklist

| # | Item | Result |
|---|------|--------|
| 1 | Diff coherence | PASS — both files serve #473; registry entry + tests only. No unrelated changes. |
| 2 | Description accuracy | PASS — PR body matches the diff exactly. |
| 3 | Test coverage | PASS — RED (2 failing) → GREEN (2 passing) shown; tests load the actual production registry. |
| 4 | Demo evidence | N/A — config/hook fix PR (Class 0), not a story with ACs; fix-pr-delivery skips demos. |
| 5 | Commit quality | PASS — conventional `fix(config):`, Refs #473. |
| 6 | Diff size | PASS — ~90 lines added, well under the 500-line flag. |
| 7 | Missing changes | PASS with one advisory (see finding 1 — workflow files not in diff, parity unverifiable). |
| 8 | Dependency status | Notes dependence on #527 (workflow correction) and sibling #300 PR; no hard correctness coupling — the registry entry is harmless even if #527 lags. |

## Findings

### [ADVISORY] Pattern-vs-workflow parity unverifiable from the diff
The value of this PR rests on the literal pattern `.factory/holdout-scenarios/scenario-selection.json`
exactly matching what the phase-4 workflow writes. Neither phase-4 workflow file is in the diff, so I
cannot confirm the workflow writes that exact path with no intermediate glob segment. A per-cycle
subdirectory (e.g. `.factory/holdout-scenarios/<cycle>/scenario-selection.json`) would NOT match this
literal, non-globbed pattern. The two new tests assert against the same hard-coded string, so they
cannot catch a registry-vs-runtime mismatch. **Recommend the maintainer confirm the workflow's write
path is the literal, un-nested string before merge.** This is the only thing that could make the fix a
no-op in production while tests stay green.

### [SUGGESTION] Disclosed EC-007 pre-existing defect is real test-coverage rot
The description notes EC-007 loads the production registry via a hard-coded `../../../../` path that
resolves one level too high and "silently falls back to a fixture" — meaning EC-007 likely validates a
fixture, not the shipped registry (vacuous-pass risk on an existing test). Under the production-grade
default, the cleanest resolution is to repoint EC-007 at the same ancestor-walk helper this PR
introduces rather than deferring to a separate observation. Not a blocker for this PR's scope; the fix
is a few lines and reuses code already added here.

### [ADVISORY] Ancestor-walk locator robust against vacuous passes, not out-of-tree builds
`production_registry_path_i473()` walks up from `CARGO_MANIFEST_DIR` and hard-panics if the registry
isn't found — a genuine improvement over EC-007 that prevents vacuous passes. Residual risk: a
packaged/vendored build where `plugins/vsdd-factory/config/` is not an ancestor would panic rather than
skip. Acceptable (loud failure > silent skip); noted for awareness.

### [NIT] `*_i473` issue-number suffix on helper identifiers
Reads slightly unusual as a permanent API surface, but namespaces the helpers to avoid collision with
the EC-007 harness and is self-documenting. Fine to keep.

## What I verified
- The new registry entry uses a literal (non-glob) `canonical_path_pattern`, so it cannot over-match
  other `.factory/` paths.
- The entry is placed inside the Holdout Evaluations section, before the Measurements section, and
  mirrors the neighboring entry schema (`artifact_type` / `canonical_path_pattern` / `description` /
  `enforcement_level: block`).
- RED evidence shows both tests failing with `NoMatch` / `ARTIFACT_PATH_UNREGISTERED` before the entry;
  GREEN shows both passing after — confirming the tests are load-bearing and tied to the registry entry.

PR #719, opened 2026-07-20
fix(hooks): recognize named S-<PREFIX>.<NAME> story IDs in merge-prereq extraction

### LINKED ISSUE #658: validate-pr-merge-prerequisites hook: STORY_ID regex doesn't recognize S-<PREFIX>.<NAME> IDs, falls back to first S-N.NN substring anywhere in prompt
## Summary

`hooks/validate-pr-merge-prerequisites.sh` (plugin rc.22) extracts the target story ID
from the Agent-dispatch prompt via:

```bash
STORY_ID=$(echo "$PROMPT" | grep -oE 'STORY-[0-9]+' | head -1 || true)
if [[ -z "$STORY_ID" ]]; then
  STORY_ID=$(echo "$PROMPT" | grep -oE 'S-[0-9]+\.[0-9]+' | head -1 || true)
fi
```

This assumes every story ID is either `STORY-NNN` or `S-N.NN` (pure numeric). Our
project (switchboard-blue) also uses named story IDs of the form `S-BL.<NAME>` (e.g.
`S-BL.DISCOVERY-WIRE`, `S-BL.CLI-SURFACE-COMPLETION`, `S-BL.PE-RECEIVE-LOOP`) — neither
regex matches these, so `STORY_ID` extraction silently falls through to whatever
`S-[0-9]+\.[0-9]+`-shaped substring happens to appear *anywhere else* in the prompt
text, with no verification that it's actually the story being merged.

## Reproduction

Dispatching a `github-ops` merge for PR #123 (story `S-BL.DISCOVERY-WIRE`) with a
prompt whose body incidentally mentioned an unrelated *dependency* story
(`S-7.02`, already merged months earlier via a different PR) caused the hook to
resolve `STORY_ID=S-7.02` instead of `S-BL.DISCOVERY-WIRE`, then block the merge
citing `.factory/code-delivery/S-7.02/pr-review.md` and
`.factory/code-delivery/S-7.02/security-review.md` as missing — evidence files for a
story that has nothing to do with the PR being merged.

Error surfaced:
```
BLOCKED by validate-pr-merge-prerequisites: pr-manager 9-step lifecycle incomplete for
S-7.02: missing Missing: .factory/code-delivery/S-7.02/pr-review.md (PR review not
conducted);Missing: .factory/code-delivery/S-7.02/security-review.md (security review
not conducted);
```

## Secondary finding (this project's own convention gap)

Auditing every story directory under this project's `.factory/code-delivery/` (44
directories, spanning PRs #113 through #122, all successfully merged) found **zero**
directories with a standalone `security-review.md` file — every prior PR embedded its
security review content directly in the PR description body instead (see merged PR
#120's "Security Review" section for the established pattern). Check 3's fallback path
(grep the PR description for "security.*clean|security.*no finding|security.*pass|no
security") also requires the match to be within a single line, which markdown
section-heading + verdict-on-next-line formatting (the natural way to write this)
defeats.

Net effect: even for stories whose ID *does* match the recognized regex shapes, this
check would have blocked every prior merge in this project's history had it existed
then, despite those PRs having genuinely conducted and documented security review. The
file-existence check's real-world false-negative rate on this project's actual
established convention appears to be ~100%.

## Suggested fix

1. Broaden `STORY_ID` extraction to also match named story IDs — e.g.
   `S-[A-Za-z]+\.[A-Za-z0-9-]+` as a third fallback, or (more robustly) resolve the
   story ID from the current git branch name / `.factory/STATE.md` rather than
   free-text grep over an LLM-authored prompt, which is inherently fragile (any
   incidental mention of an unrelated identifier in the prompt can hijack the match).
2. Fix the single-line grep assumption in check 3 to tolerate the security-review
   verdict appearing on a different line than the "Security Review" heading (e.g.
   `grep -Pzo` multiline match, or just search for a verdict-shaped line
   independently of proximity to the heading).
3. Consider whether file-existence is the right signal at all, given point 2 above —
   this project's actual, working, multi-PR-validated convention is "security review
   content lives in the PR description," not a sibling file.

## Context

Same plugin version/class as drbothen/vsdd-factory#625 (`validate-policies-schema` hook
enforcing a fixture-derived schema this project never adopted) and #626 (single-identity
review — `reviewDecision` structurally empty). This is the third rc.22 hook we've hit
whose validation logic doesn't match an established, working project convention.

arcavenai:
## Why

`validate-pr-merge-prerequisites.sh` extracts the target story ID from the
github-ops dispatch prompt by trying `STORY-[0-9]+` and then `S-[0-9]+\.[0-9]+`.
A named story such as `S-BL.DISCOVERY-WIRE` matches neither pattern, so
extraction falls through to the first `S-N.NN`-shaped substring anywhere in the
prompt — typically a merged dependency like `S-7.02`. The hook then blocks the
merge citing the *wrong* story's missing evidence files, a story that has
nothing to do with the PR being merged. Any project using named story IDs
(alpha-prefixed) hits this on every merge dispatch whose prompt mentions a
numeric dependency.

## What changed

Added a precedence-aware third branch to the extraction chain. Order is
most-specific-first: `STORY-NNN`, then the named form `S-<PREFIX>.<NAME>`
(`S-[A-Za-z]+\.[A-Za-z0-9-]+`), then the pure-numeric `S-N.NN`. The named
pattern is inserted *before* the numeric pattern rather than appended after it —
appending would leave `S-N.NN` winning first and would not fix the bug. When a
named ID is present it is now resolved even if a numeric dependency ID also
appears elsewhere in the prompt.

Extended `tests/pr-lifecycle-hooks.bats` with four cases: named target resolves
past a numeric dependency (pass), named target blocks citing its own ID when
evidence is missing, `STORY-NNN` still wins over any S-form, and pure-numeric
still resolves when no named ID is present.

## Test evidence

Extraction repro, before the fix (raw logic), target is `S-BL.DISCOVERY-WIRE`:

```
$ echo 'gh pr merge 123 story S-BL.DISCOVERY-WIRE depends on S-7.02' | grep -oE 'S-[0-9]+\.[0-9]+' | head -1
S-7.02          # dependency hijacks the match
```

New regression tests against the UNFIXED hook (red):

```
1..2
not ok 1 pr-merge-prerequisites: resolves named story, not a numeric dependency in the prompt (issue #658)
#   `[ "$status" -eq 0 ]' failed
not ok 2 pr-merge-prerequisites: blocks citing the named story ID when its evidence is missing (issue #658)
#   `[ "$status" -eq 2 ]' failed
```

Full `pr-lifecycle-hooks.bats` suite against the FIXED hook (green), 49/49:

```
1..49
ok 11 pr-merge-prerequisites: passes when all evidence files exist
...
ok 19 pr-merge-prerequisites: warns when delivery dir not found
ok 20 pr-merge-prerequisites: resolves named story, not a numeric dependency in the prompt (issue #658)
ok 21 pr-merge-prerequisites: blocks citing the named story ID when its evidence is missing (issue #658)
ok 22 pr-merge-prerequisites: STORY-NNN still takes precedence over any S-form in the prompt
ok 23 pr-merge-prerequisites: pure-numeric S-N.NN still resolves when no named id present
...
ok 49 pr-manager step-8d: force-delete error taxonomy classifies transient errors
```

Closes: #658

### DIFF
--- a/plugins/vsdd-factory/hooks/validate-pr-merge-prerequisites.sh
+++ b/plugins/vsdd-factory/hooks/validate-pr-merge-prerequisites.sh
@@ -46,10 +46,23 @@ if ! echo "$PROMPT" | grep -qiE "gh pr merge|pr merge|merge.*PR"; then
   exit 0  # not a merge dispatch
 fi
 
-# Extract story ID from prompt
+# Extract story ID from prompt.
+#
+# Precedence matters: named story IDs (S-<PREFIX>.<NAME>, e.g.
+# S-BL.DISCOVERY-WIRE) must be recognized BEFORE the pure-numeric S-N.NN form.
+# Otherwise a numeric substring elsewhere in the prompt — typically a merged
+# dependency such as S-7.02 — hijacks the match and the hook blocks citing the
+# wrong story's missing evidence (issue #658). Most-specific-first:
+#   1. STORY-NNN            (explicit long form)
+#   2. S-<PREFIX>.<NAME>    (named story, alpha prefix)
+#   3. S-N.NN               (pure-numeric story)
 STORY_ID=$(echo "$PROMPT" | grep -oE 'STORY-[0-9]+' | head -1 || true)
 if [[ -z "$STORY_ID" ]]; then
-  # Try S-N.NN format
+  # Named story: alpha prefix, alnum/hyphen name (e.g. S-BL.DISCOVERY-WIRE)
+  STORY_ID=$(echo "$PROMPT" | grep -oE 'S-[A-Za-z]+\.[A-Za-z0-9-]+' | head -1 || true)
+fi
+if [[ -z "$STORY_ID" ]]; then
+  # Pure-numeric story (e.g. S-7.02)
   STORY_ID=$(echo "$PROMPT" | grep -oE 'S-[0-9]+\.[0-9]+' | head -1 || true)
 fi
 if [[ -z "$STORY_ID" ]]; then
--- a/plugins/vsdd-factory/tests/pr-lifecycle-hooks.bats
+++ b/plugins/vsdd-factory/tests/pr-lifecycle-hooks.bats
@@ -240,6 +240,56 @@ EOF
   [[ "$output" == *"WARNING"* ]]
 }
 
+# ------------------------------------------------------------------------
+# Named story-ID extraction precedence (issue #658)
+#
+# S-<PREFIX>.<NAME> ids (e.g. S-BL.DISCOVERY-WIRE) must be recognized before
+# the pure-numeric S-N.NN form. Otherwise a numeric substring elsewhere in the
+# prompt — typically a merged dependency (S-7.02) — hijacks the match and the
+# hook blocks citing the wrong story's missing evidence.
+# ------------------------------------------------------------------------
+
+@test "pr-merge-prerequisites: resolves named story, not a numeric dependency in the prompt (issue #658)" {
+  # Named target story: fully evidenced.
+  mkdir -p "$WORK/.factory/code-delivery/S-BL.DISCOVERY-WIRE"
+  echo "# Description" > "$WORK/.factory/code-delivery/S-BL.DISCOVERY-WIRE/pr-description.md"
+  echo "# Review" > "$WORK/.factory/code-delivery/S-BL.DISCOVERY-WIRE/pr-review.md"
+  echo "# Security" > "$WORK/.factory/code-delivery/S-BL.DISCOVERY-WIRE/security-review.md"
+  # Merged dependency directory exists but carries no evidence — the old regex
+  # would resolve here and block.
+  mkdir -p "$WORK/.factory/code-delivery/S-7.02"
+  _run_pretool_agent "vsdd-factory:github-ops" \
+    "cd $WORK && gh pr merge 123 --squash for story S-BL.DISCOVERY-WIRE depends on S-7.02"
+  [ "$status" -eq 0 ]
+}
+
+@test "pr-merge-prerequisites: blocks citing the named story ID when its evidence is missing (issue #658)" {
+  mkdir -p "$WORK/.factory/code-delivery/S-BL.PE-RECEIVE-LOOP"
+  _run_pretool_agent "vsdd-factory:github-ops" \
+    "cd $WORK && gh pr merge 88 --squash for story S-BL.PE-RECEIVE-LOOP"
+  [ "$status" -eq 2 ]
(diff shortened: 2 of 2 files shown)


Zious11 (approved):
# PR Review — #719 fix(hooks): recognize named S-<PREFIX>.<NAME> story IDs in merge-prereq extraction

**Verdict:** APPROVE
**Date:** 2026-07-22
**Closes:** #658

---

## Summary

`validate-pr-merge-prerequisites.sh` extracts the target story ID from the github-ops dispatch prompt using two patterns: `STORY-[0-9]+` then `S-[0-9]+\.[0-9]+`. Named story IDs such as `S-BL.DISCOVERY-WIRE` matched neither, so extraction fell through to the first pure-numeric `S-N.NN` substring — typically a merged dependency like `S-7.02`. The hook then blocked the merge citing the wrong story's missing evidence files. This PR inserts a named-ID branch between the two existing patterns, in most-specific-first order.

---

## Changes reviewed

**`plugins/vsdd-factory/hooks/validate-pr-merge-prerequisites.sh`** — 1 hunk

- Adds a comment block explaining the precedence requirement and why appending would not fix the bug (the pure-numeric pattern would win first).
- Inserts a second `if [[ -z "$STORY_ID" ]]; then` block with `grep -oE 'S-[A-Za-z]+\.[A-Za-z0-9-]+' | head -1` between STORY-NNN and pure-numeric extraction. Order: STORY-NNN → S-<PREFIX>.<NAME> → S-N.NN.
- The named regex `S-[A-Za-z]+\.[A-Za-z0-9-]+` correctly requires an alphabetic prefix, so `S-7.02` cannot match it. `STORY-NNN` is distinct (starts with `STORY-` not `S-`), so no cross-contamination.
- `|| true` appended to each `grep` call is consistent with the existing branches and prevents `set -e` from exiting on zero-match.

**`plugins/vsdd-factory/tests/pr-lifecycle-hooks.bats`** — 4 new tests

- Named target fully evidenced, prompt also contains a numeric dependency → passes (story-id extraction resolves the named ID, not the dependency).
- Named target missing evidence → blocks citing the named story ID, not the dependency's ID.
- `STORY-NNN` still takes precedence over any S-form.
- Pure-numeric `S-N.NN` still resolves when no named ID present.

---

## Findings

No findings. The most-specific-first ordering is the correct fix approach; appending the named branch after the numeric branch would not have fixed the bug (as the PR body correctly explains). The regex correctly scopes to alphabetic prefixes, preventing the numeric `S-N.NN` form from accidentally matching the named branch. Test coverage is complete across all four precedence cases.

---

## Verdict rationale

The fix is minimal, correctly ordered, and ships with full four-case regression coverage for the extraction precedence chain. No correctness issues were found.

PR #720, opened 2026-07-20
fix(hooks): accept 7-char short SHAs in verify-sha-currency develop cite

### LINKED ISSUE #629: bug(hooks): verify-sha-currency SHA regex requires 8+ chars — 7-char SHAs cited in STATE.md report NOT_FOUND (false negative on every run)
## Observed (engine v1.0.0-rc.22, downstream switchboard-blue)

`verify-sha-currency.sh` reports `STATE.md develop cited: NOT_FOUND` on every run because the downstream STATE.md frontmatter cites `develop_head: 4c276d9` (7 chars — git's default short-SHA length on this repo) while the script's extraction regex requires an 8-char minimum. The check still PASSes overall, so the currency comparison for the develop SHA is silently skipped rather than performed — a false negative that defeats the check's purpose for exactly the field it most needs to verify.

State-manager agents have been treating this as "pre-existing known, doesn't fail the check" — i.e. the broken sub-check has normalized.

## Proposal

Accept 7-40 char hex SHAs (`[0-9a-f]{7,40}`) in the extraction regex, and make an unextractable cited SHA a WARN rather than a silent NOT_FOUND-but-PASS.

Provenance: 2026-07-12 board-close burst report, switchboard-blue. Severity LOW (diagnostic integrity).

arcavenai:
## Why

`verify-sha-currency.sh` extracts the cited develop SHA with
`grep -oE 'develop_head: "?[0-9a-f]{8,40}'` and truncates it to 8 chars. Git's
common default short-SHA length is 7. A STATE.md that cites `develop_head: 4c276d9`
(7 chars, correct value) reports `develop cited: NOT_FOUND`, and because the
comparison is guarded by `!= "NOT_FOUND"` the currency check is silently
*skipped* while the overall run still PASSes — a false negative on exactly the
field this check exists to verify. State-manager agents had normalized this as
"pre-existing known, doesn't fail the check."

## What changed

- Extraction regex `{8,40}` → `{7,40}` and dropped the `cut -c1-8` truncation so
  the cited value is preserved at whatever length it was written.
- The develop comparison is now a prefix match against the *full* actual SHA
  (`${ACTUAL_DEV_FULL:0:${#CITED_DEV_STATE}}`), which is the git-canonical way to
  compare a short SHA of any length (7..40). String-equality against a fixed
  8-char `ACTUAL_DEV` would have false-positive FAILed a legitimate 7-char cite
  once the regex accepted it. The same prefix comparison is applied to the
  HANDOFF.md cite for consistency.
- Added a WARN when STATE.md declares a `develop_head:` field but no hex SHA can
  be extracted from it, so the silent-skip-PASS class cannot recur — an
  unparseable cite is now surfaced instead of swallowed.

## Test evidence

`templates/` scripts have no bats suite in the repo (verified: no `.bats` file
references `verify-sha-currency` or invokes any `templates/*.sh`). Red/green was
done by direct invocation against a throwaway git fixture whose STATE.md cites
the correct develop head as a 7-char short SHA.

Before the fix (upstream script, 7-char correct cite):

```
Actual develop HEAD      : 203fdca5
STATE.md    develop cited      : NOT_FOUND
PASS: all SHA currency + burst hygiene checks pass      # exit 0 — silently skipped
```

After the fix, across SHA-length and drift cases:

```
1. correct 7-char cite  -> exit 0 ; cited: 203fdca ; PASS: all SHA currency + burst hygiene checks pass
2. correct 8-char cite  -> exit 0 ; PASS
3. correct 40-char cite -> exit 0 ; PASS
4. STALE 7-char cite     -> exit 1 ; FAIL: SHA drift or burst-hygiene issues detected
5. empty develop_head    -> exit 0 ; WARN emitted ; PASS (with WARN)
6. no develop_head field -> exit 0 ; no WARN ; PASS
```

A correct 7-char cite now PASSes with the SHA actually extracted; a stale cite
FAILs; a declared-but-unextractable field WARNs; an absent field stays clean.

Closes: #629

### DIFF
--- a/plugins/vsdd-factory/templates/verify-sha-currency.sh
+++ b/plugins/vsdd-factory/templates/verify-sha-currency.sh
@@ -148,12 +148,12 @@ fi
 
 if [ "$CITED_DEV_STATE" != "NOT_FOUND" ] \
     && [ "${ACTUAL_DEV_FULL:0:${#CITED_DEV_STATE}}" != "$CITED_DEV_STATE" ]; then
-  echo "FAIL: develop SHA in STATE.md is stale (cited=$CITED_DEV_STATE actual=$ACTUAL_DEV)"
+  echo "FAIL: develop SHA in STATE.md is stale (cited=$CITED_DEV_STATE actual=${ACTUAL_DEV_FULL:0:${#CITED_DEV_STATE}} full=$ACTUAL_DEV_FULL)"
   FAIL=1
 fi
 if [ "$CITED_DEV_HANDOFF" != "NOT_FOUND" ] && [ "$CITED_DEV_HANDOFF" != "NO_HANDOFF" ] \
     && [ "${ACTUAL_DEV_FULL:0:${#CITED_DEV_HANDOFF}}" != "$CITED_DEV_HANDOFF" ]; then
-  echo "FAIL: develop SHA in SESSION-HANDOFF.md is stale (cited=$CITED_DEV_HANDOFF actual=$ACTUAL_DEV)"
+  echo "FAIL: develop SHA in SESSION-HANDOFF.md is stale (cited=$CITED_DEV_HANDOFF actual=${ACTUAL_DEV_FULL:0:${#CITED_DEV_HANDOFF}} full=$ACTUAL_DEV_FULL)"
   FAIL=1
 fi
 
--- a/plugins/vsdd-factory/templates/verify-sha-currency.sh
+++ b/plugins/vsdd-factory/templates/verify-sha-currency.sh
@@ -106,8 +106,14 @@ if [ ! -f "$STATE_MD" ]; then
   exit 1
 fi
 
-CITED_DEV_STATE=$(grep -oE 'develop_head: "?[0-9a-f]{8,40}' "$STATE_MD" 2>/dev/null \
-  | head -1 | grep -oE '[0-9a-f]{8,40}' | cut -c1-8 || echo "NOT_FOUND")
+# A develop_head cite may be a 7-char short SHA (git's common default) up to a
+# full 40-char SHA. Accept the whole 7..40 range and keep the cited value intact
+# — do NOT truncate. The comparison below is a prefix match against the actual
+# full develop SHA, which is correct for any short-SHA length (issue #629). The
+# prior {8,40} + `cut -c1-8` form reported NOT_FOUND for 7-char cites and then
+# silently PASSed, defeating the check on exactly the field it exists to verify.
+CITED_DEV_STATE=$(grep -oE 'develop_head: "?[0-9a-f]{7,40}' "$STATE_MD" 2>/dev/null \
+  | head -1 | grep -oE '[0-9a-f]{7,40}' | head -1 || echo "NOT_FOUND")
 
 if [ -f "$HANDOFF_MD" ]; then
   CITED_DEV_HANDOFF=$(grep -oE 'develop HEAD[^|`]*`?[0-9a-f]{8}' "$HANDOFF_MD" 2>/dev/null \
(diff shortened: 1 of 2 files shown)


Zious11 (changes requested):
# PR Review — #720 fix(hooks): accept 7-char short SHAs in verify-sha-currency develop cite

**Verdict:** REQUEST_CHANGES
**Date:** 2026-07-22
**Closes:** #629

---

## Summary

`verify-sha-currency.sh` extracted the `develop_head:` SHA with `{8,40}` and then truncated to 8 chars via `cut -c1-8`. A correct 7-char cite (`develop_head: 4c276d9`) produced `CITED_DEV_STATE=NOT_FOUND` and the currency check was silently skipped while the script exited 0 — a false negative on the field this check exists to verify. The PR fixes extraction to accept 7..40 chars, removes the truncation, and switches the comparison to a prefix match against the full actual SHA.

The fix itself is technically correct. The blocking finding is the absence of bats regression tests.

---

## Changes reviewed

**`plugins/vsdd-factory/templates/verify-sha-currency.sh`** — 2 hunks

- Hunk 1 (extraction): `{8,40}` → `{7,40}`, removes `cut -c1-8`, adds `head -1` for determinism on multi-match. The cited value is now preserved at its original length.
- Hunk 2 (comparison and WARN): Replaces `$ACTUAL_DEV` (fixed 8-char) comparison with `${ACTUAL_DEV_FULL:0:${#CITED_DEV_STATE}}`. `ACTUAL_DEV_FULL` is already defined in the script (`git -C "$PROJECT_ROOT" rev-parse develop`); the prefix slice is the git-canonical comparison for a short SHA of any length.
  - Adds WARN block: if `develop_head:` field is present in STATE.md but no hex SHA could be extracted, emits a WARN and sets `WARN=1`. Prevents the silent-skip-PASS class from recurring on malformed cites.
  - HANDOFF comparison similarly upgraded to `${ACTUAL_DEV_FULL:0:${#CITED_DEV_HANDOFF}}`. HANDOFF extraction still uses `{8}` fixed length (unchanged by this PR); the prefix comparison is effectively a no-op change for HANDOFF since the extracted value will always be 8 chars, but it is consistent.
- Exit code behavior: `WARN=1` does not affect exit code (the script exits 0 when `FAIL=0`, with or without WARNs). The WARN case for unextractable `develop_head:` correctly exits 0 rather than failing — an absent or malformed cite should surface as a warning, not block an otherwise clean burst.

---

## Findings

### F1 — MEDIUM: No bats regression tests for the fix

**Severity:** MEDIUM
**Triage routing:** test-writer

The PR author explicitly states "templates/ scripts have no bats suite in the repo (verified: no .bats file references verify-sha-currency or invokes any templates/*.sh). Red/green was done by direct invocation against a throwaway git fixture." This is acknowledged but does not satisfy the production-grade default.

The bats framework is present in `plugins/vsdd-factory/tests/` with 90+ test files, including `wave-gate-hooks.bats` which covers behavior that transitively depends on `verify-sha-currency.sh` via `validate-wave-gate-prerequisite.sh`. The infrastructure to add a `verify-sha-currency.bats` (or extend `wave-gate-hooks.bats`) exists. The specific regression scenarios are well-specified in the PR body:

1. Correct 7-char cite → exit 0, SHA extracted and matched (the primary bug case).
2. Correct 8-char cite → exit 0 (no regression).
3. Correct 40-char cite → exit 0.
4. Stale 7-char cite → exit 1, FAIL emitted.
5. `develop_head:` field present but no hex SHA extractable → exit 0, WARN emitted.

The fix closes a false-negative in a pipeline gate. Without a regression test, a future refactor of the extraction regex can silently restore the bug.

**Required action:** Add a `plugins/vsdd-factory/tests/verify-sha-currency.bats` covering at least cases 1 and 4 above (the primary bug case and its FAIL counterpart). Dispatching to test-writer.

---

### F2 — LOW: Error message shows 8-char `$ACTUAL_DEV` when comparison uses variable-length prefix

**Severity:** LOW (cosmetic; no correctness impact)

The FAIL message reads:
```
echo "FAIL: develop SHA in STATE.md is stale (cited=$CITED_DEV_STATE actual=$ACTUAL_DEV)"
```

`$ACTUAL_DEV` is the 8-char truncation; the comparison used `${ACTUAL_DEV_FULL:0:${#CITED_DEV_STATE}}`. For a 7-char cite that is stale, the message shows an 8-char actual against a 7-char cited, which can be confusing to read. A more consistent message would use `${ACTUAL_DEV_FULL:0:${#CITED_DEV_STATE}}` or just `${ACTUAL_DEV_FULL:0:8}` as context.

This does not change behavior (the FAIL is still emitted, the exit is still 1). It can be resolved in the same commit as F1.

---

## Verdict rationale

The core fix is technically correct: `ACTUAL_DEV_FULL` is properly defined, the prefix-match comparison is the git-canonical approach for variable-length short SHAs, and the WARN path for unextractable cites closes the silent-skip class. The reason for REQUEST_CHANGES is F1: a pipeline gate fix that closes a false-negative should ship with regression bats coverage. The bats infrastructure exists and the test scenarios are fully specified by the PR's own empirical evidence section.

arcavenai:
Requested changes addressed at 6d52c89e:
- **Comparable FAIL messages**: each drift line now prints `cited=… actual=…` truncated to the cited length plus `full=<40-char SHA>`, so the operator can see the mismatch at a glance without re-deriving prefixes.
- **New regression suite**: `tests/verify-sha-currency.bats` (5 scenarios: pass, stale STATE cite, stale HANDOFF cite, message-format shape, --project-root invocation) built on a real `git init -b develop` fixture.

Ready for re-review.

Zious11 (approved):
## Re-review — commit 6d52c89

### F1 (bats coverage) — ADDRESSED

`tests/verify-sha-currency.bats` adds 5 scenarios. Three are load-bearing against the pre-fix code:

- **7-char correct cite**: old `{8,40}` regex fails to extract → output shows `NOT_FOUND` → `[[ "$output" != *"NOT_FOUND"* ]]` fails. New `{7,40}` extracts correctly → passes.
- **Stale 7-char cite**: old code produces `NOT_FOUND`, comparison guard skips, exits 0. Test expects `status -eq 1` → fails against old code. New code correctly identifies drift.
- **Unextractable WARN**: old code has no WARN block → assertion on WARN text fails. New WARN path triggers correctly.

The 8-char and 40-char correct-cite tests are non-regression guards (both old and new pass), still valid as future-refactor pins. Auto-discovery confirmed: `run-all.sh` iterates `tests/*.bats`; no manual registration required.

### F2 (FAIL message format) — ADDRESSED

FAIL message now prints `actual=${ACTUAL_DEV_FULL:0:${#CITED_DEV_STATE}} full=$ACTUAL_DEV_FULL`. The stale test asserts the format directly with `[[ "$output" == *"actual=${DEV_FULL:0:7}"* ]]`.

### Sibling-site sweep (TD-VSDD-060) — clean

`{8,40}` / `cut -c1-8` appear only in the changed file. No other scripts duplicate the extraction logic.

**Approve.** Both blocking findings are fully addressed with load-bearing test coverage. No new findings.

PR #721, opened 2026-07-20
fix(hooks): scope validate-bc-title index lookup to the Title-header table

### LINKED ISSUE #566: bug(hooks/validate-bc-title): bc_h1_index_drift false positive — matcher hits first BC-id occurrence in BC-INDEX (satisfaction table), not the §2 navigation-table title column
## Summary

The `validate-bc-title` hook's `bc_h1_index_drift` check fires a false positive on every edit to a BC file when that BC's id appears in BC-INDEX **before** the §2 navigation table — e.g. in a secondary-satisfaction/capability table. The matcher appears to grab the first `BC-S.SS.NNN` occurrence in BC-INDEX and treat the adjacent cell text as the BC's indexed title.

## Environment

- vsdd-factory plugin 1.0.0-rc.22 (claude-mp marketplace), greenfield run, Phase-3 wave-9 spec burst
- Hook: `validate-bc-title`, code `bc_h1_index_drift`

## Reproduction (observed)

1. BC-INDEX.md contains `BC-2.1.001` in a capability-satisfaction table row (adjacent cell text `CAP-001`) at an earlier byte offset than the §2.1 navigation table row (which correctly carries the title "Drop-In SSH_AUTH_SOCK Replacement").
2. Edit `specs/behavioral-contracts/BC-2.1.001.md` (any edit — in our case a version bump v1.2→v1.3 that did not touch the H1 or the index row).
3. Hook fires `bc_h1_index_drift` claiming the BC's H1 ("BC-2.1.001: Drop-In SSH_AUTH_SOCK Replacement") drifts from the indexed title, reporting the indexed title as "CAP-001".

The actual §2.1 table row was in agreement with the H1 the whole time; the hook compared against the wrong table.

## Root cause (inferred)

Title extraction from BC-INDEX matches the first occurrence of the BC id anywhere in the file instead of scoping to the §2 per-subsystem navigation tables (or instead of parsing the table whose header row contains a Title column).

## Impact

- Every edit to an affected BC file emits a spurious blocking/warning validation, training agents to ignore the hook (alert-fatigue on a validator that exists to catch real H1↔index drift).
- Any BC referenced in a satisfaction/secondary table before §2 is affected — this is structural, not project-specific.

## Suggested fix

Scope the index-side title lookup to the §2 navigation table(s): either match on the table header signature (`| BC | Title | ... |`) or on the section heading, and ignore BC-id occurrences elsewhere in BC-INDEX. A regression fixture: BC-INDEX with the BC id appearing in a non-title table above §2.

## Related

Distinct from #277/#287 (real count/version propagation drift) — this is a validator matching bug producing false positives, not stale derived content.

arcavenai:
## Why

`validate-bc-title` fires a false `bc_h1_index_drift` on every edit to a BC whose id appears in BC-INDEX *before* the §2 navigation table — e.g. in a capability-satisfaction table where the adjacent cell is `CAP-001` (#566). The title-lookup awk matched the first BC-INDEX row whose 2nd pipe-field equals the BC id, with no section scoping, so it grabbed the satisfaction cell and reported "CAP-001" as the indexed title even though the §2 nav-table title agreed with the H1 the whole time. This is a validator matching bug (false positive), not stale derived content — and it trains agents to ignore a hook whose entire job is catching real H1↔index drift. Any BC referenced in a table above §2 is affected, so it's structural, not project-specific. The fix anchors the lookup to the table that actually carries titles.

## What changed

- `plugins/vsdd-factory/hooks/validate-bc-title.sh`
  - Rewrote the index-title awk to scope to the navigation table: a header row whose 2nd data column is `Title` (case-insensitive) opens a "title table"; only a BC-id row *inside* such a table is treated as authoritative. A blank/non-`|` line closes the scope. When no Title-headed table exists (headerless single-row indexes, as in the existing `canonical-format-invariant.bats` fixture), it falls back to the first-occurrence match — preserving prior behavior for those cases.

- `plugins/vsdd-factory/tests/hooks.bats` — added a `validate-bc-title` regression section (3 cases): the #566 false-positive fixture (satisfaction table with `CAP-001` before a §2 Title table that agrees with the H1) now asserts exit 0; a genuine H1-vs-nav-table drift still blocks (exit 2, `bc_h1_index_drift`); a headerless single-row index still blocks on real drift via the fallback (exit 2). The pre-existing `validate-bc-title` case in `canonical-format-invariant.bats` (headerless index, mismatched title) still passes.

## Test evidence

BC-INDEX layout under test: a capability-satisfaction table (BC-2.1.001 → `CAP-001`) precedes the §2 navigation table whose title cell correctly matches the BC's H1 ("Drop-In SSH_AUTH_SOCK Replacement").

RED — unfixed hook, edit to BC-2.1.001.md (H1 and nav row untouched):

```
exit=2
BLOCKED by validate-bc-title: BC H1 title "Drop-In SSH_AUTH_SOCK Replacement" != BC-INDEX title "CAP-001" (POLICY 7: bc_h1_is_title_source_of_truth). Fix: Update BC-INDEX title to match H1, or fix H1 if it is wrong. Code: bc_h1_index_drift.
```

GREEN — fixed hook, same fixture (lookup scopes to the §2 Title table, agrees with H1):

```
exit=0
[]
```

GREEN regression — fixed hook, nav-table title changed to "Stale Divergent Title" (a real drift) still blocks:

```
exit=2
BLOCKED by validate-bc-title: BC H1 title "Drop-In SSH_AUTH_SOCK Replacement" != BC-INDEX title "Stale Divergent Title" (POLICY 7: bc_h1_is_title_source_of_truth). Fix: Update BC-INDEX title to match H1, or fix H1 if it is wrong. Code: bc_h1_index_drift.
```

Full suites, local (macOS): `bats tests/hooks.bats` 48 ok / 0 fail; `bats tests/integration/hooks/canonical-format-invariant.bats` 19 ok / 0 fail (pre-existing validate-bc-title block case intact). `bash -n` clean. `check-bats-orphans.sh`: all hook references resolve.

Closes: #566

### DIFF
--- a/plugins/vsdd-factory/hooks/validate-bc-title.sh
+++ b/plugins/vsdd-factory/hooks/validate-bc-title.sh
@@ -66,17 +66,31 @@ if [[ ! -f "$BC_INDEX" ]]; then
   exit 0  # BC-INDEX doesn't exist yet
 fi
 
-# Extract the title for this BC from BC-INDEX
+# Extract the title for this BC from BC-INDEX.
 # Table format: | BC-S.SS.NNN | Title | ... |
+#
+# The BC id can legitimately appear in BC-INDEX more than once — e.g. in a
+# capability-satisfaction table (| BC | Satisfies | ... |) that precedes the
+# §2 navigation table. Matching the first occurrence anywhere grabbed the
+# satisfaction cell (e.g. "CAP-001") and reported it as the indexed title,
+# firing a false bc_h1_index_drift (#566). Scope the lookup to the table
+# whose header row carries a "Title" column; only that table's title cell is
+# authoritative. Fall back to the first occurrence when no Title-headed table
+# is present (headerless single-row indexes), preserving prior behavior.
 INDEX_TITLE=$(awk -F'|' -v bc="$BC_ID" '
+  function trim(s) { gsub(/^[ \t]+|[ \t]+$/, "", s); return s }
   {
-    gsub(/^[ \t]+|[ \t]+$/, "", $2)
-    if ($2 == bc) {
-      gsub(/^[ \t]+|[ \t]+$/, "", $3)
-      print $3
-      exit
+    # Non-table line ends any table scope.
+    if ($0 !~ /^[ \t]*\|/) { in_title_table = 0; next }
+    c2 = trim($2); c3 = trim($3)
+    # A header row with "Title" in the 2nd data column opens a scoped table.
+    if (tolower(c3) == "title") { in_title_table = 1; next }
+    if (c2 == bc) {
+      if (in_title_table && scoped == "") scoped = c3
+      if (fallback == "") fallback = c3
     }
   }
+  END { if (scoped != "") print scoped; else print fallback }
 ' "$BC_INDEX")
 
 if [[ -z "$INDEX_TITLE" ]]; then
--- a/plugins/vsdd-factory/tests/hooks.bats
+++ b/plugins/vsdd-factory/tests/hooks.bats
@@ -378,3 +378,71 @@ require_bash4_hook_interp() {
   run bash -c 'echo "{\"tool_input\":{\"file_path\":\".factory/STATE.md\"}}" | "'"$HOOKS"'/validate-count-propagation.sh" 2>&1'
   [ "$status" -eq 0 ]
 }
+
+# ---------- validate-bc-title ----------
+
+@test "validate-bc-title: title lookup scopes to the Title-header table (#566)" {
+  # The BC id appears first in a capability-satisfaction table (adjacent cell
+  # "CAP-001") before the §2 navigation table. The nav table title agrees with
+  # the H1, so this edit must pass — the matcher must not grab the CAP-001 cell.
+  local bc_dir=".factory/specs/behavioral-contracts"
+  cat > "$bc_dir/BC-2.01.001.md" <<'EOF'
+# BC-2.01.001: Drop-In SSH_AUTH_SOCK Replacement
+EOF
+  cat > "$bc_dir/BC-INDEX.md" <<'EOF'
+# BC-INDEX
+
+## 1. Capability Satisfaction
+
+| BC | Satisfies | Notes |
+|----|-----------|-------|
+| BC-2.01.001 | CAP-001 | primary |
+
+## 2. Navigation
+
+| BC | Title | Subsystem | Priority |
+|----|-------|-----------|----------|
+| BC-2.01.001 | Drop-In SSH_AUTH_SOCK Replacement | SS-02 | P0 |
+EOF
+  run bash -c 'echo "{\"tool_input\":{\"file_path\":\".factory/specs/behavioral-contracts/BC-2.01.001.md\"}}" | "'"$HOOKS"'/validate-bc-title.sh" 2>&1'
+  [ "$status" -eq 0 ]
+}
+
+@test "validate-bc-title: genuine H1/index drift in the nav table still blocks" {
+  local bc_dir=".factory/specs/behavioral-contracts"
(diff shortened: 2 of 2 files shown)


Zious11 (approved):
# PR #721 — Fresh-eyes PR review

**PR:** fix(hooks): scope validate-bc-title index lookup to the Title-header table
**Branch:** fix/bc-title-index-scoping → develop
**Issue:** #566
**CI:** GREEN (bats-full-suite linux, bats-darwin-leg, bats-wave-handoff, cargo-host ubuntu+macos, all build-dispatcher targets)

## Verdict: APPROVE

The core fix is correct for the realistic BC-INDEX layout and is backed by a
genuinely discriminating test: the pass/block pair share the same misleading
`CAP-001` first-occurrence cell, so they prove the scoping works rather than
just the happy path. No regression risk versus current behavior. All findings
below are non-blocking.

## What I verified (hand-trace of the awk state machine)

For `-F'|'`, a row `| BC | Title | ... |` yields `$2=" BC "`, `$3=" Title "`,
so `c3` is the 2nd *data* column — the header check `tolower(c3)=="title"` is
the right column. `scoped = c3` stores the trimmed value since `c3 = trim($3)`
runs first. Separator rows `|----|----|` start with `|`, so they do not reset
scope, and neither `c2==bc` nor `tolower(c3)=="title"` matches on the dashes —
skipped cleanly. Section headings / blank lines do not start with `|`, so
`in_title_table` closes correctly. The `scoped==""` guard means the first
title-table match wins and `END` prefers `scoped` over `fallback`, so the
preceding capability-satisfaction cell (`CAP-001`) is correctly ignored when a
nav title exists.

All three added tests trace to their asserted exit codes. No `bash -n` issue
visible; the awk is single-quoted inside `$()` so bash does not parse it, and
the awk constructs (`gsub` alternation in a function, `tolower`, `-F'|'`) are
portable across gawk and the macOS one-true-awk — consistent with CI green on
both platforms.

## Findings

### MINOR / SUGGESTION — coupling to the exact header string "Title" in a hardcoded column
Both header detection and value extraction hardcode column 3 (`c3`). The scope
only opens if the §2 nav header's 2nd data column trims to exactly `title`
(case-insensitive). If the real `BC-INDEX.md` ever uses a different header
label (e.g. "Contract Title", "Name") or places the title in a different
column, `in_title_table` never opens, `scoped` stays empty, and it silently
falls back to first-occurrence — precisely the `CAP-001` cell #566 is about.
Not a regression (degrades to today's behavior), but the fix's correctness
rests on an index-header string not visible in the diff. More robust: on the
header row, record the index of whichever cell equals "title" and read that
same column on data rows, rather than assuming column 3. At minimum, document
the header contract the hook depends on.

### MINOR — test gap
Missing: (1) a headerless-fallback *pass* case (title agrees → exit 0) to lock
in that the fallback path does not over-block; (2) a case where a Title-headed
table exists but the BC appears only in the capability table and is absent from
the nav table — `scoped` stays empty and the hook falls back to the
satisfaction cell, firing a false `bc_h1_index_drift`. Multiple-Title-table
layout is also untested (first match wins) but low value for BC-INDEX.

### MINOR — silent fallback masks a malformed-index signal
When a Title-headed table is present but the BC is not found inside it,
reporting the unrelated fallback cell as "the indexed title" produces a
confusing drift message rather than a clear "BC not present in the Title table"
signal. Consider preferring a not-indexed result over the wrong cell once a
title table has been seen. Low priority.

## Checklist
- Diff coherence, description accuracy, commit scope: consistent; description
  matches the code precisely.
- Diff size: ~90 lines, well under threshold.
- canonical-format-invariant.bats fallback exercise: outside the diff, but the
  full bats suite is green in CI, so the pre-existing test passes under the new
  scoped/fallback logic — empirically satisfied.

PR #722, opened 2026-07-20
fix(hooks): normalize leading v prefix in changelog frontmatter cross-check

### LINKED ISSUE #660: bug(hooks): validate-changelog-monotonicity does not normalize the 'v' version prefix — false-positive blocks and un-fixable frontmatter
## Summary

The `validate-changelog-monotonicity` hook string-compares the frontmatter `version:` value against the top changelog-table row without normalizing a leading `v`. It treats `"1.1"` and `"v1.1"` as unequal. This produces two distinct failures:

1. **False-positive block** — when frontmatter is bare (`"1.5"`) and the changelog table uses the `vX.Y` style (`| v1.5 | ... |`), the hook fires on every version bump even though the versions are semantically identical.
2. **Un-fixable frontmatter (the more insidious one)** — when a file lands with a v-prefixed frontmatter (`version: "v1.1"`) and a v-prefixed changelog row, the hook *passes* (both sides carry the `v`, so the naive string compare matches). But you then cannot normalize the frontmatter to the project convention (bare `"1.1"`) on a **tracked** file, because doing so creates a `"1.1"` (frontmatter) vs `"v1.1"` (changelog) mismatch that the hook flags. The convention drift is frozen in place.

## Concrete instances (one arc, switchboard-blue)

- **BC-2.03.002 v1.5** (2026-07-14): frontmatter `"1.5"` vs changelog `v1.5` → false-positive fire on an unrelated re-point edit. Landed correctly after manual override; normalized as "known quirk."
- **S-BL.ADMISSION-SYNC-WIRE-rulings.md** (2026-07-15): authored with frontmatter `version: "v1.1"` (v-prefixed, diverging from the fleet convention of bare `"1.0"`/`"1.11"`). A product-owner pass **deferred** the normalization fix specifically because normalizing `"v1.1"` → `"1.1"` while the changelog row stayed `v1.1` would trip the hook. The fix could only be applied later by a state-manager pass — and only because the file was still **untracked** (first commit → no prior committed version for the monotonicity comparison, so the internal check was bypassed). On a tracked file the same normalization would have been blocked.

## Root cause

The version comparison in the monotonicity check compares raw strings. It needs to strip a single optional leading `v`/`V` from both operands before comparing (and ideally parse as semver rather than string-compare, which would also fix `"1.10"` vs `"1.9"` ordering if that is string-compared anywhere).

## Suggested fix

Normalize both the frontmatter version and the changelog-row version with something equivalent to `s/^[vV]//` before comparison; compare as parsed semver, not as strings. This makes `"1.1"` ≡ `"v1.1"` and unblocks both the false-positive path and the convention-normalization path.

## Impact

Low severity individually (each instance is a one-off manual override), but it (a) trains agents to reflexively override the hook — eroding the signal — and (b) actively prevents a convention-conformance fix on tracked files, which is the opposite of what a governance hook should do. Recurring: 3 instances across ~2 days on one project.

Filed from switchboard-blue (project uses `S-BL.<NAME>` story ids and vsdd-factory rc.22). Layer-2 defect record per the three-layer capture rubric; no fix commitment implied.

arcavenai:
## Why

`validate-changelog-monotonicity.sh` cross-checks the frontmatter `version:`
against the top changelog-table row by raw string comparison, with no `v`-prefix
normalization. It treats `"1.5"` and `"v1.5"` as unequal, producing two failures:
a false-positive block whenever bare frontmatter meets a `vX.Y` changelog row
(fires on every version bump), and — more insidiously — an un-fixable tracked
file, because normalizing a `"v1.1"` frontmatter to the bare convention while the
changelog row stays `v1.1` trips the check, freezing the convention drift in
place. A governance hook that prevents a convention-conformance fix is doing the
opposite of its job.

## What changed

Strip a single optional leading `[vV]` from *both* operands
(`${FM_VERSION#[vV]}` / `${FIRST_VERSION#[vV]}`) before comparing. Minimal
normalization only — no full semver parsing was introduced, matching the
existing string-compare design of the ordering checks and the sibling
`validate-state-pin-freshness.sh` hook, which already normalizes v-prefixes.
`"1.5"` ≡ `"v1.5"` now, both the false-positive and the normalization-lock paths
are unblocked, and a genuine version mismatch is still caught.

Extended `tests/corpus-lint.bats` with three cases: bare-vs-v-prefixed (pass),
v-prefixed-vs-bare (pass), and a genuine mismatch after normalization (still
blocks, exit 2).

## Test evidence

Before the fix (unfixed hook), bare `"1.5"` frontmatter vs `v1.5` changelog row:

```
BLOCKED by validate-changelog-monotonicity: CHANGELOG MONOTONICITY VIOLATION in BC-vpfx.md: Frontmatter version '1.5' != top changelog version 'v1.5'. Fix: Set frontmatter 'version:' to match the top changelog row. Code: changelog_frontmatter_mismatch.
exit=2
```

New regression tests against the UNFIXED hook (red — the mismatch guard stays
green, proving it is a real check, not always-green):

```
1..3
not ok 1 changelog-monotonicity: treats bare frontmatter and v-prefixed changelog row as equal (issue #660)
not ok 2 changelog-monotonicity: treats v-prefixed frontmatter and bare changelog row as equal (issue #660)
ok 3 changelog-monotonicity: still blocks a genuine version mismatch after v-normalization (issue #660)
```

Full `corpus-lint.bats` suite against the FIXED hook (green), 89/89:

```
1..89
ok 14 changelog-monotonicity: blocks frontmatter/changelog version mismatch
ok 15 changelog-monotonicity: treats bare frontmatter and v-prefixed changelog row as equal (issue #660)
ok 16 changelog-monotonicity: treats v-prefixed frontmatter and bare changelog row as equal (issue #660)
ok 17 changelog-monotonicity: still blocks a genuine version mismatch after v-normalization (issue #660)
...
```

Closes: #660

### DIFF
--- a/plugins/vsdd-factory/hooks/validate-changelog-monotonicity.sh
+++ b/plugins/vsdd-factory/hooks/validate-changelog-monotonicity.sh
@@ -143,9 +143,17 @@ while IFS= read -r line; do
   fi
 done < "$FILE_PATH"
 
-# Cross-check frontmatter version vs top changelog row
+# Cross-check frontmatter version vs top changelog row.
+#
+# Normalize a single leading 'v'/'V' from both operands before comparing so a
+# bare frontmatter ("1.5") and a v-prefixed changelog row ("v1.5") are treated
+# as equal (issue #660). Without this, a bare-vs-prefixed pair false-positive
+# blocks on every bump, AND a v-prefixed tracked file cannot be normalized to
+# the bare convention (doing so trips this check) — freezing the drift in place.
 if [[ -n "$FM_VERSION" ]] && [[ -n "$FIRST_VERSION" ]]; then
-  if [[ "$FM_VERSION" != "$FIRST_VERSION" ]]; then
+  FM_VERSION_NORM="${FM_VERSION#[vV]}"
+  FIRST_VERSION_NORM="${FIRST_VERSION#[vV]}"
+  if [[ "$FM_VERSION_NORM" != "$FIRST_VERSION_NORM" ]]; then
     ERRORS="${ERRORS:+$ERRORS\n}Frontmatter version '$FM_VERSION' != top changelog version '$FIRST_VERSION'"
     [[ -z "$FIRST_CODE" ]] && FIRST_CODE="changelog_frontmatter_mismatch"
   fi
--- a/plugins/vsdd-factory/tests/corpus-lint.bats
+++ b/plugins/vsdd-factory/tests/corpus-lint.bats
@@ -201,6 +201,55 @@ EOF
   [[ "$output" == *"Frontmatter version"* ]]
 }
 
+@test "changelog-monotonicity: treats bare frontmatter and v-prefixed changelog row as equal (issue #660)" {
+  cat > "$WORK/.factory/specs/behavioral-contracts/BC-test.md" << 'EOF'
+---
+version: "1.5"
+---
+# BC Test
+## Changelog
+| Version | Burst | Date | Author | Change |
+|---------|-------|------|--------|--------|
+| v1.5 | pass-10 | 2026-04-20 | po | Latest |
+| v1.4 | pass-5 | 2026-04-19 | po | Prior |
+EOF
+  _run_hook validate-changelog-monotonicity.sh "$WORK/.factory/specs/behavioral-contracts/BC-test.md"
+  [ "$status" -eq 0 ]
+}
+
+@test "changelog-monotonicity: treats v-prefixed frontmatter and bare changelog row as equal (issue #660)" {
+  cat > "$WORK/.factory/specs/behavioral-contracts/BC-test.md" << 'EOF'
+---
+version: "v1.1"
+---
+# BC Test
+## Changelog
+| Version | Burst | Date | Author | Change |
+|---------|-------|------|--------|--------|
+| 1.1 | pass-3 | 2026-04-20 | po | Latest |
+| 1.0 | pass-1 | 2026-04-19 | po | Initial |
+EOF
+  _run_hook validate-changelog-monotonicity.sh "$WORK/.factory/specs/behavioral-contracts/BC-test.md"
+  [ "$status" -eq 0 ]
+}
+
+@test "changelog-monotonicity: still blocks a genuine version mismatch after v-normalization (issue #660)" {
+  cat > "$WORK/.factory/specs/behavioral-contracts/BC-test.md" << 'EOF'
+---
+version: "v1.1"
+---
+# BC Test
+## Changelog
+| Version | Burst | Date | Author | Change |
+|---------|-------|------|--------|--------|
+| v1.3 | pass-10 | 2026-04-20 | po | Latest |
+| v1.2 | pass-5 | 2026-04-19 | po | Previous |
+EOF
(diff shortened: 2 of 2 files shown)


Zious11 (approved):
# PR #722 Review — fix(hooks): normalize leading v prefix in changelog frontmatter cross-check

**Verdict: APPROVE**

Branch: `fix/changelog-v-prefix` → `develop` · Closes #660 · CI: GREEN (all matrices)

## Summary

The fix is correct, minimal, and correctly scoped. `${FM_VERSION#[vV]}` strips at
most one leading `v`/`V` (shortest-match prefix removal), applied symmetrically to
both operands, so both migration directions are handled by the same two lines. I
reviewed every changed line in both files and verified all seven review dimensions
against the actual hook source. No blocking or major issues.

- **Both-directions symmetry.** The identical `#[vV]` strip is applied to both
  `FM_VERSION` and `FIRST_VERSION`, so all three semantic pairs resolve correctly:
  `("1.5","v1.5")` equal, `("v1.1","1.1")` equal, `("v1.1","v1.3")` unequal. The
  "un-fixable tracked file" scenario (normalize frontmatter to bare while the row
  stays `v`-prefixed) is genuinely unblocked.
- **Error-message fidelity.** Keeping raw `$FM_VERSION`/`$FIRST_VERSION` in the
  message (not the normalized forms) is the correct choice. When the normalized
  comparison fires it is a genuine numeric mismatch, and echoing the user's actual
  entered strings is more actionable than echoing stripped forms.
- **FIRST_CODE.** `changelog_frontmatter_mismatch` is right — post-normalization the
  check can only fire on a real version mismatch, and the mapped recommendation
  ("Set frontmatter 'version:' to match the top changelog row") is accurate.
- **Normalization scope.** `vv1.5 → v1.5` is a non-issue; real versions never carry a
  double `v`. Single-char strip is the right granularity.
- **Minimal change.** Only the cross-check block (lines 146–152) changed;
  ordering/date logic untouched.

## Findings

### [MINOR] test-coverage — uppercase-`V` path untested (corpus-lint.bats)
Both new equality tests use lowercase `v` only. The code handles `[vV]`, but no test
exercises the uppercase-`V` path (e.g. frontmatter `V1.5` vs row `1.5`). Logic is
trivially symmetric so risk is low, but a fourth test with `version: "V1.5"` against a
bare `1.5` row asserting `status -eq 0` would lock in the `V` half of the class.

### [MINOR] scope-asymmetry (observation) — validate-changelog-monotonicity.sh:126
The frontmatter-vs-top-row check now treats `v1.5` and `1.5` as equal, but the
adjacent-row duplicate detector (`[[ "$PREV_VERSION" == "$VERSION" ]]`) still compares
raw strings. During a mixed-prefix migration, two rows that are the same semantic
version (`v1.5` then `1.5`) would slip past duplicate detection. This is a pre-existing
weakness (that check already only catches exact string equality, not true monotonic
decrease), so the PR is correctly scoped to the frontmatter cross-check and this should
NOT block. Flagged only so it's a conscious deferral rather than an oversight; worth a
follow-up if convention-migration files pass through this hook mid-migration.

## Scoping note (monotonicity ordering)
The row-ordering checks do not need v-normalization for this PR's stated goal, so
limiting the change to the frontmatter cross-check is the right call.

No BLOCKER or MAJOR findings. CI green across all matrices. Recommend merge; the two
MINOR items can ship as-is or fold into a follow-up.

PR #723, opened 2026-07-20
fix(config): register L1 product-brief path in artifact-path-registry

### LINKED ISSUE #300: bug(artifact-path-registry): no entry for L1 product-brief — create-brief output falls outside template/drift/register governance

arcavenai:
## Why

`skills/create-brief` writes the canonical L1 product brief to `.factory/specs/product-brief.md`, but the artifact-path registry has no entry that matches it. Because `validate-artifact-path` hard-blocks any `.factory/` write with no matching registry entry (`ARTIFACT_PATH_UNREGISTERED`), the brief's own canonical write is blocked — and the brief, the root of the `traces_to` chain that every downstream artifact points back to, silently falls outside template-compliance, drift-check, and register-artifact governance. This adds the missing entry so the brief is both writable and governed, locked in with a test against the real shipped registry.

## What changed

- `plugins/vsdd-factory/config/artifact-path-registry.yaml`: added a `product-brief` entry under the "Product Requirements / State" section (beside `prd`), mirroring the neighboring entries' schema exactly (`artifact_type` / `canonical_path_pattern` / `description` / `enforcement_level: block`). Pattern is the literal path create-brief documents in its § Output: `.factory/specs/product-brief.md`.
- `crates/hook-plugins/validate-artifact-path/src/tests.rs`: two tests that load the **actual production registry** (walking up from `CARGO_MANIFEST_DIR`) and assert the product-brief path returns `MatchResult::Block`, plus an end-to-end `hook_logic` check that the write is allowed (`Continue`). A helper hard-fails if the registry can't be located, so the tests can never pass vacuously.

Note: `plugins/vsdd-factory/tests/generate-registry.bats` covers the *hooks* registry generator (`hooks-registry.toml`), not `artifact-path-registry.yaml`, and the existing artifact-path-registry coverage lives entirely in the `validate-artifact-path` Rust tests — so that's where this coverage belongs, alongside the neighboring canonical-path match tests.

## Test evidence

RED (before the registry entry, tests already in place):

```
running 2 tests
test tests::test_issue_300_product_brief_hook_allows_write ... FAILED
test tests::test_issue_300_product_brief_matches_production_registry ... FAILED

---- test_issue_300_product_brief_matches_production_registry ----
assertion `left == right` failed: #300: canonical product-brief output
'.factory/specs/product-brief.md' must match a block entry in the production
registry (currently NoMatch).
  left: NoMatch
 right: Block

---- test_issue_300_product_brief_hook_allows_write ----
#300: hook_logic must Continue (allow) the write ...; got Block { reason:
"... Write to '.factory/specs/product-brief.md' under .factory/ has no
matching entry ... Code: ARTIFACT_PATH_UNREGISTERED." }

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 48 filtered out
```

GREEN (after adding the registry entry):

```
running 2 tests
test tests::test_issue_300_product_brief_hook_allows_write ... ok
test tests::test_issue_300_product_brief_matches_production_registry ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 48 filtered out
```

Full crate suite (no regressions):

```
test result: ok. 50 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out   (lib)
test result: ok. 4 passed; 0 failed; ...                                       (kani_path_matching)
test result: ok. 6 passed; 0 failed; ...                                       (proptest_registry_load)
```

Pairs with the sibling registry PR for #473 — separate entries in different sections; either merges cleanly first (verified: both patches apply to develop tip in either order, and the combined tree compiles + passes).

Closes: #300

### DIFF
--- a/crates/hook-plugins/validate-artifact-path/src/tests.rs
+++ b/crates/hook-plugins/validate-artifact-path/src/tests.rs
@@ -384,6 +384,86 @@ fn test_BC_4_11_001_ac002_matches_canonical_prd_path_returns_match() {
     }
 }
 
+// -----------------------------------------------------------------------
+// Issue #300: skills/create-brief writes the canonical L1 product brief to
+// .factory/specs/product-brief.md, but the registry had no matching entry —
+// so the brief (root of the traces_to chain) fell outside path governance
+// and its own canonical write would be blocked by validate-artifact-path
+// with ARTIFACT_PATH_UNREGISTERED.
+//
+// These tests load the ACTUAL production registry (walking up from
+// CARGO_MANIFEST_DIR) and assert the product-brief path now matches with
+// block enforcement, and that the hook allows the write end-to-end.
+//
+// Refs: #300.
+// -----------------------------------------------------------------------
+
+/// Locate the shipped production registry by walking up from
+/// CARGO_MANIFEST_DIR until `plugins/vsdd-factory/config/artifact-path-registry.yaml`
+/// is found. Layout-independent (unlike a hard-coded `../` count) so it can
+/// never silently fall back to a fixture and pass vacuously.
+fn production_registry_path_i300() -> std::path::PathBuf {
+    let manifest_dir =
+        std::env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR must be set during test");
+    let mut dir = std::path::PathBuf::from(&manifest_dir);
+    loop {
+        let candidate = dir.join("plugins/vsdd-factory/config/artifact-path-registry.yaml");
+        if candidate.exists() {
+            return candidate;
+        }
+        if !dir.pop() {
+            panic!(
+                "could not locate plugins/vsdd-factory/config/artifact-path-registry.yaml \
+                 walking up from CARGO_MANIFEST_DIR ({manifest_dir})"
+            );
+        }
+    }
+}
+
+/// Read the shipped production registry as YAML text.
+fn production_registry_yaml_i300() -> String {
+    let path = production_registry_path_i300();
+    std::fs::read_to_string(&path)
+        .unwrap_or_else(|e| panic!("failed to read registry at {}: {}", path.display(), e))
+}
+
+#[test]
+fn test_issue_300_product_brief_matches_production_registry() {
+    // #300: .factory/specs/product-brief.md is the canonical write target for
+    // skills/create-brief (SKILL.md § Output). It MUST match a block entry in
+    // the shipped registry, otherwise the brief's own write is blocked with
+    // ARTIFACT_PATH_UNREGISTERED and the L1 root of the traceability chain
+    // falls outside template/drift/register governance.
+    let registry =
+        load_registry(&production_registry_yaml_i300()).expect("production registry must load");
+    let path = ".factory/specs/product-brief.md";
+    let result = matches_canonical(path, &registry);
+    assert_eq!(
+        result,
+        MatchResult::Block,
+        "#300: canonical product-brief output '{}' must match a block entry in the \
+         production registry (currently {:?}). Without the entry, validate-artifact-path \
+         blocks the create-brief write with ARTIFACT_PATH_UNREGISTERED.",
+        path,
+        result
+    );
+}
+
+#[test]
+fn test_issue_300_product_brief_hook_allows_write() {
+    // End-to-end: a Write to the canonical product-brief path must NOT be blocked.
+    // run_logic serves the real registry YAML via the read_file callback, so
+    // this exercises the full hook path against the shipped registry.
+    let payload = make_payload("Write", Some(".factory/specs/product-brief.md"));
+    let (hook_result, _log, _events) = run_logic(payload, Ok(production_registry_yaml_i300()));
+    assert!(
+        matches!(hook_result, HookResult::Continue),
+        "#300: hook_logic must Continue (allow) the write to the canonical \
+         product-brief path; got {:?}",
+        hook_result
+    );
+}
+
 #[test]
 fn test_BC_4_11_001_ac002_matches_canonical_cycle_doc_path_returns_match() {
     let yaml = multi_entry_registry_yaml();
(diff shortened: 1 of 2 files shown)


Zious11 (approved):
# PR Review — #723 fix(config): register L1 product-brief path in artifact-path-registry

**Reviewer:** pr-reviewer (fresh-eyes, different-model cognitive diversity)
**Branch:** fix/register-product-brief → develop
**Issue closed:** #300
**CI:** GREEN (bats-full-suite linux, bats-darwin-leg macOS, bats-wave-handoff macOS, cargo-host ubuntu+macos, all build-dispatcher targets, SAST Semgrep)

## Verdict: APPROVE

Minimal, correct fix for a real governance gap. Tests lock the fix against the **real shipped registry** (not a fixture), which is the right call. No BLOCKER or MAJOR findings. Three MINOR follow-ups noted below; none block merge.

## Files reviewed (all changed files in diff)

- `plugins/vsdd-factory/config/artifact-path-registry.yaml` — new `product-brief` entry
- `crates/hook-plugins/validate-artifact-path/src/tests.rs` — two issue-#300 regression tests + helpers

## What I verified

- **Schema conformance:** New entry has all four fields (`artifact_type`, `canonical_path_pattern`, `description`, `enforcement_level`) in the same order and quoting style as the adjacent `prd` entry (path quoted, description bare). YAML parses cleanly — no colon-space in the description, so the em-dash and `skills/create-brief` slash do not create a mapping/scalar ambiguity.
- **enforcement_level: block is correct:** A *matching* block entry produces `HookResult::Continue` (write governed-and-allowed); an *unregistered* path is what triggers `ARTIFACT_PATH_UNREGISTERED`. Consistent with `prd` and other spec entries. The two tests confirm both halves: `MatchResult::Block` from `matches_canonical`, and `HookResult::Continue` from `run_logic`.
- **Placement:** `product-brief` inserted at the head of the "Product Requirements / State" section, ahead of `prd` — mirrors pipeline order (brief precedes PRD). Sensible.
- **Loop safety:** The `dir.pop()` upward walk is safe — `pop()` returns `false` at filesystem root, hitting the `panic!` branch. No infinite-loop risk. Resolution is deterministic in the monorepo (walks up from `crates/hook-plugins/validate-artifact-path` to repo root containing `plugins/vsdd-factory/config/...`).
- **Function reuse:** `load_registry` / `matches_canonical` are the same pure helpers used by neighboring `BC-4.11.001` tests; feeding them the production YAML string is appropriate and side-effect-free.

## Findings

| Severity | Category | Finding | Suggestion |
|----------|----------|---------|------------|
| MINOR | coverage | Hook-level test exercises only `make_payload("Write", …)`. `validate-artifact-path` likely also fires on `Edit`/`apply_patch`. Path-matching is tool-agnostic once the path is extracted, so `matches_canonical` covers the actual bug — this is defense-in-depth. | Add an `Edit` variant of `test_issue_300_product_brief_hook_allows_write` to guard against a future re-break if `create-brief` ever writes via `Edit`/`apply_patch`. |
| MINOR | coverage | Tests hard-panic if the shipped registry is not found in an ancestor dir. Desirable for CI (fail loud on drift) but couples the crate's unit tests to repo layout — the crate cannot be tested in a vendored/published-standalone context. | Acceptable trade-off given this is an internal workspace crate and the PR's explicit goal is to test the real registry. Flagged as a conscious decision only. |
| MINOR (informational) | dependency | Sibling registry PR (#473). No logic-level ordering dependency visible in this diff — distinct exact-match `canonical_path_pattern`s, no overlapping globs in the hunk, so first-match precedence is not a concern here. Only residual risk is a textual git conflict if #473 also inserts near the "Product Requirements / State" header. | Cannot confirm without #473's diff; would be a trivial rebase, not a semantic problem. "Applies cleanly in either order" is plausible. |

## Out-of-scope / could-not-verify (information wall)

- Exact write mechanism used by `skills/create-brief` (Write vs Edit/apply_patch).
- Whether `product-brief` `artifact_type` collides with an existing entry elsewhere in the full registry file.
- Whether any earlier broad-glob entry would shadow this exact path under first-match precedence.

All three are low-risk given the green `matches_canonical` assertion against the real registry. A follow-up could confirm.

Issue #724, opened 2026-07-20
bug(ci+tests): bats-full-suite fails every PR — live sprint-state.yaml assertions gate unrelated PRs on base artifact drift (PC3 def-b violation at factory-artifacts tip)


arcavenai:
## Observable

`bats-full-suite (linux)` fails on **every currently-open PR** (#714–#723) with one identical failure, and reproduces locally against factory-artifacts `27c37d11`:

```
not ok 14 test_partitions_sorted_by_full_graph_depth_def_b
FAIL (BC-5.41.004 PC3 / ADR-026 §Decision 3a def-b): Partition B intra-partition ordering violations:
FAIL PartB ord 158 S-5.07 (depth=14) -> ord 159 S-21.01 (depth=1): depth decreases
```

CI evidence: e.g. run 29760563515 job 88413844881 (PR #714) — same single failure on all ten.

## Root cause A — data: sprint-state.yaml violates PC3 def-b

`stories/sprint-state.yaml` (factory-artifacts `27c37d11`) has the five E-21 stories **appended at the end of Partition B** (ordinals 159–163: S-21.01…S-21.05). STORY-INDEX.md `depends_on` for all five is `[]`, so full-graph wave-depth = 1 — under BC-5.41.004 PC3 v1.4 (def-b, ADR-026 §Decision 3a, human directive "full rigor option b") they belong in Partition B's depth-1 lexicographic group, not after depth-14 stories. Timeline suggests the D-861/D-862/D-865 tri-surface status-flip bursts appended rather than insert-sorted (STATE.md D-866 records the pipeline paused with bookkeeping reconciliation deferred, so this may already be on your radar).

**Fix A (yours; one commit on factory-artifacts):** re-sort Partition B per def-b — moves the five S-21 rows to the depth-1 group. `bats -f partitions_sorted_by_full_graph_depth plugins/vsdd-factory/tests/sprint-state-format.bats` goes green the moment the order is restored.

## Root cause B — gate design: live-artifact assertions run in PR CI, where they can only false-attribute

[...]

PR #725, opened 2026-07-20
fix(tests): skip live sprint-state.yaml assertions under CI — restore documented local-only design

arcavenai:
## Fix: sprint-state-format.bats live-artifact assertions under CI (D-865)

**Finding:** \`sprint-state-format.bats\` live sprint-state.yaml assertions (\`test_real_production_file_completeness_and_status_fidelity\`, \`test_partitions_sorted_by_full_graph_depth_def_b\`, and related) unconditionally access factory-artifacts data. In \`pull_request\` CI runs, the runner mounts the BASE repo's factory-artifacts (data a PR cannot modify), so these tests fail on every PR regardless of the PR's own changes. This restores the documented local-only design.

**Severity:** LOW (test-infra only; production hook chain unaffected)

### What Changed

Added \`_skip_live_artifact_in_ci\` guard function to \`plugins/vsdd-factory/tests/sprint-state-format.bats\` with 5 callsites covering all tests that access the live \`_PRODUCTION_SPRINT_STATE\` factory-artifacts file.

Guard keys on \`GITHUB_EVENT_NAME == "pull_request"\` — the correct narrow signal:
- \`pull_request\` runs: skip cleanly (BASE factory-artifacts mount; PR cannot modify)
- \`push\` runs on develop: full enforcement maintained (canary for data drift)
- Local runs (env var unset): full enforcement maintained

### Why

Resolves persistent \`pull_request\` CI failures on any PR that touches develop when the live sprint-state.yaml data has drift. The \`pull_request\`-scoped guard is more production-grade than a blanket \`GITHUB_ACTIONS\` skip because it preserves the live canary on develop push runs.

Related: #724

### Testing

- [x] CI: all 13 checks green including \`bats-full-suite (linux)\`
- [x] Guard function fires on \`pull_request\` events; local and push runs fall through
- [x] 5 callsites cover all live-data-accessing tests in scope

### DIFF
--- a/plugins/vsdd-factory/tests/sprint-state-format.bats
+++ b/plugins/vsdd-factory/tests/sprint-state-format.bats
@@ -161,14 +161,15 @@ teardown() {
 # concrete instance: a PC3 def-b ordering violation in sprint-state.yaml failed
 # ten unrelated PRs at once).
 #
-# This guard restores the documented local-only design explicitly: skip under
-# GitHub Actions, keep full enforcement in local runs (and in the factory's
-# own hooks). If push-run enforcement is wanted, narrow the condition to
-# [ "${GITHUB_EVENT_NAME:-}" = "pull_request" ].
+# This guard skips only pull_request runs, where the base-mount mismatch
+# applies. Push runs on develop keep these tests live as a canary: there CI
+# mounts the artifact state develop actually ships, so a real sprint-state
+# corruption still surfaces on the push run. Local runs and the factory's
+# own hooks keep full enforcement either way.
 # ---------------------------------------------------------------------------
 _skip_live_artifact_in_ci() {
-  if [ -n "${GITHUB_ACTIONS:-}" ]; then
-    skip "live .factory artifact assertions are local-only by design (PORTABILITY notes; issue #724) — CI mounts the base repo's factory-artifacts, which a PR cannot modify"
+  if [ "${GITHUB_EVENT_NAME:-}" = "pull_request" ]; then
+    skip "live .factory artifact assertions skip on pull_request runs (PORTABILITY notes; issue #724) — CI mounts the base repo's factory-artifacts, which a PR cannot modify; push runs keep this canary"
   fi
 }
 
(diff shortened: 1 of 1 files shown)


arcaven:
This cleanly unblocks the batch, and it's good that you already flagged the narrower `GITHUB_EVENT_NAME = pull_request` variant. One soft argument for making that the default: keying on `GITHUB_ACTIONS` also drops these five live-data checks from push runs on `develop`, where CI mounts the artifact state develop actually ships — so a real sprint-state.yaml corruption (as opposed to base-mount-vs-PR drift) would stop surfacing there too. The `pull_request`-only guard fixes the false-fail on unrelated PRs while keeping that push-run canary; local and hook enforcement stay intact either way.

arcavenai:
Agreed and adopted in 7d1d3961 — the guard now keys on `GITHUB_EVENT_NAME = pull_request`, keeping the five live-data checks as a push-run canary on develop. Verified all three modes locally: pull_request-sim skips exactly 5, push-sim runs them (and currently fails on the real PC3 def-b violation in live sprint-state.yaml — which is the canary doing its job), local unchanged. One consequence worth naming: develop push runs stay red on this suite until the #724 data re-sort lands; that red is a true report of the live violation, not noise.

arcavenai:
The red on the latest run is unrelated to this PR and looks like a flake: the guard itself works exactly as intended — all five live-artifact tests skip with the pull_request message (including `def_b`, previously failing the whole batch). The single failure is `precompact-routing.bats` TC-EC001 (two PreCompact plugins, block propagation exit code), which (a) passed at 18:17–18:18 UTC on two sibling PR runs against the identical develop base and dispatcher code, (b) is untouched by this diff (bats-only, one file), and (c) passes 5/5 consecutive local runs against a fresh release build of develop's dispatcher. The test's shape — spawn two plugins and assert exit-code propagation — is timing-sensitive on a loaded runner. Re-triggering via close/reopen to confirm. (Separately: the warning-level `git exit 128` annotation on this and every job traces to a dangling gitlink committed to develop — filed as #741 with a local repro.)

Zious11 (approved):

PR #726, opened 2026-07-20
fix(factory-obs+emit-event): correct JSON smoke-test to key=value and leave a breadcrumb when a JSON blob is dropped

### LINKED ISSUE #296: bug(factory-obs+emit-event): documented smoke-test passes JSON to emit-event, which silently drops the payload (key=value only)
## Summary

The `factory-obs` skill (`skills/factory-obs/SKILL.md:100`) documents a smoke-test command that passes a **JSON string** to `emit-event`:

```bash
"${CLAUDE_PLUGIN_ROOT}/bin/emit-event" '{"type":"hook.action","hook":"manual-test","action":"smoke"}'
```

But `bin/emit-event` does not accept JSON. Its contract (per its own header comment, lines 14–21) is **`key=value` positional pairs**:

```bash
emit-event type=hook.block hook=destructive-command-guard reason=rm_root
```

The arg-parsing loop skips any argument without an `=` (lines 90–93). A JSON blob like `{"type":"hook.action",...}` contains no top-level `=`, so **every field is silently dropped**. The script still writes a line and exits 0 — so it *looks* like it worked.

## Reproduction

```console
$ emit-event '{"type":"hook.action","hook":"manual-test","action":"smoke"}'
$ tail -1 .factory/logs/events-$(date +%F).jsonl
{"ts":"2026-06-26T23:16:17-0500","ts_epoch":1782533777,"schema_version":1}
```

Note `type`, `hook`, `action` are **all absent** — only the auto-injected fields survive.

Correct invocation:

```console
$ emit-event type=hook.action hook=manual-test action=smoke
$ tail -1 .factory/logs/events-$(date +%F).jsonl
{"type":"hook.action","hook":"manual-test","action":"smoke","ts":"...","ts_epoch":...,"schema_version":1}
```

## Why this is more than a doc typo

`emit-event` is deliberately failure-tolerant — **exit 0 on every path, no stderr** (lines 8–12, 44–45). That's correct for hook instrumentation (telemetry must never break the pipeline), but it means the wrong syntax produces **zero feedback**: no error, a written line, exit 0. An operator running the documented smoke test sees the event file grow and a clean exit, and reasonably concludes ingestion is healthy — when in fact the event is empty. This actively *masks* the mistake the docs lead them into.

It also defeats the smoke test's purpose. The point of the smoke test is to verify the named event (`hook=manual-test`) appears in Grafana. With the payload stripped, you can't distinguish your test event from any other bare heartbeat line, so the verification step is meaningless.

## Fix

1. **Doc fix (primary):** correct `skills/factory-obs/SKILL.md:100` to the `key=value` form:
   ```bash
   "${CLAUDE_PLUGIN_ROOT}/bin/emit-event" type=hook.action hook=manual-test action=smoke
   ```
   I grepped the rest of the shipped plugin — this is the **only** JSON-form invocation; all other docs (e.g. `create-adr/SKILL.md`, `factory-lock`, `factory-unlock`) already use `key=value` correctly. So it's a one-line doc fix.

2. **Hardening (optional, recommended):** `emit-event` could detect an arg that starts with `{` / looks like JSON and is the sole arg, and either (a) parse it as a JSON object, or (b) record a `_emit_parse_warning` field in the emitted event. Option (b) preserves the exit-0 contract while leaving a breadcrumb for the next operator, without changing the documented interface. Given the silent-by-design contract, a self-documenting `--help`/usage path (currently `emit-event --help` prints nothing and exits 0) would also help — though that arguably belongs in a separate enhancement.

## Acceptance criteria

- [ ] `skills/factory-obs/SKILL.md` smoke test uses `key=value` syntax and produces an event containing `hook=manual-test`.
- [ ] (optional) `emit-event` leaves a detectable signal when handed an arg with no `=`, rather than silently dropping it.

## Found during

`/vsdd-factory:factory-obs register` + smoke test on `aae-orc/akey`, `vsdd-factory@1.0.0-rc.21`, darwin-arm64. The smoke test "passed" (file grew, exit 0) but Grafana content search for the event turned up nothing, which is what surfaced the dropped payload.

## Related

- #208 (the smoke test is the verification step in the register-while-running golden path; a smoke test that silently emits empty events compounds that UX gap)

arcavenai:
## Why

The `factory-obs` skill documents a smoke test that passes a JSON string to `emit-event`, but `emit-event` only parses `key=value` positional pairs — a JSON blob has no top-level `=`, so every field is silently dropped while the script still writes a line and exits 0. An operator running the documented smoke test sees the event file grow and a clean exit and reasonably concludes ingestion is healthy, when the event is actually empty. This is the only JSON-form invocation in the shipped plugin; every other doc already uses `key=value`. The one-line doc fix stops leading operators into the footgun, and a payload-preserving breadcrumb makes the drop detectable for anyone who hits it another way.

## What changed

- `skills/factory-obs/SKILL.md`: the smoke test now uses `type=hook.action hook=manual-test action=smoke` (the correct `key=value` form). The emitted event now carries the named fields, so the Grafana verification step is meaningful again.
- `bin/emit-event`: when an argument has no `=` **and** looks like a JSON object/array (`{`/`[` prefix), the emitter records an `_emit_parse_warning` field in the event it writes. This is the issue's option (b): it preserves the tool's inviolable contract (exit 0 on every path, no stdout/stderr on success, only ever writes the daily log) and matches the tool's existing self-instrumenting style (it already auto-injects `ts`/`schema_version`/`session_id`). A stderr warning was deliberately NOT chosen — the header contract at lines 8-12 promises "No output on stdout or stderr on success," and breaking that could break the hooks that call `emit-event`. Ordinary no-`=` garbage args remain silently tolerated (unchanged); only JSON-looking tokens are flagged, since those are almost certainly a caller who mistook the tool for a JSON sink.
- `tests/emit-event.bats`: 6 new tests (44 → 50 `@test` blocks) following the existing drop-path pattern — JSON-form still exits 0, produces no stdout/stderr, leaves the `_emit_parse_warning` breadcrumb (object and array forms), and confirms the dropped payload; plus two regression guards (normal `key=value` call has no breadcrumb and its fields survive; plain non-JSON garbage does not trigger the breadcrumb).

## Test evidence

RED — the documented JSON-form command drops the payload (before the doc fix):

```
$ VSDD_LOG_DIR=$TMP ./bin/emit-event '{"type":"hook.action","hook":"manual-test","action":"smoke"}'
exit=0
--- emitted line ---
{"ts":"2026-07-20T12:20:24-0500","ts_epoch":1784568024,"schema_version":1}
--- does it contain hook=manual-test? ---
ABSENT
--- fields present ---
schema_version,ts,ts_epoch
```

GREEN (a) — corrected `key=value` smoke test carries the fields:

```
$ VSDD_LOG_DIR=$TMP ./bin/emit-event type=hook.action hook=manual-test action=smoke
exit=0
{"type":"hook.action","hook":"manual-test","action":"smoke","ts":"...","ts_epoch":...,"schema_version":1}
--- hook field: manual-test ---
```

GREEN (b) — JSON-form now leaves a breadcrumb, still exit 0, still no stderr:

```
$ STDERR=$(VSDD_LOG_DIR=$TMP ./bin/emit-event '{"type":"hook.action",...}' 2>&1 >/dev/null)
exit=0 stderr=''
{"ts":"...","ts_epoch":...,"schema_version":1,"_emit_parse_warning":"arg dropped: expected key=value pairs, got a JSON-looking token (see emit-event usage)"}
--- _emit_parse_warning: arg dropped: expected key=value pairs, got a JSON-looking token (see emit-event usage) ---
```

Regression — normal call and plain garbage do NOT trigger the breadcrumb:

```
$ ./bin/emit-event type=test hook=x        → warn=ABSENT
$ ./bin/emit-event xxx yyy type=test       → warn=ABSENT  hook_survives=test
```

Full bats suite (macOS, bash 5.3, jq /usr/bin, bats 1.x) — 50 pass, 1 skip (jq-missing sim not possible when jq is in /usr/bin):

```
$ bats tests/emit-event.bats
ok 1..50 (test 21 skipped: jq installed in /usr/bin)
```

Closes: #296

### DIFF
--- a/plugins/vsdd-factory/bin/emit-event
+++ b/plugins/vsdd-factory/bin/emit-event
@@ -86,9 +86,18 @@ fi
 jq_args=()
 filter='{}'
 idx=0
+saw_jsonish_arg=0
 for arg in "$@"; do
   case "$arg" in
     *=*) ;;
+    # Breadcrumb for the #296 footgun: an arg with no `=` is silently
+    # dropped (this is deliberate garbage-tolerance). But an arg that
+    # *looks like* a JSON object/array is almost certainly a caller who
+    # mistook this tool for a JSON sink — the entire payload vanishes with
+    # no signal. Flag it in the emitted event (never stdout/stderr — the
+    # exit-0/silent contract above is inviolable) so the next operator has
+    # a trail instead of a bare heartbeat line.
+    \{*|\[*) saw_jsonish_arg=1; continue ;;
     *) continue ;;   # skip args without =
   esac
   key="${arg%%=*}"
@@ -115,6 +124,15 @@ TS_EPOCH=$(date +%s 2>/dev/null) || exit 0
 
 jq_args+=(--arg _ts "$TS" --arg _ts_epoch "$TS_EPOCH")
 filter="${filter} | .ts = \$_ts | .ts_epoch = (\$_ts_epoch | tonumber) | .schema_version = 1"
+
+# --- Breadcrumb: caller passed a JSON-looking arg with no key=value pair ----
+# See the arg-parsing loop above. Records that the payload was dropped, so an
+# operator running the (historically JSON-form) smoke test isn't fooled by a
+# bare heartbeat line into thinking ingestion is healthy. Preserves the
+# exit-0/no-stderr contract — the signal rides in the event itself.
+if [ "$saw_jsonish_arg" -eq 1 ]; then
+  filter="${filter} | ._emit_parse_warning = \"arg dropped: expected key=value pairs, got a JSON-looking token (see emit-event usage)\""
+fi
 
 # --- Auto-inject session_id from env if caller didn't provide one ----------
 # Priority order: caller-provided key (already in filter) > VSDD_SESSION_ID
--- a/plugins/vsdd-factory/skills/factory-obs/SKILL.md
+++ b/plugins/vsdd-factory/skills/factory-obs/SKILL.md
@@ -97,7 +97,7 @@ After `up`, it's helpful to remind the user that events only appear once a
 factory hook fires. A quick smoke test:
 
 ```bash
-"${CLAUDE_PLUGIN_ROOT}/bin/emit-event" '{"type":"hook.action","hook":"manual-test","action":"smoke"}'
+"${CLAUDE_PLUGIN_ROOT}/bin/emit-event" type=hook.action hook=manual-test action=smoke
 ```
 
 Events should be visible in Grafana within ~10 seconds.
--- a/plugins/vsdd-factory/tests/emit-event.bats
+++ b/plugins/vsdd-factory/tests/emit-event.bats
@@ -56,6 +56,63 @@ _logfile() {
   [ "$status" -eq 0 ]
 }
 
+# ---------- JSON-form footgun breadcrumb (#296) ----------
+# emit-event takes key=value pairs, not JSON. A JSON blob has no top-level `=`,
+# so every field is silently dropped. When the sole/leading token looks like a
+# JSON object or array, leave a breadcrumb IN THE EVENT (never stdout/stderr —
+# the silent/exit-0 contract is inviolable) so the drop is detectable.
+
+@test "emit-event: JSON-form arg still exits 0 (contract preserved)" {
+  run "$HELPER" '{"type":"hook.action","hook":"manual-test","action":"smoke"}'
+  [ "$status" -eq 0 ]
+}
+
+@test "emit-event: JSON-form arg produces no stdout/stderr" {
+  run "$HELPER" '{"type":"hook.action"}'
+  [ "$status" -eq 0 ]
+  [ -z "$output" ]
+}
+
+@test "emit-event: JSON-object arg leaves _emit_parse_warning breadcrumb" {
+  run "$HELPER" '{"type":"hook.action","hook":"manual-test","action":"smoke"}'
+  [ "$status" -eq 0 ]
+  local f
+  f=$(_logfile)
+  [ -n "$f" ]
+  [ "$(jq -r '._emit_parse_warning // "ABSENT"' < "$f")" != "ABSENT" ]
+  # The JSON payload's fields were still dropped (the breadcrumb documents that).
+  [ "$(jq -r '.hook // "ABSENT"' < "$f")" = "ABSENT" ]
+}
+
+@test "emit-event: JSON-array arg also leaves the breadcrumb" {
+  run "$HELPER" '["type","hook.action"]'
+  [ "$status" -eq 0 ]
+  local f
+  f=$(_logfile)
+  [ "$(jq -r '._emit_parse_warning // "ABSENT"' < "$f")" != "ABSENT" ]
+}
+
+@test "emit-event: normal key=value call has NO breadcrumb" {
+  run "$HELPER" type=hook.action hook=manual-test action=smoke
+  [ "$status" -eq 0 ]
+  local f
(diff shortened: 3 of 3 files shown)


Zious11 (approved):
# PR Review — #726

`fix(factory-obs+emit-event): correct JSON smoke-test to key=value and leave a breadcrumb when a JSON blob is dropped`

**Verdict: APPROVE**

Fresh-eyes review against diff, PR description, and inlined test evidence. No BLOCKING or MAJOR findings.

## Checklist verification

1. **Diff coherence** — PASS. All three files (`skills/factory-obs/SKILL.md`, `bin/emit-event`, `tests/emit-event.bats`) serve the single #296 fix. No unrelated changes.
2. **Description accuracy** — PASS (one minor discrepancy, see findings). PR body matches the diff.
3. **Test coverage** — PASS. New drop-path, breadcrumb (object + array), and two regression guards target the changed lines.
4. **Demo evidence** — PASS (proportionate). RED / GREEN(a) / GREEN(b) / regression transcripts inlined in the PR body cover both failure and fixed paths. Acceptable for a CLI-tool doc+breadcrumb change of this size.
5. **Commit quality** — PASS. Conventional `fix(...)` subject; `Closes: #296`.
6. **Diff size** — PASS. ~90 lines, well under the 500-line threshold.
7. **Missing changes** — PASS. Doc fix + breadcrumb + tests fully address the stated issue.
8. **Dependency status** — PASS. Self-contained; Class 0, develop-tracked files only; no upstream PR gating.

## Findings

| Severity | Category | Finding | Suggestion |
|----------|----------|---------|------------|
| ADVISORY | correctness/edge-case | The JSON-detection branch (`\{*\|\[*`) is only reached for args with no `=`, because `*=*` is matched first in the `case`. A caller passing a JSON blob whose string values contain `=` (e.g. `'{"cmd":"a=b"}'`) matches `*=*` instead, gets no breadcrumb, and silently creates a garbage key (`{"cmd":"a` = `b"}`). Honest best-effort limitation — the documented footgun (JSON with `:` separators, no top-level `=`) is fully covered, and `*=*`-first ordering is pre-existing behavior. Non-blocking. | Add a one-line comment noting the detection is heuristic, not exhaustive. |
| ADVISORY | description | PR body says "7 new tests" but the displayed diff shows 6 `@test` blocks (hunk header claims +63 lines, so the diff view appears truncated). | Confirm the 7th "confirms the dropped payload" test is present on the branch. |
| NIT | coverage | No single test exercises the mixed case (one JSON-ish arg + valid `key=value` pairs) proving valid fields survive AND the breadcrumb fires together. | Cheap to add; not required. |

## Notes

- Contract reasoning is sound: the change adds a field to the event JSON only, never touching stdout/stderr, preserving the "silent on success, exit 0, writes only the daily log" header contract. The `continue` correctly prevents the JSON-ish token from being parsed as a `key=value` pair.
- Choosing the in-event breadcrumb over a stderr warning is the right call given hook callers depend on the silent contract.
- Case-glob escaping (`\[*`) correctly makes `[` literal rather than a bracket expression.

Approving: well-scoped, correctly reasoned, guards both the new behavior and against regressions.

Zious11:
Verified on branch: 6 new @test blocks (44 on develop → 50 on branch). PR body says '7 new tests' — please correct the description (no code change needed).

arcavenai:
Corrected — thanks for the count. The body now reads **6 new tests (44 → 50 `@test` blocks)**, verified against the merge commit: `git show 8a397904^:...emit-event.bats | grep -c '^@test'` → 44, same on `8a397904` → 50.

PR #727, opened 2026-07-20
fix(tooling): resolve .factory worktree cwd-independently in factory-cas-push.sh

### LINKED ISSUE #631: bug(tooling): factory-cas-push.sh assumes repo-root cwd — internal 'cd .factory' fails when invoked from inside the factory worktree
## Summary

`bin/factory-cas-push.sh` (shipped in the plugin, introduced with the single-writer lock/lease work in #170) assumes it is invoked from the **repo root**: it performs an internal `cd .factory` (relative path) before pushing. When the invoking agent's working directory is already the `.factory` worktree — which is the *natural* cwd for a state-manager-role agent whose entire job is inside that worktree — the script fails with:

```
cannot change to '.factory': No such file or directory
```

## Environment interaction that makes this worse

Agent shell threads in this harness do not reliably preserve `cd` state between bash calls; we've observed a thread's cwd silently reset to its original working directory mid-task (we reported a sibling-worktree variant of this in a #457 comment). For an agent whose thread cwd IS `.factory`, every reset lands it back in exactly the directory where the script's relative `cd` breaks. In one session we hit the failure twice; both recovered on first retry by prefixing `cd <repo-root> &&` in the same bash call — but that depends on the agent diagnosing the failure correctly rather than concluding the push tooling is broken.

## Proposed fix

Make the script cwd-independent:
- Resolve the repo root internally — e.g. `git rev-parse --show-toplevel` (with the caveat that when run from inside the `.factory` worktree this returns the worktree root, so either detect "already in the factory worktree" and skip the `cd`, or resolve via `git worktree list` / walking up for the expected layout), and
- fail with an explicit, actionable message ("run from repo root or worktree; could not locate factory worktree from $PWD") rather than the raw `cd` error.

## Metrics

2 failures in one session, 2/2 recovered first-try with an explicit repo-root prefix; 0 lost work. Root cause diagnosed by the affected agent itself on the second occurrence.

arcavenai:
## Why

`bin/factory-cas-push.sh` ran a relative `git -C .factory` throughout, assuming it was always invoked from the repo root. When the caller's working directory is the `.factory` worktree — the natural cwd for a state-manager-role agent whose entire job lives inside that worktree — the relative path resolved to `.factory/.factory` and the script died with `cannot change to '.factory': No such file or directory`. Worse, that raw `cd` failure surfaced through the fetch step and got reported as `state-burst CAS push failed — fetch error ... Retry after resolving network`, sending the operator down a network rabbit-hole for a cwd bug. This makes the CAS-push tooling reliable from the one cwd where it is most likely to be run.

## What changed

- `bin/factory-cas-push.sh`: added a cwd-independent `.factory` resolution block and routed all four git steps (fetch, rev-parse, cat-file, push) through the resolved `$FACTORY_DIR`. Resolution order:
  1. `cwd/.factory` exists → use it directly (repo-root invocation; relative, byte-identical to the historical behavior, no git call).
  2. else derive `<main-worktree>/.factory` from `git worktree list --porcelain` (first entry is always the main worktree, from any worktree). This covers the caller standing inside `.factory` (where `cwd/.factory` would be `.factory/.factory`, absent) and callers in sibling story worktrees.
  3. else fail with an actionable message (`could not locate the .factory worktree from <pwd>. Run from the repo root or the .factory worktree.`) instead of the raw `cd` error.

  This mirrors the repo's own convention: emit-event, factory-query, factory-sla, and factory-dashboard already locate the factory dir via the same `git worktree list --porcelain` first-entry idiom. Checking `cwd/.factory` first is deliberate — it's the fast path, keeps repo-root behavior identical, and avoids invoking git in the common case.
- `tests/factory-cas-push.bats`: 4 new tests using a real bare-origin + main-worktree + linked `.factory` fixture (no git stub): repo-root succeeds, inside-`.factory` succeeds (and does NOT emit the old `cannot change to '.factory'` error), sibling story worktree succeeds, and an unrelated non-git dir fails with the actionable message.

## Test evidence

RED — pristine upstream/develop script, run from inside `.factory` and from an unrelated dir:

```
===== ORIGINAL from INSIDE .factory =====
fatal: cannot change to '.factory': No such file or directory
state-burst CAS push failed — fetch error before push. Retry after resolving network.
exit=1
===== ORIGINAL from unrelated dir =====
fatal: cannot change to '.factory': No such file or directory
state-burst CAS push failed — fetch error before push. Retry after resolving network.
exit=1
```

GREEN — fixed script across all four cwd scenarios (real bare-origin + main + linked `.factory` fixture):

```
===== from repo root =====
state-burst CAS push succeeded (SHA: 490a1ec...)
exit=0
===== from INSIDE .factory =====
state-burst CAS push succeeded (SHA: 490a1ec...)
exit=0
===== from a sibling story worktree =====
state-burst CAS push succeeded (SHA: 5dbf574...)
exit=0
===== from an unrelated non-git dir =====
state-burst CAS push failed — could not locate the .factory worktree from /tmp/... . Run from the repo root or the .factory worktree.
exit=1
```

Full bats suite (macOS, bash 5.3, bats 1.x) — the 5 pre-existing CAS-push tests plus the 4 new cwd-resolution tests all pass:

```
$ bats tests/factory-cas-push.bats
1..9
ok 1 test_BC_5_40_001_cas_push_rejected_on_concurrent_write
ok 2 test_BC_5_40_001_cas_push_stale_sha_after_fetch
ok 3 test_BC_5_40_001_cas_push_object_absent_after_fetch
ok 4 test_BC_5_40_001_fetch_failure_aborts_push
ok 5 test_BC_5_40_001_verify_git_push_hook_unchanged
ok 6 cwd-resolution: invocation from repo root succeeds (#631)
ok 7 cwd-resolution: invocation from INSIDE .factory worktree succeeds (#631)
ok 8 cwd-resolution: invocation from a sibling story worktree succeeds (#631)
ok 9 cwd-resolution: invocation from an unrelated non-git dir fails clearly (#631)
```

The 4 new tests were confirmed genuinely RED against the pristine upstream script (test 7 fails on exit 1 with the raw cd error; test 9's actionable-message assertions fail because the original emits the generic fetch-error message) — not tautological.

Closes: #631

### DIFF
--- a/plugins/vsdd-factory/bin/factory-cas-push.sh
+++ b/plugins/vsdd-factory/bin/factory-cas-push.sh
@@ -7,19 +7,21 @@
 # Usage:
 #   factory-cas-push.sh
 #
-# No arguments. Operates on the .factory/ worktree relative to the caller's working
-# directory (consistent with state-manager burst discipline). Requires git >= 2.6 for
+# No arguments. Operates on the .factory/ worktree. The worktree is located
+# cwd-independently (see FACTORY_DIR resolution below): it works from the repo
+# root, from inside the .factory worktree itself (the state-manager's natural
+# cwd — issue #631), and from a sibling story worktree. Requires git >= 2.6 for
 # the --force-with-lease=<refname>:<sha> explicit form (ADR-025 Decision 8).
 #
 # Sequence:
-#   Step 1: git -C .factory fetch origin factory-artifacts
+#   Step 1: git -C "$FACTORY_DIR" fetch origin factory-artifacts
 #           On non-zero exit → print AC-010 fetch-failure error to stderr; exit non-zero.
 #           The push MUST NOT proceed with a potentially stale EXPECTED_SHA.
 #
-#   Step 2: EXPECTED_SHA=$(git -C .factory rev-parse origin/factory-artifacts)
+#   Step 2: EXPECTED_SHA=$(git -C "$FACTORY_DIR" rev-parse origin/factory-artifacts)
 #           Capture the expected remote tip SHA immediately after fetch.
 #
-#   Step 3: git -C .factory push --force-with-lease=factory-artifacts:"${EXPECTED_SHA}" \
+#   Step 3: git -C "$FACTORY_DIR" push --force-with-lease=factory-artifacts:"${EXPECTED_SHA}" \
 #               origin factory-artifacts
 #           On non-zero exit (--force-with-lease check failed = concurrent write detected) →
 #           print AC-005 CASPushRejected error to stderr; exit non-zero.
@@ -41,11 +43,48 @@
 set -euo pipefail
 
 # ---------------------------------------------------------------------------
+# Resolve the .factory worktree cwd-independently (issue #631).
+#
+# The former relative `git -C .factory` assumed the caller was standing in the
+# repo root. When the caller's cwd IS the .factory worktree — the natural cwd
+# for a state-manager-role agent whose whole job lives there — the relative
+# `.factory` resolved to `.factory/.factory` and every git step died with
+# "cannot change to '.factory': No such file or directory", which then got
+# reported as a fetch error, sending the operator down a network rabbit-hole.
+#
+# The .factory worktree is always `<main-worktree>/.factory` (created via
+# `git worktree add .factory factory-artifacts`). The main worktree is the
+# first entry of `git worktree list --porcelain` regardless of which worktree
+# we invoke from — this is the same idiom emit-event, factory-query,
+# factory-sla, and factory-dashboard already use to locate the factory dir.
+#
+# Resolution order:
+#   1. cwd/.factory exists → repo-root invocation. Use it directly (relative,
+#      byte-identical to the historical behavior; no git call needed).
+#   2. else derive <main-worktree>/.factory from `git worktree list`. Covers
+#      the caller standing INSIDE .factory (where cwd/.factory would be
+#      .factory/.factory — absent) and callers in sibling story worktrees.
+#   3. else fail with an actionable message (not the raw `cd` error).
+# ---------------------------------------------------------------------------
+
+if [ -d ".factory" ]; then
+  FACTORY_DIR=".factory"
+else
+  MAIN_WT="$(git worktree list --porcelain 2>/dev/null | awk '/^worktree /{print $2; exit}')" || MAIN_WT=""
+  if [ -n "$MAIN_WT" ] && [ -d "$MAIN_WT/.factory" ]; then
+    FACTORY_DIR="$MAIN_WT/.factory"
+  else
+    printf 'state-burst CAS push failed — could not locate the .factory worktree from %s. Run from the repo root or the .factory worktree.\n' "$(pwd)" >&2
+    exit 1
+  fi
+fi
+
+# ---------------------------------------------------------------------------
 # Step 1: Fetch to synchronize origin/factory-artifacts before capturing SHA.
 # On failure, abort immediately — do NOT proceed with a stale EXPECTED_SHA.
 # ---------------------------------------------------------------------------
 
-if ! git -C .factory fetch origin factory-artifacts 2>&1; then
+if ! git -C "$FACTORY_DIR" fetch origin factory-artifacts 2>&1; then
   printf 'state-burst CAS push failed — fetch error before push. Retry after resolving network.\n' >&2
   exit 1
 fi
@@ -56,7 +95,7 @@ fi
 # CASPushRejected-class message and abort — do NOT push with an invalid SHA.
 # ---------------------------------------------------------------------------
 
-if ! EXPECTED_SHA="$(git -C .factory rev-parse origin/factory-artifacts 2>&1)"; then
+if ! EXPECTED_SHA="$(git -C "$FACTORY_DIR" rev-parse origin/factory-artifacts 2>&1)"; then
   printf 'state-burst CAS push failed — stale SHA after fetch: origin/factory-artifacts ref could not be resolved. Re-fetch and retry.\n' >&2
   exit 1
 fi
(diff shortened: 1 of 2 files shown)


Zious11 (changes requested):
# Fresh-Eyes PR Review — PR #727

**PR:** fix(tooling): resolve .factory worktree cwd-independently in factory-cas-push.sh
**Branch:** fix/cas-push-worktree-cwd → develop
**Issue:** #631
**CI:** GREEN (bats-full-suite, bats-darwin-leg, bats-wave-handoff, cargo-host ubuntu+macos, build-dispatcher all targets, SAST Semgrep)

## Verdict: REQUEST_CHANGES

The core fix is sound and solves the reported cwd-dependency bug with a clean, backward-compatible fast path. Blocking on one MAJOR defect in the new fallback extraction (path-with-spaces truncation) plus a latent fragility, both on the exact code path this PR introduces.

---

## Findings

### [BLOCKING / MAJOR] `awk '{print $2}'` truncates worktree paths containing spaces

```sh
MAIN_WT="$(git worktree list --porcelain 2>/dev/null | awk '/^worktree /{print $2; exit}')" || MAIN_WT=""
```

`git worktree list --porcelain` emits `worktree <absolute-path>` with the path **raw and unescaped** — only the `-z` variant is safe for special characters; `--porcelain` is not. `awk` splits on whitespace, so `$2` captures only the first token of the path. If the main worktree lives at e.g. `/Users/zious/Application Support/vsdd-factory`, `MAIN_WT` becomes `/Users/zious/Application`, `[ -d "$MAIN_WT/.factory" ]` is false, and the script prints the "could not locate" error and exits 1.

This is precisely the fallback the PR exists to add, so the defect hides behind the fast path in the common case and in CI (fixtures use no-space paths). It fails the moment the repo is cloned under a space-containing path — plausible on macOS, which this project explicitly targets. The PR carefully quotes `"$FACTORY_DIR"` everywhere to survive spaces, then truncates the path during extraction, which defeats the quoting.

**Fix (trivial):**
```sh
| awk '/^worktree /{ sub(/^worktree /, ""); print; exit }'
```
or `substr($0, 10)`. Per the project production-grade default, fix in scope rather than ship as "good enough for our path."

### [SUGGESTION / MINOR] `awk … exit` under `set -o pipefail` can wipe a valid `MAIN_WT`

`awk` calls `exit` on first match, closing the pipe read end early. If `git worktree list` is still writing, it takes SIGPIPE (exit 141); with `pipefail` the pipeline reports non-zero, the assignment "fails," and `|| MAIN_WT=""` overwrites the already-captured path with empty — routing execution into the error/exit-1 branch even though resolution succeeded.

Practically unreachable: worktree-list output for a handful of worktrees is well under the ~64KB pipe buffer, so `git` completes and exits 0 before `awk` closes the pipe. Only bites with hundreds of worktrees. Flagged as latent fragility on the critical path. Cleanest hardening: parse without early `exit`, or use `git rev-parse --path-format=absolute --git-common-dir` and take its parent, sidestepping porcelain parsing entirely.

### [NIT] Fast path `[ -d ".factory" ]` does not verify the dir is the factory worktree

Accepts any directory literally named `.factory` in cwd without confirming it is the factory-artifacts worktree. Low risk here (`.factory` is a gitignored worktree mount, not committed content). Noted for completeness.

---

## Could Not Fully Verify

- **New bats tests.** The 4 tests and `_setup_worktree_fixture` are elided in the diff (`[... omitted for brevity]`), so git-identity setup, worktree wiring, and `$WORK` isolation across tests are unassessable from what was provided. Before merge, the test source should be reviewed, and a test asserting resolution works when the main worktree path contains a **space** should be added (a no-space fixture will not catch the BLOCKING finding above).

## Verified Good

- **Path tracing:** cwd == `.factory` → fast path false → fallback derives `<main>/.factory` correctly. Repo-root and sibling-worktree cwds resolve correctly.
- **First-entry-is-main:** `git worktree list` always emits the main worktree first from any worktree — the `awk … exit` first-match assumption is a guaranteed git property.
- **`|| MAIN_WT=""` vs `set -e`:** correctly prevents `set -e` abort when not in a git repo (modulo the SIGPIPE edge above).
- **Backward compat:** repo-root fast path yields `git -C ".factory"`, byte-identical to the old `git -C .factory` (quoting inert for a no-space literal). No regression.
- **`$FACTORY_DIR` quoting:** quoted at all four call sites.
- **`$(pwd)` in error path:** best-effort in the failure branch; a `pwd` failure yields at worst an empty interpolation, not a functional problem.

---

**Bottom line:** Correct, no-regression fix for the cwd-dependency bug. Blocking only on the space-truncation defect — a one-line fix on the very path this PR introduces — and requesting the elided test source (with a space-path case) for verification.

arcavenai:
Requested changes addressed at d2cbd9bb:
- **MAJOR — space-safe extraction**: the fallback now strips the `worktree ` prefix with `sub()` instead of field-splitting with `{print $2}`, so a main-worktree path containing spaces survives intact rather than truncating to its first token.
- **MINOR — SIGPIPE**: dropped the early `exit` from the awk so `pipefail` can't observe a mid-stream pipe close.
- New bats case resolves from a story worktree under a space-containing parent dir; suite 10/10.

Ready for re-review.

Zious11 (approved):
Re-review at d2cbd9bb — APPROVE

Both REQUEST_CHANGES findings are resolved.

**MAJOR (space-path truncation) — ADDRESSED.** The `sub(/^worktree /, "")` replacement strips only the `worktree ` prefix and preserves the full path including spaces. Traced end-to-end: fallback fires from the sibling-worktree cwd, `git worktree list --porcelain` emits the space-containing main-worktree path, `sub()` yields the intact path, `$FACTORY_DIR` is set correctly, all four git steps execute successfully. The 5th new bats case (`main worktree path containing a SPACE`) exercises the awk fallback with a space-in-path fixture — genuinely RED against the old `{print $2}` code, not tautological. TD-VSDD-059 closure satisfied.

**MINOR (SIGPIPE/pipefail edge) — ADDRESSED.** The `!found`/`found=1` guard consumes the full stream rather than closing the pipe early. No SIGPIPE; no spurious pipefail abort.

**NIT (fast-path `[ -d ".factory" ]` doesn't validate worktree identity)** — unchanged from prior review, low risk, acceptable.

No new findings. 10/10 bats suite per author comment. Fix is structurally correct and production-grade.

PR #729, opened 2026-07-20
feat(factory-health): content-based factory-artifact leak scan for the product branch

### LINKED ISSUE #515: bug(factory-health): factory artifact red-gate-log-S-12.08.md tracked at repo ROOT on develop — outside .factory/, invisible to path-prefix leak checks

arcavenai:
## Why

Issue #515's Fix 1 (removing the root-level `red-gate-log-S-12.08.md`) shipped in
PR #524, and the maintainer kept the issue open for the systemic guard. The
`.factory/`-path-prefix leak check proposed in #341 can only see artifacts under
`.factory/`; a factory artifact committed to a product path — the root, `docs/`,
anywhere — is invisible to it, and since it has no live counterpart on
`factory-artifacts` it never diverges or self-surfaces. This adds the content-based
guard the maintainer asked for: it keys on the frontmatter `document_type:` rather
than location, so it catches arbitrary-path leaks that `ls-files '.factory/*'`
structurally cannot.

## What changed

- `bin/factory-artifact-leak-scan.sh` (new, executable): a self-contained,
  deterministic scanner modeled on the existing `bin/` helpers (`factory-lock-status.sh`,
  `multi-repo-scan`). It flags a tracked file as a leaked factory artifact iff:
  its path is outside `.factory/`; it is not plugin machinery
  (`plugins/vsdd-factory/{templates,tests,skills,rules}/` carry example frontmatter);
  its `document_type` is a factory-produced type — one declared by a template in
  `templates/*.md`, the same enumeration `validate-template-compliance.sh` trusts;
  and its `document_type` is not a product deliverable. `--list` / `--count` /
  table modes; exit 1 on leaks, 0 when clean. Advisory by design (reports, never
  mutates) per the issue discussion.
- `skills/factory-health/SKILL.md`: adds check #10 invoking the helper, mirroring
  how check #7 invokes `bin/factory-lock-status.sh`. No new hooks-registry entry
  and no enforcement gate — the least-drift registration, matching the maintainer's
  "factory-health check" framing.
- `tests/factory-artifact-leak-scan.bats` (new): 8 cases including the three the
  issue asks for (tracked factory frontmatter outside `.factory/` → detected; same
  content under `.factory/` → clean; product file without factory frontmatter →
  clean), plus deep-path detection, the deliverable allowlist, untracked-file
  exclusion, and plugin-machinery exclusion. Mirrors `bin.bats`/`relocate-artifact.bats`
  fixture-repo conventions; the scanner takes `VSDD_REPO_ROOT`/`CLAUDE_PLUGIN_ROOT`
  overrides so tests run against a throwaway git repo.

## Test evidence

GREEN — clean `develop` (post-#524):

```
$ bin/factory-artifact-leak-scan.sh
0 leaked factory artifacts found. Product tree is clean.
exit=0
```

This scanned all 255 `document_type`-bearing tracked `.md` files on develop with
zero false positives — including the 25 legitimately-tracked deliverables under
`docs/demo-evidence/` (see the judgment call below).

RED — reintroducing the exact artifact #515 removed, tracked at the repo root:

```
$ bin/factory-artifact-leak-scan.sh
Leaked factory artifacts detected on the product branch (#515):

document_type             tracked path
------------------------  ------------
red-gate-log              red-gate-log-S-12.08.md
...
exit=1
```

Bats suite:

```
1..8
ok 1 factory-artifact-leak-scan: tracked factory artifact outside .factory is detected (#515)
ok 2 factory-artifact-leak-scan: same artifact under .factory/ is clean (#515)
ok 3 factory-artifact-leak-scan: product file without factory frontmatter is clean (#515)
ok 4 factory-artifact-leak-scan: product-deliverable doctype in docs/ is clean (#515)
ok 5 factory-artifact-leak-scan: leak in an arbitrary deep product path is detected (#515)
ok 6 factory-artifact-leak-scan: untracked leak is not flagged (only tracked files scanned) (#515)
ok 7 factory-artifact-leak-scan: plugin template/fixture frontmatter is not flagged (#515)
ok 8 factory-artifact-leak-scan: clean product tree reports registry-clean (#515)
```

`tests/skills.bats` still passes 58/58 (SKILL.md edit breaks no structural check);
`check-bats-orphans.sh` clean; all `bin/*` syntax-check clean.

## Design judgment call (please sanity-check)

The one place invention was unavoidable is separating factory artifacts that
*leaked* from factory document_types the project *intentionally* ships on the
product branch. On develop, 255 tracked files carry `document_type:` frontmatter.
Most are plugin machinery (`templates/`, `tests/fixtures/`, `skills/*/fixtures/`)
— excluded by path. But `docs/demo-evidence/**` legitimately tracks
`demo-evidence-report` / `demo-evidence-index` deliverables on the product branch.

I derived the split from the repo's own single source of truth rather than a hand
list: `config/artifact-path-registry.yaml` (ADR-016) homes **every** registered
artifact type under `.factory/`, and the demo-evidence doctypes appear **nowhere**
in it — they have no `.factory/` canonical home, so they are product deliverables.
`red-gate-log`, by contrast, maps to `.factory/cycles/{cycle-id}/{story-id}/implementation/`
(registry line ~122, "Per-story implementation artifact (e.g. red-gate-log)").

Because the registry keys on `artifact_type` (e.g. `cycle-story-implementation`)
and frontmatter keys on `document_type` (e.g. `red-gate-log`), a fully-automatic
`document_type → registry` join would be fragile. So the scanner uses an explicit,
small, documented `PRODUCT_TRACKED_DOCTYPES` allowlist (currently
`demo-evidence-report`, `demo-evidence-index`) — exactly the template-backed
doctypes with no `.factory/` registry home that develop actually tracks in the
product tree. It is auditable and extended only with the same evidence. If you'd
prefer this be read directly from the registry (accepting a `document_type ↔
artifact_type` mapping table), that is a straightforward follow-up — flagging it
here rather than choosing the heavier mechanism unilaterally.

## Coverage boundary

This scan is **markdown-only by construction**: it walks `git ls-files '*.md'`
and keys on frontmatter `document_type:`. Several registry-homed factory
artifacts are not markdown (`sprint-state.yaml`, `wave-state.yaml`,
`policies.yaml`, `release-config.yaml`, `reference-manifest.yaml`,
`regression-state.json`, `adversary-convergence-state.json`) and carry no
frontmatter, so a leak of one of those to a product path is structurally
outside this check's detection method — not just outside its file filter. The
`red-gate-log` case motivating #515 was markdown, so the shipped check covers
the observed leak class; extending to non-markdown artifacts would need a
different signal (filename/path patterns from the artifact-path registry) and
is flagged here as a follow-up rather than widened unilaterally.

## Close rationale

With this PR, both halves of #515's proposed fix are delivered: PR #524
(merged, `69e8de30`) removed the leaked root-level artifact, and this PR is
the systemic content-based guard the issue's fix 2 asked for — the same
remaining scope named in the maintainer's keep-open comment. The coverage
boundary above (markdown-frontmatter detection) matches the issue's proposed
mechanism; extending to non-markdown registry artifacts is tracked as
follow-up, not part of #515's ask.

Closes: #515


arcaven:
The advisory-only registration (factory-health check #10, no hook or gate) reads as the right posture here — a content leak-scan is a diagnostic, and keeping it out of the deterministic gate set means a mis-parse or a newly-tracked product doctype can only surface a noisy health line rather than block the factory. The one surface worth watching is `PRODUCT_TRACKED_DOCTYPES`: while it stays advisory the two-item allowlist is low-risk, but if it ever graduates toward enforcement, landing the registry-join follow-up you flagged first would let the exclusion set derive rather than drift. Nice that the factory-doctype set itself already reads from `templates/` rather than a hand-kept list.

Zious11 (changes requested):
# PR #729 — Fresh-eyes review

**Title:** feat(factory-health): content-based factory-artifact leak scan for the product branch
**Author:** arcavenai · **Base:** develop · **Class:** Fix-PR (backlog remediation, issue #515) · **Size:** +299 / -0 across 3 files

## Verdict: REQUEST_CHANGES

No CRITICAL/HIGH defect survived verification, but two MEDIUM detection-completeness gaps
directly undermine the guard's stated purpose ("catch arbitrary-path leaks that `ls-files`
structurally cannot"). Both are ~1-line fixes and were confirmed against the real tree.
Under the production-grade default ("if it's worth doing in v1, do it correctly in v1"), a
leak detector shipped with two known classes of undetectable leaks is not production-grade
for a `feat`. Fix M1 + M2, then this is a clean approve.

### Ground-truth verification performed (fresh-eyes, against develop)
- Reproduced the script's exact `_frontmatter_doctype` awk + main loop and ran it over
  `git ls-files '*.md'` on clean develop → **0 leaks**, matching the PR's evidence. Claim is accurate.
- Refuted a candidate CRITICAL: `plugins/vsdd-factory/agents/*.md` files contain
  `document_type: story` — but only as *body* examples, not in the first frontmatter block.
  The awk's `fm==1` gate correctly ignores them. No false-positive. Good, defensible design.
- Verified `unset 'FACTORY_DOCTYPES[$dt]'` (single-quoted, variable subscript) correctly
  removes allowlisted keys on bash 5.3. The demo-evidence-report / demo-evidence-index files
  under `docs/demo-evidence/**` are correctly exempted (this is why the 0-leak result holds).
- Confirmed the M1 fix (drop `-maxdepth 1`) introduces **no** new false positives in the current tree.

---

## Findings

### M1 — MEDIUM (correctness / coverage) — `find -maxdepth 1` blinds the guard to subdirectory-template doctypes
`bin/factory-artifact-leak-scan.sh`, the FACTORY_DOCTYPES build loop:
```
find "$TEMPLATES" -maxdepth 1 -name '*.md' -type f
```
The doctype universe is derived only from top-level template files. Confirmed on develop:
`plugins/vsdd-factory/templates/adversary-prompt-templates/{phase-1d-spec-review,phase-5-code-review,phase-2-story-review}.md`
all carry `document_type: adversary-prompt-template`, and that doctype exists in **no**
top-level template. So `adversary-prompt-template` is never added to `FACTORY_DOCTYPES`, and a
file bearing that frontmatter leaked to any product path (root, `docs/`, …) is invisible to the
scanner — exactly the failure mode this PR claims to close.

**Failure scenario:** an `adversary-prompt-template` artifact is committed to `docs/` → scanner
reports "Product tree is clean. exit=0" → leak ships undetected.

**Suggestion:** drop `-maxdepth 1` (verified safe — no new false positives on the current tree;
the `templates/*` path exclusion already prevents the templates themselves from being flagged).
If restricting to a curated doctype registry is intentional, that intent must be documented and
the registry must be authoritative — silent under-inclusion is not acceptable for a completeness guard.

### M2 — MEDIUM (correctness / coverage) — `PRODUCT_TRACKED_DOCTYPES` allowlist is doctype-global, not path-scoped
```
PRODUCT_TRACKED_DOCTYPES=(demo-evidence-report demo-evidence-index)
```
These doctypes are exempted **everywhere** in the product tree, not just under their canonical
`docs/demo-evidence/` home. A genuinely-leaked `demo-evidence-report` at the repo root or in an
arbitrary directory is silently exempted — again the precise class of leak the guard exists to catch.

**Failure scenario:** `demo-evidence-report`-frontmatter file committed to repo root → exempted → undetected.

**Suggestion:** scope the exemption to `(doctype, path-prefix)` pairs, e.g. only exempt
`demo-evidence-*` when the path matches `docs/demo-evidence/`; flag the same doctype anywhere else.

### M3 — MEDIUM (correctness) — awk does not anchor the opening `---` to line 1; latent false-positive + full-file read
`_frontmatter_doctype` increments `fm` on any `^---[[:space:]]*$` line and treats the region
after the *first* such line as frontmatter. It does not require the opening fence to be line 1.
Empirically 0 hits on develop today, but any product doc that uses `---` as a thematic-break /
section separator and then has a start-of-line `document_type:` before the next `---` will be
misparsed and can false-positive — a real risk in a repo dense with factory documentation.
Secondary: a file with no closing fence (`---` once or never) is read in full looking for `fm==2`.

**Suggestion:** require the opening fence at `NR==1` (`NR==1 && /^---[[:space:]]*$/`) and only
treat `document_type` as frontmatter when it precedes the closing fence. This also bounds the read.

### A1 — ADVISORY (maintainability / drift) — hardcoded allowlist duplicates registry-derivable knowledge
Author explicitly flagged this. The allowlist is derived by hand "from absence in
`config/artifact-path-registry.yaml`," so the two drift independently. Under the production-grade
default this should not be left as an unguarded judgment call.

**Suggestion:** either derive `PRODUCT_TRACKED_DOCTYPES` from the registry at runtime, or add a
bats case asserting the hardcoded list matches the registry so drift fails CI rather than
silently degrading detection.

### A2 — ADVISORY (portability) — CRLF handling relies on `[[:space:]]` matching `\r`
`^---[[:space:]]*$` and the trim `gsub` assume `[[:space:]]` matches `\r`. True on gawk; not
guaranteed on all awk variants. Given the repo already ships CRLF fixtures
(`docs/demo-evidence/S-17.03/fixtures/fixture-crlf-foreign.md`), CRLF-fronted artifacts are a
real input. **Suggestion:** strip trailing `\r` explicitly, or add a CRLF bats case.

### A3 — ADVISORY (review limitation / wiring) — SKILL.md and bats bodies not in the provided diff
Only `bin/factory-artifact-leak-scan.sh` (162 lines) was reviewable; the ~137 lines of
`tests/factory-artifact-leak-scan.bats` (8 cases) and `skills/factory-health/SKILL.md` check #10
were elided. Two items need explicit confirmation before merge:
1. **Test coverage** — do the 8 cases cover: no-frontmatter file, malformed/single-fence
   frontmatter, allowlisted doctype in canonical vs non-canonical path (M2), a subdir-only
   doctype (M1), CRLF (A2), and the `--count`/`--list`/table modes + exit codes (0/1/2)?
2. **Advisory wiring** — the script exits `1` on leaks and `2` on error. Confirm check #10 in
   `SKILL.md` treats non-zero as advisory. If factory-health runs under `set -e` and invokes the
   helper directly, a single leak (exit 1) would abort the whole health check rather than warn.
   This flips "advisory by design" into "blocking by accident."

### A4 — ADVISORY (flake analysis, per team-lead #737 question) — unlikely cause; one residual risk to close
`#737` bats-full-suite (linux) single flake is **unlikely** to originate here. `mktemp -d`
allocates a unique dir under `$TMPDIR` (/tmp), outside the repo tree, so there is no cross-suite
path collision, and other suites' temp fixtures are untracked and outside any repo `git ls-files`
would scan. The script resolves `TEMPLATES` from the real plugin (`CLAUDE_PLUGIN_ROOT`/`_SELF_DIR`),
which is deterministic regardless of temp fixtures.

Residual risk (cannot fully rule out without the bats source): the script's `REPO_ROOT` falls back
to `git rev-parse --show-toplevel`. If a test does **not** set `VSDD_REPO_ROOT` and does not `cd`
into an isolated `git init` fixture, discovery walks up to the *real* vsdd-factory repo. That would
be a deterministic wrong-repo scan, not a race — but combined with nondeterministic CWD under
parallel bats `--jobs` it could present as an intermittent failure. Also verify the fixture uses
`git init` + **local** config (never `git config --global`) and a per-test `HOME`, so parallel
suites can't clobber shared git/global state.

**Suggestion:** confirm each bats case exports `VSDD_REPO_ROOT` to its temp fixture and uses local
git config + isolated HOME. If already done, #737 is not attributable to this PR.

---

## Checklist coverage
1. Diff coherence — OK, all changes serve #515.
2. Description accuracy — GREEN/0-leak claim verified accurate against develop.
3. Test coverage — cannot fully verify (A3); gaps flagged for M1/M2/M3/A2.
4. Demo evidence — only the GREEN/clean path shown; no RED (seeded-leak → exit 1 + table) demonstration. Recommend adding the error-path transcript.
5. Commit quality — `feat` scope correct (new feature, not a fix). Fine.
6. Diff size — 299 lines, under 500. Fine.
7. Missing changes — guard mechanism present as promised; M1/M2 are completeness gaps within it.
8. Dependency status — #524 (Fix 1) already merged; consistent with description.

arcavenai:
Both detection-completeness gaps closed at e6fb86c4:
- **Recursive scan**: dropped `-maxdepth 1`, so leaked artifacts in subdirectories are caught.
- **Doctype→home map**: `PRODUCT_TRACKED_HOMES` (doctype → expected directory) replaces the global allowlist — a product-tracked doctype found *outside* its declared home is now reported as a leak instead of skipped.
- Two regression tests cover exactly those cases; suite 10/10.

Ready for re-review.

Zious11 (changes requested):
## Re-review at e6fb86c4

M1 and M2 from the prior review are correctly addressed. The `-maxdepth 1` removal is verified safe on the current tree (no new false positives), the path-scoped `PRODUCT_TRACKED_HOMES` associative array is the right structural fix for the global-allowlist gap, and the two new regression tests (cases 9 and 10) pin both. The detection logic is sound.

Three issues block approval.

---

### BLOCKER — POLICY 21: `bin/factory-artifact-leak-scan.sh` is a new `.sh` file

`policies.yaml` POLICY 21 (`no_new_shell_scripts`, D-836 2026-07-13) prohibits new `.sh` files repository-wide. The `grandfather_clause.files` list contains exactly 5 S-19.01 entries (timing-collision pre-policy); this script is not among them. The policy's own verification step is explicit: "Any PR adding NEW .sh files beyond this explicit list remains a POLICY 21 BLOCKER."

The advisory character of the check does not create a carve-out — POLICY 21 scope is "operational bin/ tools, hooks, CI/test helpers" in full. The parallel to `bin/factory-lock-status.sh` (check #7) is the wrong model: that file is grandfathered pre-D-836; this one is not.

This was present from the first commit of the PR and was not flagged by the prior review. Neither of the two new commits (CI retrigger + M1/M2 detection fixes) addresses it.

**Required fix:** implement `factory-artifact-leak-scan` as a Rust binary in `crates/` or a WASM plugin in `hook-plugins/`. The detection logic — including the M1 recursion fix and the M2 path-scoped exemption — should carry forward into the replacement implementation. The `.bats` test file is not in scope (POLICY 21 says `.sh` specifically; the bats test suite is the established convention and is fine as-is).

---

### MEDIUM (M3, carried from prior review) — awk frontmatter parser lacks `NR==1` anchor

`_frontmatter_doctype` treats any `^---` line as a frontmatter opener, not just line 1. A `.md` document with a `---` horizontal rule before a `document_type:` line can false-positive. The prior review suggested `NR==1 && /^---[[:space:]]*$/` as a one-line fix; neither of the two new commits touches the awk function. No bats case covers this input class.

---

### MEDIUM (N1, new) — SKILL.md check #10 remediation references a variable that does not exist

The M2 fix renamed `PRODUCT_TRACKED_DOCTYPES` to `PRODUCT_TRACKED_HOMES` and changed its interface from a flat doctype list to a `(doctype → canonical-home-directory)` associative array. The SKILL.md check #10 guidance was not updated: it still tells an operator to "add its `document_type` to the helper's `PRODUCT_TRACKED_DOCTYPES` allowlist with justification." That variable does not exist in the script. An operator following this guidance cannot locate the variable or supply the required path mapping. This is a regression introduced by the M2 fix commit (e6fb86c4).

---

### Prior findings closed

- **M1** (find `-maxdepth 1` blind spot): addressed — recursive scan confirmed, test case 9 pins it.
- **M2** (doctype-global allowlist): addressed — path-scoped `PRODUCT_TRACKED_HOMES` map, test case 10 pins it.
- **A3** (bats file not in diff): resolved — all 10 cases now visible; `VSDD_REPO_ROOT` override confirmed in every `setup()`.
- **A4** (flake/parallel isolation): `git config` runs without `--global` (defaults to local repo config); per-test `mktemp -d` fixture; `VSDD_REPO_ROOT` set in `setup()`. Acceptable isolation.

Once the POLICY 21 replacement is in place (with M3 and N1 corrected in that implementation), this is a clean approve — the detection design is correct and the test coverage structure is solid.

arcavenai:
Round-2 findings addressed at 48e04651 — per-finding closure:

**POLICY 21 (BLOCKER) — FIXED.** The scanner is reimplemented as a Rust binary: workspace crate `crates/factory-artifact-leak-scan` (std-only lib with the detection core + thin CLI), and `bin/factory-artifact-leak-scan.sh` is deleted. The CLI is contract-identical to the shell version — same modes (`--list` / `--count` / table), same output text, same env contract (`CLAUDE_PLUGIN_ROOT` / `VSDD_REPO_ROOT` override), same exit codes (0 clean / 1 leaks / 2 usage-env). The M1 recursive template walk and the M2 path-scoped `PRODUCT_TRACKED_HOMES` exemption carry forward into the crate (pinned by unit tests `test_m1_collects_doctypes_recursively` / `test_m2_product_tracked_exempt_only_under_home` plus the existing bats cases 9 and 10, which pass unchanged against the binary). The bats suite stays bats per your scoping note, now invoking `target/release/factory-artifact-leak-scan` with the repo's skip-if-absent convention; the `bats-full-suite` CI build step builds the scanner alongside `factory-dispatcher` so the suite runs rather than skips.

**M3 (carried) — FIXED.** The Rust parser anchors the opening fence to line 1: `frontmatter_doctype` reads the first line, returns `None` unless it is `---`, and stops at the closing fence. A mid-document thematic break can no longer open a phantom frontmatter block. Covered by lib unit tests (`test_m3_thematic_break_is_not_frontmatter`, `test_m3_doctype_after_closing_fence_ignored`) and new bats case 11 for the input class you named. One deliberate sensitivity choice, documented in the code: a line-1 block left unclosed at EOF still yields its doctype — a malformed leaked artifact shouldn't evade the guard.

**N1 (new) — FIXED.** `factory-health` SKILL.md check #10 is rewritten: the remediation guidance now names `PRODUCT_TRACKED_HOMES` and its `(document_type -> home directory)` pair interface (matching the binary's own table-mode guidance text), and the invocation is the built binary with a build-once fallback. Added an explicit skip-and-note instruction for plugin consumers without a Rust toolchain so the health check degrades honestly rather than failing on a missing binary.

Gates run locally: `cargo fmt --check --all`, `clippy --all-targets -D warnings`, 11/11 crate unit tests, 11/11 bats, and `cargo test --workspace` (green except `full_stack_plugin_invocation` async sink-timing failures — the known #757 flake class, failing subset varies between runs, untouched by this diff).

Ready for re-review.

arcaven:
Re-review request. All three blocking findings from the 2026-07-23 round are addressed at `48e04651`, which landed later the same day.

Per finding:

**POLICY 21 (BLOCKER): FIXED.** The scanner is now the workspace crate `crates/factory-artifact-leak-scan` (lib + binary). No `.sh` file remains anywhere in the PR. The M1 recursion and M2 path-scoped exemption both carried forward into the replacement, each with a named test: `test_m1_collects_doctypes_recursively`, `test_m2_product_tracked_exempt_only_under_home`. The bats suite stays as the review scoped it.

**M3: FIXED.** `frontmatter_doctype` anchors the opening fence to line 1. Two tests pin it: `test_m3_thematic_break_is_not_frontmatter` and `test_m3_doctype_after_closing_fence_ignored`.

**N1: FIXED.** SKILL.md check #10 no longer references `PRODUCT_TRACKED_DOCTYPES`. It and the binary's remediation message now both point at `PRODUCT_TRACKED_HOMES` in the crate.

Coverage at `48e04651`: 11 crate unit tests, 11 bats cases. CI is 13 passing and 1 skipped (the release/* guard, not applicable here).

Nothing else from that review is outstanding: M1, M2, A3 and A4 were closed in the prior round.

arcaven:
Re-checked against develop `0959e34b` (fork synced, `.factory` worktree at `77992e20`). Two observations, neither of them a problem with the change itself.

**The conflict is one line.** `Cargo.toml`'s `members` list has grown from 4 entries to 25 while this sat, so resolving it is inserting `"crates/factory-artifact-leak-scan"` into the now-alphabetical list. I did that locally to check the rest: the crate builds clean against current develop and `cargo test -p factory-artifact-leak-scan` passes 11/11.

**The GREEN evidence in the body no longer reproduces**, and the reason looks more interesting than a defect in the scanner. Running the built binary over a local merge of this branch onto develop:

```
Leaked factory artifacts detected on the product branch (#515):

document_type             tracked path
------------------------  ------------
behavioral-contract       docs/demo-evidence/S-21.07/scenarios/indeterminate-index-corrupt/factory/specs/behavioral-contracts/ss-05/BC-5.39.010.md
exit=1
```

That file is a demo fixture and says so in its own frontmatter: `last_amended: "2026-01-01 (v1.6) — demo fixture: S-21.07 secondary-index UTF-8 decode failure"`. It landed 2026-08-15 in #776, after this branch point, so the body's "255 files, zero false positives" was accurate when it was captured.

The class may be worth closing before merge. A recorded scenario embeds a whole factory tree at `docs/demo-evidence/<story>/scenarios/<name>/factory/`, and skip rule 1 matches only directories literally named `.factory` or `.factory-project`, so the dotless fixture copy falls through. It is the same "frontmatter here is an example, not a live artifact" case that `is_plugin_machinery` already handles for `plugins/vsdd-factory/{templates,tests,skills,rules}/`. Two files match that shape today (`BC-INDEX.md` alongside the flagged one), and both arrived with a single scenario capture, so the count may track scenario recordings rather than staying flat.

The advisory-only registration looks well chosen for exactly this reason: today the cost is a wrong red in factory-health check #10 rather than a blocked factory.

Issue #244, opened 2026-06-24
bug(observability): claude-cost dashboard cache-hit-ratio formula omits cacheCreation tokens — displays ~100% when true value is ~89%


arcavenai:
## Summary

The "Cache hit ratio" panel in `claude-cost.json` computes:

```
cacheRead / (input + cacheRead)
```

It omits `cacheCreation` tokens from the denominator. `cacheCreation` represents cache-miss prefix-write traffic — full-freight new-input cost. With it left out, any session that does prefix-caching at all displays a near-100% ratio regardless of actual cache effectiveness.

Empirically observed right now on a session with one Claude Code instance under load:
- `input` = 122 tokens
- `cacheRead` = 8,126,790 tokens
- `cacheCreation` = 1,000,449 tokens

Dashboard formula: `8126790 / (122 + 8126790) = 99.998%` → displays **100%**.

Correct formula: `8126790 / (122 + 8126790 + 1000449) = 89.04%` → should display **89%**.

The displayed value is wrong by ~11 percentage points and trends toward false-100% the more a session writes new cache prefixes.

## Severity

**High** for accuracy — the panel is a cost-decision signal. Operators looking at "100% cache hit" conclude the session is fully cache-served and won't investigate prompt structure or context bloat. The true 89% would surface the ~11% cache-creation overhead that drives most of the cost.

### 1. The current formula

`tools/observability/grafana-dashboards/claude-cost.json`:
```
sum(increase(claude_code_token_usage_tokens_total{type="cacheRead"}[$__range])) / (sum(increase(claude_code_token_usage_tokens_total{type="input"}[$__range])) + sum(increase(claude_code_token_usage_tokens_total{type="cacheRead"}[$__range])))
```

### 2. The metric IS being emitted for all four types

```
$ curl -s "http://localhost:9090/api/v1/query?query=claude_code_token_usage_tokens_total" \
   | jq -r '.data.result[] | .metric.type' | sort -u
cacheCreation
cacheRead
input
output
```

### 3. Real-world skew

```
$ python3 -c '
input_t=122; cacheRead=8126790; cacheCreation=1000449
print("dashboard:    ", cacheRead/(input_t+cacheRead)*100, "%")
print("correct:      ", cacheRead/(input_t+cacheRead+cacheCreation)*100, "%")'
dashboard:     99.99849891906103 %
correct:       89.04108108128145 %
```

## Reproduction

```bash
[...]

Issue #243, opened 2026-06-24
bug(observability): five Grafana dashboards use '| json | attributes_X' LogQL syntax that returns zero series — Claude-side fields are structured metadata, not JSON body


arcavenai:
## Summary

Five of the seven shipped Grafana dashboards (`claude-overview.json`, `factory-today.json`, `factory-prs.json`, `factory-roi.json`, `factory-subagents.json`) use a LogQL query pattern that does not match the data shape Claude Code's OTel exporter actually produces. Every affected panel returns **zero series** even when matching log entries exist in Loki by the tens of thousands.

The broken pattern:
```logql
... | json | attributes_event_name = "tool_result" ...
```

The reason it fails:
1. The Claude Code OTel exporter emits log entries whose **body is a plain string** like `claude_code.tool_result`, not JSON. So `| json` extracts nothing.
2. The interesting fields (`event_name`, `tool_name`, `hook_event`, `hook_name`, `session_id`, `tool_use_id`, `decision_source`, etc.) are present on every log entry as **structured metadata**, not as JSON body fields.
3. The Loki OTLP ingester (per `loki-config.yaml::otlp_config`) promotes **only** the factory-side resource attributes (`service.name`, `event_type`, `hook`, `reason`, `severity`) to stream labels. Claude-side attributes deliberately remain as structured metadata.

## Severity

**High.** Operator-visible: every Claude-source dashboard panel that filters or groups by `attributes_X` is permanently broken. Combined with the parallel issue on factory-side events not reaching Loki, the operator sees a partially-populated Grafana with the Claude side showing huge "Total events" while every other Claude panel is empty.

### 1. The dashboards' query pattern

`tools/observability/grafana-dashboards/claude-overview.json`:
```
"expr": "sum(count_over_time({service_name=\"claude-code\"} | json | attributes_event_name = \"tool_result\" [$__range]))"
"expr": "count(count by (attributes_session_id) (count_over_time({service_name=\"claude-code\"} | json | attributes_session_id != \"\" [$__range])))"
"expr": "sum by (attributes_event_name) (count_over_time({service_name=\"claude-code\"} | json [$__interval]))"
"expr": "topk(10, sum by (attributes_tool_name) (count_over_time({service_name=\"claude-code\"} | json | attributes_tool_name != \"\" [$__range])))"
```

Same pattern across `factory-today.json`, `factory-prs.json`, `factory-roi.json`, `factory-subagents.json`.

### 2. The actual data shape

[...]

PR #728, opened 2026-07-20
fix(hooks/validate-count-propagation): exclude historical sections from count-drift comparison

arcavenai:
## Why

`count_propagation_drift` fires on every edit to BC-INDEX.md / STATE.md once any
count has ever changed, because `_extract_counts` scans every line — including
frozen changelog rows and phase-history rows — and compares those historical
mentions against the current count. A changelog entry that correctly records
"Lists all 36 BCs" for v1.0, or a Phase Progress row that froze "PRD (38 BCs)"
at phase-1 close, is flagged as drift against the live `total_bcs: 41`. That is
permanent alert fatigue on a check meant to catch real drift, and it creates a
perverse incentive to rewrite immutable history to silence the hook — which, per
the issue thread, already induced a real history-integrity violation. This
restricts count extraction to current-count sites so frozen records are allowed
to disagree, while genuine current-count drift still blocks.

## What changed

- `hooks/validate-count-propagation.sh`: `_extract_counts` now tracks H2 section
  boundaries and skips count mentions inside historical sections — `## Changelog`,
  `## Change Log`, `## Historical Content`, `## Phase Progress`, `## Decisions Log`.
  Boundary detection uses the repo's own idiom (`^## <name>` opens, next `^## `
  closes), the same convention `validate-changelog-monotonicity.sh` uses. A new
  `_is_historical_heading` helper holds the section list.
- Same file: the no-count guard changes from `[[ ${#SOURCE_COUNTS[@]} -eq 0 ]]`
  to `[[ -z "${SOURCE_COUNTS[*]:-}" ]]`. Under `set -u`, expanding the element
  count of an *empty associative array* is itself an unbound-variable error in
  bash 4+. Skipping historical sections can leave a file whose only counts were
  historical with an empty source map, making that path reachable — so it must
  exit 0 cleanly, not crash. (This is a pre-existing latent bug: on plain develop
  the same crash already occurs for any triggering file with zero count patterns;
  this change fixes it as a side effect of making the path reachable.)
- `tests/hooks.bats`: four regression cases (historical count in sibling not
  drift; historical count before current not drift; genuine current-count drift
  still blocks; source with only historical counts exits clean without crashing).
  Guarded by `require_bash4_count_prop` — the hook needs bash 4+ (`declare -A`),
  which is present on the ubuntu `validate` CI job (`/bin/bash` is bash 5.x) but
  not on macOS (`/bin/bash` is 3.2, where the cases skip).

## Test evidence

[...]

PR #731, opened 2026-07-20
fix(observability): include cacheCreation in cache-hit-ratio denominator

arcavenai:
## Why

The "Cache hit ratio" panel in `claude-cost.json` computes
`cacheRead / (input + cacheRead)`, omitting `cacheCreation` — the cache-*write*
tokens billed at a premium and which are not cache hits. Any session that
writes new cache prefixes therefore reads as ~100% cache-hit even when a
substantial fraction of its input-side tokens are the expensive cache-creation
kind. Because operators make budget and prompt-structure decisions off this
gauge, a false 100% actively misleads. Adding `cacheCreation` to the
denominator makes the ratio mean "share of input-side tokens served from
cache," which is what the panel title implies.

## What changed

`plugins/vsdd-factory/tools/observability/grafana-dashboards/claude-cost.json`,
"Cache hit ratio" panel (single occurrence):

- `expr`: added `+ sum(increase(claude_code_token_usage_tokens_total{type="cacheCreation"}[$__range]))` to the denominator, giving `cacheRead / (input + cacheRead + cacheCreation)`.
- `description`: restated the formula and the explicit metric definition — **"share of input-side tokens served from cache"** — and noted why cacheCreation belongs in the denominator (premium-billed, not a hit; omitting it inflates toward a false 100%). Stating the definition explicitly so you can veto it if you intend something different.

I did **not** add the optional sibling "Cache write ratio" panel the issue
mentioned ("if desired") — kept the diff to the two lines that fix the actual
defect. Easy to add as a follow-up if you want it. Two-line change, no
structural JSON edits.

## Test evidence

1. **The metric type exists.** `cacheCreation` is one of the four
   `claude_code_token_usage_tokens_total{type=...}` buckets — the same
   dashboard's "Token usage by type" panel description (`claude-cost.json:303`)
   enumerates `input, output, cacheRead, cacheCreation`, and the issue's
   Prometheus query confirms all four are emitted. So the added term resolves.

2. **Executed arithmetic on the issue's realistic magnitudes:**
   ```
   $ python3 -c 'i=122; r=8126790; c=1000449
   print("dashboard:", r/(i+r)*100)
   print("correct:  ", r/(i+r+c)*100)'
   dashboard: 99.99849891906103
   correct:   89.03767474519744
   ```
   Broken formula displays **100%**; corrected displays **89%** — an ~11
   percentage-point overstatement, and the omitted ~11% is composed of the
   premium-billed cache-write tokens.

   *Ground-truth note:* the issue body states the correct value as `89.04108%`.
   My executed arithmetic on the stated inputs (input=122, cacheRead=8126790,
   cacheCreation=1000449) gives `89.03767%`. The two differ in the third
   decimal (the issue's figure corresponds to cacheCreation≈999950, not the
   1000449 it lists). Both round to 89% and the ~11-point skew is identical, so
   the fix and its impact are unchanged — flagging the arithmetic slip in the
   issue for the record rather than papering over it.

3. **JSON validity.** `jq empty claude-cost.json` exits 0.

4. **Live verification (for the maintainer).** With the stack up
   (`"${CLAUDE_PLUGIN_ROOT}/bin/factory-obs" up`), on any session where
   `cacheCreation > 0` the gauge should now render < 100%. Cross-check against
   Prometheus:
   ```bash
   for t in input cacheRead cacheCreation; do
     v=$(curl -s "http://localhost:9090/api/v1/query?query=sum(claude_code_token_usage_tokens_total{type=\"$t\"})" | jq -r '.data.result[0].value[1] // "0"')
     echo "$t: $v"
   done
   # ratio = cacheRead / (input + cacheRead + cacheCreation)
   ```
   The issue's acceptance criterion also asks for a math test asserting the
   formula; there is no dashboard-validation harness in the tree today, so this
   PR relies on the static + executed-arithmetic evidence above and leaves the
   test harness (shared with #243/#245) as a follow-up.

Using `Refs` rather than `Closes`: the issue's math-test acceptance criterion is deferred to a follow-up, so the merge should not auto-close it.

Refs: #244

### DIFF
--- a/plugins/vsdd-factory/tools/observability/grafana-dashboards/claude-cost.json
+++ b/plugins/vsdd-factory/tools/observability/grafana-dashboards/claude-cost.json
@@ -182,7 +182,7 @@
     {
       "id": 5,
       "title": "Cache hit ratio",
-      "description": "cacheRead / (input + cacheRead). High is good — cache reads are much cheaper than fresh input. Low numbers suggest prompt caching isn't benefiting your workflow much (often fixable by restructuring prompts or system instructions).",
+      "description": "Share of input-side tokens served from cache: cacheRead / (input + cacheRead + cacheCreation). High is good — cache reads are much cheaper than fresh input. cacheCreation (cache-write) tokens are included in the denominator because they are billed at a premium and are not cache hits; omitting them inflates the ratio toward a false 100%. Low numbers suggest prompt caching isn't benefiting your workflow much (often fixable by restructuring prompts or system instructions).",
       "type": "gauge",
       "datasource": { "type": "prometheus", "uid": "prometheus" },
       "gridPos": { "h": 4, "w": 4, "x": 20, "y": 0 },
@@ -217,7 +217,7 @@
         {
           "refId": "A",
           "datasource": { "type": "prometheus", "uid": "prometheus" },
-          "expr": "sum(increase(claude_code_token_usage_tokens_total{type=\"cacheRead\"}[$__range])) / (sum(increase(claude_code_token_usage_tokens_total{type=\"input\"}[$__range])) + sum(increase(claude_code_token_usage_tokens_total{type=\"cacheRead\"}[$__range])))"
+          "expr": "sum(increase(claude_code_token_usage_tokens_total{type=\"cacheRead\"}[$__range])) / (sum(increase(claude_code_token_usage_tokens_total{type=\"input\"}[$__range])) + sum(increase(claude_code_token_usage_tokens_total{type=\"cacheRead\"}[$__range])) + sum(increase(claude_code_token_usage_tokens_total{type=\"cacheCreation\"}[$__range])))"
         }
       ]
     },


arcaven:
One small process note, if the maintainer prefers issues to auto-close only on full resolution: since the PR defers #244's math-test acceptance criterion to a follow-up, `Refs: #244` rather than `Closes: #244` would keep the issue open to track that remaining harness work — the squash-merge to develop fires the close keyword otherwise. The arithmetic and ground-truth here look solid, including the honest flag on the issue's own third-decimal slip.

arcavenai:
Switched to `Refs: #244` — the math-test acceptance criterion is deferred to a follow-up, so the squash-merge shouldn't auto-close the issue. Same rule we applied to #728 earlier today.

Zious11 (approved):
# PR Review — #731 fix(observability): include cacheCreation in cache-hit-ratio denominator

**Verdict:** APPROVE
**Base:** develop | **Class:** fix-PR (backlog remediation) | **Refs:** #244

## What I verified

1. **PromQL fix — CONFIRMED correct.** The added denominator term `+ sum(increase(claude_code_token_usage_tokens_total{type="cacheCreation"}[$__range]))` yields `cacheRead / (input + cacheRead + cacheCreation)`. Parentheses are balanced, escaped quotes preserved, and the `[$__range]` selector matches the sibling terms. This implements the described fix: cacheCreation tokens are input-side, premium-billed, and not cache hits — their prior omission inflated the ratio toward a false ~100%.
2. **Description updated accurately.** The semantic reframe ("share of input-side tokens served from cache") is the honest definition for this denominator, and the updated description explains why cacheCreation belongs there.
3. **JSON validity.** Author asserts `jq empty claude-cost.json` exits 0. No structural JSON edits.
4. **Executed arithmetic verified.** The author's Python output (99.998% → 89.038%) demonstrates the ~11 percentage-point overstatement. The 3rd-decimal difference vs the issue's stated value (89.04108%) is adequately explained by slightly different input values and does not affect the fix's correctness.

## Findings

### [LOW] Empty-vector / "No data" edge on the new term (pre-existing dashboard style)
In PromQL, if the `cacheCreation` counter series has never been scraped within `$__range` (not merely flat — a flat-but-present counter yields `increase()` = 0, which is safe), the third term is an empty vector and the entire denominator collapses to empty, making the gauge show "No data." This same failure mode already applied to the existing `input` and `cacheRead` terms. A hardened form (`... or vector(0)`) would be more robust, but applying it only to the new term while leaving the two existing terms bare would be inconsistent style. Correct scoping call for a 2-line fix PR — not a blocker.

### [LOW] Deferred math-test AC should be anchored to a concrete follow-up
The `Refs: #244` trailer correctly avoids auto-close since the math-test acceptance criterion is deferred. Under the production-grade default, "deferred to follow-up" is a defer-pattern smell unless the AC is anchored to a tracked future issue/story. Not a merge blocker, but the AC should be explicitly tracked so it does not get lost.

### [INFO] Division-by-zero unchanged and unregressed
Adding a non-negative term strictly reduces (never increases) div-by-zero likelihood. No regression.

## Summary
The change is a correct, minimal, and well-justified fix: it puts the premium-billed cache-write tokens into the denominator so the gauge reports the true share of input-side tokens served from cache instead of a misleading near-100%. The PromQL is syntactically and semantically sound, the description accurately reflects both the formula and the rationale, and JSON validity is asserted. The only notes are a pre-existing empty-vector "No data" edge (correctly left consistent with the rest of the panel) and a deferred test AC that should be anchored to a concrete follow-up.

PR #735, opened 2026-07-20
fix(skills): create factory-artifacts orphan branch via plumbing — no checkout dance, signed commit

### LINKED ISSUE #204: bug(factory-health): orphan-branch recipe strands the session on factory-artifacts when 'git checkout -' fails
## Summary

The orphan-branch creation recipe in `/vsdd-factory:factory-health` can leave the operator stranded on the new `factory-artifacts` branch with an empty index, because `git checkout -` will refuse to return to the previous branch when the working tree contains files that look "untracked" from the new branch's perspective.

## Current recipe (from skills/factory-health/SKILL.md step 1)

```bash
git checkout --orphan factory-artifacts
git rm -rf --cached . 2>/dev/null || true
git commit --allow-empty -m "chore: initialize factory-artifacts orphan branch"
git checkout -   # return to previous branch
```

## What actually happens

After the `git rm -rf --cached .`, the index is empty while the working tree has all of the previous branch's tracked files. When `git checkout -` runs, git sees those files as "untracked content that would be overwritten by checkout" and aborts:

```
error: The following untracked working tree files would be overwritten by checkout:
        CONTRIBUTING.md
        Dockerfile.test
        ...
Aborting
```

The skill does not check the return code of `git checkout -`, so it reports success while the operator is silently stranded on `factory-artifacts` with no files indexed. The next step (`git worktree add .factory factory-artifacts`) then runs from the wrong branch context, and subsequent state-validation checks misreport.

## Repro

1. Clean working tree on `develop` (or any source branch).
2. Run the four-line recipe above.
3. Check `git rev-parse --abbrev-ref HEAD` — still `factory-artifacts`, not the original branch.

## Why the recipe is fragile

`git checkout -` and `git checkout <branch>` both refuse to switch when the index is empty but the working tree has files matching the target branch's tree — git treats those files as untracked and won't overwrite them. The recipe assumes the switch always succeeds.

## Recovery (manual)

```bash
git read-tree develop      # sync index to develop's tree
git checkout develop       # now succeeds as a pure HEAD swap
```

But this requires operator knowledge and is not documented in the skill.

### Option A: use a throwaway worktree to create the orphan branch (recommended)

```bash
# Don't disturb the parent worktree at all
git worktree add --detach .factory-init-tmp HEAD
(
  cd .factory-init-tmp
  git checkout --orphan factory-artifacts
  git rm -rf --cached . 2>/dev/null || true
  git commit --allow-empty -S -m "chore: initialize factory-artifacts orphan branch"
)
git worktree remove .factory-init-tmp --force
```

The parent worktree stays on the source branch the whole time. No `git checkout -` needed.

### Option B: plumbing-only branch creation

```bash
empty_tree=$(git mktree </dev/null)
init_commit=$(git commit-tree -S -m "chore: initialize factory-artifacts orphan branch" "$empty_tree")
git branch factory-artifacts "$init_commit"
```

Cleanest — no checkout dance at all. Signs the commit explicitly (see #X for missing `-S`).

### Option C: keep the current recipe but recover on failure

Detect the stranded state and recover:
```bash
git checkout --orphan factory-artifacts
git rm -rf --cached . 2>/dev/null || true
git commit --allow-empty -S -m "chore: initialize factory-artifacts orphan branch"
prev=\$(git rev-parse --abbrev-ref @{-1})
if ! git checkout "$prev"; then
  git read-tree "$prev"
  git checkout "$prev"
fi
```

## Acceptance criteria

- [ ] Recipe never leaves the operator on `factory-artifacts` after orphan-branch creation.
- [ ] Recipe works whether the parent branch is clean or has untracked files.
- [ ] Recipe is idempotent (safe to re-run).
- [ ] Commit is signed (relies on neither user gitconfig nor CI env).

## Found during

First-time `/factory-health` invocation on `switchboard-blue` (2026-06-23, `vsdd-factory@1.0.0-rc.21`). The recipe ran without errors, but `git rev-parse --abbrev-ref HEAD` showed `factory-artifacts` instead of `develop` afterward. The user had to manually run `git read-tree develop && git checkout develop` to recover.

arcavenai:
## Why

Both health skills create the `factory-artifacts` orphan branch by moving
`HEAD` onto it and then switching back:

```bash
git checkout --orphan factory-artifacts
git rm -rf --cached . 2>/dev/null || true
git commit --allow-empty -m "chore: initialize factory-artifacts orphan branch"
git checkout -   # return to previous branch
```

The return step is unchecked, and it is the failure point. After
`git checkout --orphan`, `git checkout -` has no `@{-1}` reflog entry to
resolve, so it exits non-zero with `pathspec '-' did not match any file(s)`
and `HEAD` is left **stranded on `factory-artifacts`**. From that moment every
subsequent pipeline commit — STATE.md writes, spec artifacts, phase-gate
backups — lands on the orphan branch instead of the working branch, silently.
The `factory-worktree-health` variant has the same fragility class: it uses
`git rm -rf .` (which deletes the working-tree files, not just the index) and a
hardcoded `git checkout develop` return that lands on the wrong branch whenever
the pipeline started somewhere other than `develop`. The commits were also
unsigned, so the initialization commit could not satisfy a signed-history
requirement.

This is a health-check skill — the one thing that is supposed to keep the
factory worktree sane. When its own repair step can strand the session on the
artifacts branch, a "health check" becomes the thing that breaks the pipeline.

## What changed

Both skills now create the branch with plumbing, so `HEAD` never moves and the
working tree is never touched:

```bash
git branch factory-artifacts \
  "$(git commit-tree -S "$(git mktree </dev/null)" \
     -m "chore: initialize factory-artifacts orphan branch")"
```

`git mktree </dev/null` produces the empty tree; `git commit-tree -S` builds a
parentless (orphan), signed root commit for it; `git branch` points the ref at
that commit without any checkout. There is no return step to fail, no working
tree to disturb, and no hardcoded branch to guess wrong.

- **`skills/factory-health/SKILL.md`** — replaced the four-command checkout
  dance in "Orphan branch exists → If missing" with the single plumbing
  command; updated the surrounding prose (the old "return to previous branch"
  comment is gone, replaced by a note on why `HEAD` no longer moves).
- **`skills/factory-worktree-health/SKILL.md`** — replaced the checkout-dance
  variant in "Step 1: Check Remote Branch → Branch does NOT exist" with the
  same plumbing command (parameterized on `${BRANCH_NAME}`, so it serves both
  `factory-artifacts` and `factory-project-artifacts`). The `git push origin
  ${BRANCH_NAME}` is retained — this recipe builds a remote branch — and the
  fragile `git checkout develop` return is removed.
- **`tests/orphan-branch-plumbing.bats`** — new focused suite (13 tests):
  behavioral invariants for the new recipe, RED reproductions proving the old
  recipes can strand `HEAD`/lose files, a gated signature-verification test,
  and contract assertions that the shipped SKILL.md files carry the fix.

## Test evidence

New suite, run from the plugin root (`bats plugins/vsdd-factory/tests/orphan-branch-plumbing.bats`):

```
1..13
ok 1 new recipe: HEAD/current branch unchanged after creation
ok 2 new recipe: factory-artifacts ref resolves to a commit
ok 3 new recipe: factory-artifacts is a parentless (orphan) empty-tree commit
ok 4 new recipe: working tree is byte-for-byte untouched
ok 5 old factory-health recipe strands HEAD (git checkout - cannot resolve)
ok 6 old factory-worktree-health recipe destroys untracked working-tree files
ok 7 old factory-worktree-health recipe: hardcoded return lands on develop, not the actual start branch
ok 8 new recipe does NOT strand: same fixture, HEAD stays put
ok 9 new recipe with -S produces a verifiable signature
ok 10 factory-health SKILL.md uses the plumbing recipe
ok 11 factory-health SKILL.md no longer contains the checkout dance in branch creation
ok 12 factory-worktree-health SKILL.md uses the plumbing recipe
ok 13 factory-worktree-health SKILL.md no longer contains the checkout dance in branch creation
```

Test 5 reproduces the stranding deterministically: it runs the old
factory-health recipe verbatim and asserts `git checkout -` exits non-zero
and leaves the current branch on `factory-artifacts` instead of the starting
branch. Below is the same sequence run by hand, showing the strand:

```
$ git checkout --orphan factory-artifacts
Switched to a new branch 'factory-artifacts'
$ git rm -rf --cached . 2>/dev/null || true
rm 'README.md'
rm 'src/main.rs'
$ git commit --allow-empty -m "chore: initialize factory-artifacts orphan branch"
$ git checkout -   # unchecked return — the defect
error: pathspec '-' did not match any file(s) known to git
git checkout - exit code: 1
current branch: factory-artifacts   <-- STRANDED (expected: develop)
```

The new recipe, run against the identical fixture (tracked files under `src/`,
a `README.md`, and an untracked `untracked.txt`), holds every invariant:

```
=== BEFORE ===
current branch: develop
HEAD: b06a781a...
status --porcelain:
?? untracked.txt
=== RUN PLUMBING RECIPE ===
git branch factory-artifacts "$(git commit-tree ... "$(git mktree </dev/null)" ...)"
=== AFTER ===
current branch: develop                             # unchanged
HEAD: b06a781a...                                    # unchanged
factory-artifacts resolves: 368f6fac...
factory-artifacts tree: 4b825dc642cb6eb9a060e54bf8d69288fbee4904   # the empty tree
factory-artifacts parents: (none)                    # orphan
status --porcelain:
?? untracked.txt                                     # working tree untouched
```

Signing verified with the recipe run verbatim (`-S`) under an ephemeral SSH
signing key (test 9, and by hand):

```
$ git branch factory-artifacts "$(git commit-tree -S "$(git mktree </dev/null)" -m "...")"
$ git verify-commit factory-artifacts
Good "git" signature ... with ED25519 key SHA256:...
current branch: develop   # HEAD still didn't move, even with signing on
```

The `factory-worktree-health` variant was verified end-to-end against a bare
remote: the plumbing recipe followed by `git push origin ${BRANCH_NAME}`
publishes an orphan, empty-tree, signed branch to the remote while the local
`HEAD` stays on the starting branch.

The signing test self-skips where SSH commit signing is unavailable, and all
behavioral tests strip `-S` so they run on any host with no key configured (as
CI runs); the `-S` in the shipped recipe is covered by the gated signing test
plus a contract assertion. Environment: git 2.50.1, bats 1.13.0.

Closes: #204

### DIFF
--- a/plugins/vsdd-factory/skills/factory-health/SKILL.md
+++ b/plugins/vsdd-factory/skills/factory-health/SKILL.md
@@ -17,12 +17,14 @@ Validate that the `.factory/` worktree is properly mounted and healthy. Auto-rep
 git branch --list factory-artifacts
 ```
 
-- **If missing**: Create it.
+- **If missing**: Create it. Build the empty root commit with plumbing and
+  point the branch ref at it directly — this never moves `HEAD`, never touches
+  the working tree, and signs the commit (`-S`). No checkout dance, so there is
+  no failed-return case that could strand the session on `factory-artifacts`.
   ```bash
-  git checkout --orphan factory-artifacts
-  git rm -rf --cached . 2>/dev/null || true
-  git commit --allow-empty -m "chore: initialize factory-artifacts orphan branch"
-  git checkout -  # return to previous branch
+  git branch factory-artifacts \
+    "$(git commit-tree -S "$(git mktree </dev/null)" \
+       -m "chore: initialize factory-artifacts orphan branch")"
   ```
 
 ### 2. Worktree is mounted
--- a/plugins/vsdd-factory/skills/factory-health/SKILL.md
+++ b/plugins/vsdd-factory/skills/factory-health/SKILL.md
@@ -18,13 +18,19 @@ git branch --list factory-artifacts
 ```
 
 - **If missing**: Create it. Build the empty root commit with plumbing and
-  point the branch ref at it directly — this never moves `HEAD`, never touches
-  the working tree, and signs the commit (`-S`). No checkout dance, so there is
-  no failed-return case that could strand the session on `factory-artifacts`.
+  point the branch ref at it directly — this never moves `HEAD` and never
+  touches the working tree. No checkout dance, so there is no failed-return
+  case that could strand the session on `factory-artifacts`. The init marker
+  commit is deliberately unsigned: `git commit-tree` does not honor
+  `commit.gpgsign`, and a hard `-S` would fail outright in environments with
+  no signing key configured (fresh CI runners, agent containers, plugin
+  consumers). Real content commits on the branch are signed later by
+  state-manager under the repo's normal commit config.
   ```bash
-  git branch factory-artifacts \
-    "$(git commit-tree -S "$(git mktree </dev/null)" \
-       -m "chore: initialize factory-artifacts orphan branch")"
+  commit=$(git commit-tree "$(git mktree </dev/null)" \
+    -m "chore: initialize factory-artifacts orphan branch") \
+    || { echo "failed to create factory-artifacts init commit" >&2; exit 1; }
+  git branch factory-artifacts "$commit"
   ```
 
 ### 2. Worktree is mounted
--- a/plugins/vsdd-factory/skills/factory-worktree-health/SKILL.md
+++ b/plugins/vsdd-factory/skills/factory-worktree-health/SKILL.md
@@ -81,13 +81,16 @@ git ls-remote --heads origin ${BRANCH_NAME}
 ```
 
 - **Branch exists:** Proceed to Step 2
-- **Branch does NOT exist:** Create it:
+- **Branch does NOT exist:** Create it. Build the empty root commit with
+  plumbing and point the branch ref at it directly, then push. `HEAD` never
+  moves and the working tree is never disturbed — so there is no `git checkout`
+  step whose failure could strand the session on `${BRANCH_NAME}` (nor any
+  hardcoded return target to guess wrong). The commit is signed (`-S`).
   ```bash
-  git checkout --orphan ${BRANCH_NAME}
-  git rm -rf .
-  git commit --allow-empty -m "chore: initialize ${BRANCH_NAME} branch"
+  git branch ${BRANCH_NAME} \
+    "$(git commit-tree -S "$(git mktree </dev/null)" \
+       -m "chore: initialize ${BRANCH_NAME} branch")"
   git push origin ${BRANCH_NAME}
-  git checkout develop
   ```
   Then proceed to Step 2.
 
--- a/plugins/vsdd-factory/skills/factory-worktree-health/SKILL.md
+++ b/plugins/vsdd-factory/skills/factory-worktree-health/SKILL.md
@@ -85,12 +85,21 @@ git ls-remote --heads origin ${BRANCH_NAME}
   plumbing and point the branch ref at it directly, then push. `HEAD` never
   moves and the working tree is never disturbed — so there is no `git checkout`
   step whose failure could strand the session on `${BRANCH_NAME}` (nor any
-  hardcoded return target to guess wrong). The commit is signed (`-S`).
+  hardcoded return target to guess wrong). The local-existence guard covers
+  the re-run case where the branch already exists locally but was never
+  pushed — the Step 1 check above is remote-only. The init marker commit is
+  deliberately unsigned: `git commit-tree` does not honor `commit.gpgsign`,
+  and a hard `-S` would fail outright in environments with no signing key
+  configured; content commits are signed later by state-manager under the
+  repo's normal commit config.
   ```bash
-  git branch ${BRANCH_NAME} \
-    "$(git commit-tree -S "$(git mktree </dev/null)" \
-       -m "chore: initialize ${BRANCH_NAME} branch")"
-  git push origin ${BRANCH_NAME}
+  if ! git show-ref --verify --quiet "refs/heads/${BRANCH_NAME}"; then
+    commit=$(git commit-tree "$(git mktree </dev/null)" \
+      -m "chore: initialize ${BRANCH_NAME} branch") \
+      || { echo "failed to create ${BRANCH_NAME} init commit" >&2; exit 1; }
+    git branch "${BRANCH_NAME}" "$commit"
+  fi
+  git push origin "${BRANCH_NAME}"
   ```
   Then proceed to Step 2.
 
--- a/plugins/vsdd-factory/tests/orphan-branch-plumbing.bats
+++ b/plugins/vsdd-factory/tests/orphan-branch-plumbing.bats
@@ -1,0 +1,290 @@
+#!/usr/bin/env bats
+# orphan-branch-plumbing.bats — behavioral + contract tests for issue #204:
+# the factory-artifacts orphan branch must be created with plumbing
+# (git commit-tree + git mktree + git branch), never with a checkout dance.
+#
+# THE BUG (#204): both factory-health and factory-worktree-health created the
+# orphan branch by switching HEAD onto it (git checkout --orphan), clearing the
+# index, committing, then switching back (git checkout - / git checkout develop).
+# The return step is unchecked. When it fails — e.g. `git checkout -` cannot
+# resolve the previous ref, or the working-tree files that became untracked
+# under the orphan would be overwritten — HEAD is silently STRANDED on
+# factory-artifacts and subsequent pipeline work commits to the wrong branch.
+# The commits were also unsigned.
+#
+# THE FIX (issue's Option B, plumbing-only):
+#   git branch <name> "$(git commit-tree -S "$(git mktree </dev/null)" -m "...")"
+# HEAD never moves, the working tree is never touched, the commit is signed.
+#
+# TEST STRATEGY
+#   - Behavioral (GREEN): run the new recipe in a scratch repo with tracked +
+#     untracked files and assert the invariants: current branch unchanged,
+#     factory-artifacts resolves, it is a parentless (orphan) empty-tree commit,
+#     and `git status --porcelain` is byte-identical before and after.
+#   - Behavioral (RED): run each OLD recipe in the same scratch fixture and
+#     assert it CAN strand HEAD on factory-artifacts — this is the defect the
+#     fix removes. These tests PASS by demonstrating the old behavior is broken.
+#   - Signing: run the recipe verbatim WITH -S under an ephemeral SSH signing
+#     key and assert the commit verifies. Skipped if SSH commit signing is
+#     unavailable in the environment (ssh-keygen missing or git too old).
+#   - Contract: the shipped SKILL.md files contain the plumbing recipe and no
+#     longer contain the fragile checkout-dance in the branch-creation block.
+#
+#   The behavioral tests strip -S from the recipe so they run on any host with
+#   no signing key configured (as CI runs). The -S delta is covered separately
+#   by the gated signing test and by the contract test that asserts -S is
+#   present in the shipped recipe text.
+#
+# Run:
+#   bats plugins/vsdd-factory/tests/orphan-branch-plumbing.bats
+
+setup() {
+  PLUGIN_ROOT="$(cd "$BATS_TEST_DIRNAME/.." && pwd)"
+  FACTORY_HEALTH="$PLUGIN_ROOT/skills/factory-health/SKILL.md"
+  WORKTREE_HEALTH="$PLUGIN_ROOT/skills/factory-worktree-health/SKILL.md"
+
+  WORK="$(mktemp -d)"
+  # Resolve symlinks (macOS /var -> /private/var) for stable path comparisons.
+  WORK="$(cd "$WORK" && pwd -P)"
+}
+
+teardown() {
+  rm -rf "$WORK"
+}
+
+# ---------------------------------------------------------------------------
+# Build a scratch repo on branch `develop` with one seed commit, a tracked
+# file, and an untracked file. Commit signing is DISABLED so the fixture is
+# CI-portable (no key required). Sets REPO and STARTING_BRANCH.
+# ---------------------------------------------------------------------------
+_setup_repo() {
+  REPO="$WORK/repo"
+  git init -q -b develop "$REPO"
+  git -C "$REPO" config user.email "test@orphan-plumbing.test"
+  git -C "$REPO" config user.name "Orphan Plumbing Test"
+  git -C "$REPO" config commit.gpgsign false
+
+  mkdir -p "$REPO/src"
+  echo "fn main() {}" > "$REPO/src/main.rs"
+  echo "# project"     > "$REPO/README.md"
+  git -C "$REPO" add -A
+  git -C "$REPO" commit -q -m "seed develop with tracked files"
+
+  # An untracked file present at recipe time — the class of working-tree state
+  # that the old checkout-dance could clobber or refuse to restore over.
+  echo "scratch notes" > "$REPO/untracked.txt"
+
+  STARTING_BRANCH="$(git -C "$REPO" branch --show-current)"
+}
+
+# The empty tree object hash is a git constant.
+EMPTY_TREE="4b825dc642cb6eb9a060e54bf8d69288fbee4904"
+
+# ===========================================================================
+# BEHAVIORAL — GREEN: the plumbing recipe holds every invariant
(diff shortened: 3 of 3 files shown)


Zious11 (changes requested):
# PR #735 Fresh-Eyes Review — fix(skills): factory-artifacts orphan branch via plumbing

**Base:** develop · **Files:** 3 (+305 / -10) · **Class:** Fix-PR (production-grade bar)
**Verdict:** REQUEST_CHANGES

---

## Summary

The core mechanism is correct and is a genuine improvement over the checkout dance.
`git branch <name> "$(git commit-tree -S "$(git mktree </dev/null)" -m ...)"` builds a
parentless empty-tree commit and points a ref at it without ever moving `HEAD` or
touching the working tree. That eliminates both original defects (stranded `HEAD` from
`git checkout -` after `--orphan`, and destructive `git rm -rf .` / wrong hardcoded
return branch).

Two issues block approval under the production-grade bar, both centered on the newly
**mandatory** `-S` signing flag: it is a robustness regression on an auto-repair path
that runs in varied environments, and the test suite treats the one hard dependency the
recipe introduces as optional — so CI cannot catch the regression.

### Review limitation
The diff contains only a placeholder comment for `tests/orphan-branch-plumbing.bats`
(the 290-line body is not in the provided diff). Findings #2 and #4 concern test
internals and are therefore assessed from the PR description and test-name list, not the
actual test source. They are marked accordingly.

---

## Findings

### [HIGH] Mandatory `-S` is a robustness regression on an auto-repair path
- **Category:** correctness / coherence
- **Where:** both SKILL.md recipes (`git commit-tree -S ...`)
- The OLD `factory-health` recipe used `git commit --allow-empty`, which is *adaptive*:
  it signs iff `commit.gpgsign=true` is configured, and otherwise produces an unsigned
  commit successfully. The NEW recipe uses `git commit-tree -S`, which **forces**
  signing unconditionally. `git commit-tree` does **not** honor `commit.gpgsign`, so
  there is no config path to opt out.
- Consequence: in any environment without a configured signing key (fresh CI runner,
  agent container, or a **downstream consumer of the distributed vsdd-factory plugin**),
  `git commit-tree -S` fails (non-zero, "gpg failed to sign the data" / SSH key load
  error). Where the old recipe *succeeded* (unsigned), the new one now *fails* — and
  this is on `factory-health` / `factory-worktree-health`, which run at session/pipeline
  start as auto-repair. A robustness regression on the repair path is worse than the bug
  being fixed for those environments.
- These SKILL.md files ship in the marketplace plugin and run in environments the author
  does not control, so "this repo mandates signed commits" does not cover all consumers.
- **Suggestion:** Preserve the strand-bug fix but make signing non-fatal for
  unconfigured environments. Options, in order of preference:
  1. Drop `-S` on the empty init marker commit (real content commits on
     `factory-artifacts` are signed later by state-manager); the init commit's signature
     carries marginal value.
  2. Make it config-driven: sign only when a signing key is present (detect via
     `git config --get user.signingkey` / `commit.gpgsign`), else fall back to unsigned —
     replicating the old adaptive behavior while keeping the no-checkout fix.
  3. If mandatory signing is a deliberate hard requirement, document the signing-key
     prerequisite in both SKILL.md files and pair it with #2 below (a hard CI gate).

### [HIGH] Test/production signing asymmetry — CI cannot catch the regression
- **Category:** test-coverage
- **Where:** test 9 ("new recipe with -S produces a verifiable signature") — skips when
  SSH signing is unavailable; production recipe hard-requires `-S`.
- The single hard dependency the recipe introduces (signing) is the one thing the suite
  treats as optional. On a machine without signing configured, tests 1–8/10–13 pass and
  test 9 skips → suite is green, while the actual recipe emitted in SKILL.md would
  **fail** on that same machine. The test matrix is exactly inverted relative to the
  runtime risk.
- **Suggestion:** Resolve jointly with #1. If signing stays mandatory, provision an
  ephemeral signing key in CI and make test 9 a **mandatory** gate (no skip) that
  additionally asserts the recipe *fails cleanly with a clear message* when no key is
  present. If signing becomes config-driven/optional (preferred), add a test for the
  no-key path proving the branch is still created (unsigned) rather than the repair
  aborting.
- *(Assessed from PR description; test body not in diff.)*

### [MEDIUM] commit-tree failure yields `git branch <name> ""` — misleading secondary error, no guard
- **Category:** correctness
- **Where:** both recipes; nested command substitution with no error handling.
- If the inner `git commit-tree` fails (e.g. signing failure per #1), the substitution
  expands to empty and the outer command becomes `git branch factory-artifacts ""`,
  which errors with `fatal: Not a valid object name: ''` — a confusing message that
  points at branch creation, not the real (signing) cause. No branch is created and the
  agent following the SKILL is left to diagnose a misleading error. The one-liner has no
  `set -euo pipefail`, no non-empty check on the captured SHA.
- **Suggestion:** Split into two steps and validate:
  ```bash
  commit=$(git commit-tree "$(git mktree </dev/null)" -m "chore: initialize factory-artifacts orphan branch") \
    || { echo "failed to create init commit" >&2; exit 1; }
  git branch factory-artifacts "$commit"
  ```
  (with `-S` added per the resolution of #1). Capturing to a variable also makes the
  failure attributable.

### [MEDIUM] Behavioral tests likely validate an embedded copy, not the actual SKILL.md recipe
- **Category:** test-coverage
- Tests 1–4 verify behavior of *a* recipe, tests 10/12 assert SKILL.md "uses the plumbing
  recipe," and 11/13 assert the checkout dance is gone. From the names, 10–13 appear to be
  substring/contract matches and 1–4 exercise a recipe string. If tests 1–4 run an embedded
  copy of the recipe rather than the exact fenced block extracted from SKILL.md, then a
  typo introduced into SKILL.md (e.g. a dropped quote, wrong flag) would still pass the
  behavioral tests while shipping a broken recipe — false confidence.
- **Suggestion:** Extract the fenced ```bash``` recipe directly from each SKILL.md and
  execute *that exact text* in the behavioral tests, binding behavior to the shipped
  content. If already done, disregard.
- *(Assessed from PR description; test body not in diff — please confirm.)*

### [LOW] factory-worktree-health checks remote existence but creates a local branch
- **Category:** correctness / edge-case
- The guard is `git ls-remote --heads origin ${BRANCH_NAME}` (remote), but the repair does
  `git branch ${BRANCH_NAME} ...` (local). On a re-run where the local branch already
  exists but the remote does not, `git branch` fails with "already exists." The remote
  guard does not cover local-branch existence.
- **Suggestion:** Guard on local existence too (`git show-ref --verify --quiet refs/heads/${BRANCH_NAME}`)
  before creating, or use `git branch -f`/create-if-absent semantics, then push.

### [LOW] Unquoted `${BRANCH_NAME}` in the worktree-health recipe and push
- **Category:** shell-hygiene
- `git branch ${BRANCH_NAME}` and `git push origin ${BRANCH_NAME}` are unquoted. The value
  is controlled, so no functional bug, but production-grade shell snippets in a distributed
  skill should quote (`"${BRANCH_NAME}"`).

### [ADVISORY] RED reproduction tests (5/6/7) freeze a historical broken recipe
- Tests 5/6/7 embed the old recipes to prove they were broken. This is good regression
  documentation and does not drift (the old recipe is no longer in SKILL.md, per tests
  11/13). But they test now-dead code and will not catch a *different* future regression.
  Acceptable; low maintenance burden. No change required.

### [ADVISORY] No coverage of the push step in factory-worktree-health
- The `git push origin ${BRANCH_NAME}` step (remote interaction) is untested. Push ordering
  itself is correct (create local ref, then push; sequential, no race — a concurrent
  remote branch would cause a safe non-fast-forward rejection, not a clobber). Hard to test
  in CI without a remote fixture; acceptable to leave uncovered.

---

## Checklist verification

1. **Diff coherence** — PASS. All three files serve the single fix.
2. **Description accuracy** — PASS. Description matches the diff; root-cause analysis is
   accurate (`git checkout -` cannot resolve `@{-1}` after `--orphan`).
3. **Test coverage** — PARTIAL. 13 tests cover behavior + RED reproductions + contract.
   Gaps: signing-failure path (#2), no-key branch creation, recipe-extraction binding (#4).
4. **Demo evidence** — N/A for a skill/shell-recipe fix; bats TAP output serves as
   evidence. Acceptable for this class.
5. **Commit quality** — Conventional format, clear message. PASS.
6. **Diff size** — 305 lines, mostly the new test file. Reasonable.
7. **Missing changes** — The plumbing recipe correctly replaces both dances. Mandatory
   `-S` introduces a new gap (#1) not present in the original intent.
8. **Dependency status** — No upstream PR dependency evident.

### Recipe correctness answers to the review questions
- `git mktree </dev/null` → empty tree object; `git commit-tree` with no `-p` → parentless
  orphan commit; `git branch <name> <sha>` → ref only, `HEAD`/worktree untouched. Mechanism
  is correct across all supported git versions (all three commands are long-stable).
- `-S` failure mode is **loud** (non-zero + stderr), not silent — but it cascades into the
  misleading `git branch <name> ""` secondary error (#3), and it is a **regression** vs the
  old adaptive `git commit` behavior (#1).
- Push ordering is correct; no race (#ADVISORY above).

---

## Verdict: REQUEST_CHANGES

Blocking on **#1 (mandatory `-S` regression)** and **#2 (test/production signing
asymmetry)**. The recipe is directionally right and fixes real bugs, but forcing signing on
a distributed auto-repair path — while the test suite makes signing optional — fails the
production-grade bar. Resolve #1 (drop or config-gate `-S`) and #2 (align the test gate),
address #3/#4 (error handling + recipe-extraction binding), and this is a clean approve.

arcavenai:
All findings addressed at 400eaa41:
- **HIGH 1 — `-S` dropped** (option 1): both SKILL.md recipes now create the init marker commit unsigned, with inline rationale (commit-tree ignores `commit.gpgsign`; a hard `-S` fails outright on keyless CI runners/containers/plugin consumers; content commits are signed later by state-manager under normal commit config).
- **HIGH 2 — test asymmetry inverted back**: the skippable signing test is gone. In its place: a **mandatory keyless-environment test** (global+system git config masked, asserts the branch is created unsigned, `%G? == N`) and a **clean-failure test** (a `user.useConfigOnly` repo where commit-tree cannot succeed — asserts the recipe fails with its own attributable message and creates *no* branch).
- **MEDIUM 3 — substitution split**: `commit=$(git commit-tree …) || { echo …; exit 1; }` then `git branch <name> "$commit"` — no more `git branch <name> ""` misdirection.
- **MEDIUM 4 — tests bound to shipped text**: behavioral tests now EXTRACT the fenced recipe from each SKILL.md and execute that exact text; a typo in the shipped snippet fails the suite.
- **LOW 5 — local-existence guard**: `git show-ref --verify --quiet refs/heads/${BRANCH_NAME}` covers the local-exists/remote-missing re-run (test included: remote ref deleted, re-run reuses the local branch and restores the push).
- **LOW 6 — quoting**: all `${BRANCH_NAME}` expansions quoted.

Suite 18/18. Ready for re-review.

Zious11 (approved):
## Re-review at 400eaa41

All six prior findings are closed. No new blocking issues.

**HIGH 1 — mandatory `-S` regression:** Addressed. `-S` is structurally absent from both SKILL.md recipes. Both files explain inline why the init marker commit is deliberately unsigned (`git commit-tree` ignores `commit.gpgsign`; a hard `-S` would fail on any keyless host). The contract test asserts `grep -F "commit-tree -S"` exits non-zero, providing a mechanical guard.

**HIGH 2 — test/production signing asymmetry:** Addressed. The skippable signing test is replaced by two mandatory tests: a keyless-environment test (global+system config masked, asserts branch created unsigned, `%G? == N`) and a clean-failure test (`user.useConfigOnly` repo, asserts the recipe fails with its own message and creates no branch). Since the production recipe no longer has `-S`, tests and production are now aligned.

**MEDIUM 3 — misleading secondary error:** Addressed. Two-step pattern in place (`commit=$(...)  || { echo "..."; exit 1; }`; then `git branch <name> "$commit"`). Clean-failure test asserts the failure message is the recipe's own, not "not a valid object name".

**MEDIUM 4 — tests bound to shipped text:** Addressed. The `_extract_recipe` awk function reads the actual fenced `commit-tree` block from each SKILL.md at test runtime and `eval`s that exact text. A typo in SKILL.md fails the behavioral tests at execution.

**LOW 5 — local-existence guard:** Addressed. `if ! git show-ref --verify --quiet "refs/heads/${BRANCH_NAME}"` wraps the creation step; the re-run test (remote ref deleted, local branch present) covers the specific case the remote-only Step 1 check misses.

**LOW 6 — unquoted expansions:** Addressed. All `${BRANCH_NAME}` and `$commit` expansions are quoted; contract test asserts the quoted forms by substring match.

No new findings. Mechanism is correct, error handling is load-bearing, tests are genuinely bound to the shipped text. 18/18 green.

Issue #206, opened 2026-06-23
bug(factory-obs): dispatcher races with /factory-health worktree setup by continuously recreating .factory/logs/


arcavenai:
## Summary

The factory-obs dispatcher, once registered against a project, persistently recreates `.factory/logs/` and writes `dispatcher-internal-*.jsonl` even when `.factory/` does not exist or has just been moved aside. This races with `/vsdd-factory:factory-health`'s worktree setup steps and causes downstream failures (see related issue on nested-worktree mount).

## Repro

1. `factory-obs register` against a fresh project. `.factory/logs/` is created with dispatcher-internal logs.
2. `mv .factory /tmp/factory-bak` — `.factory/` no longer exists.
3. Within seconds, `.factory/logs/dispatcher-internal-<date>.jsonl` reappears with fresh content.
4. Repeat the move — same result.

Observed during `/factory-health` repair: across three move-aside operations, the dispatcher recreated `.factory/logs/` three times, each with overlapping but non-identical content.

## Why this is a problem

- **Data overlap.** Each recreation produces a new `dispatcher-internal-<date>.jsonl`. If the operator moves `.factory/` aside, then the dispatcher writes a new file with the same name into the freshly-recreated `.factory/logs/`. Restoring the original loses the new writes; restoring the new loses the old writes.
- **Worktree-mount race.** `git worktree add .factory ...` requires `.factory/` to be empty/absent. The dispatcher recreating `.factory/` mid-setup causes the worktree to mount at the wrong path (see related issue).
- **Surprise mutation.** The operator cannot reason about `.factory/` state because it changes underneath them.

### Option A: write dispatcher-internal logs OUTSIDE the project (recommended)

`dispatcher-internal-*.jsonl` is **not** application/event data — it's the dispatcher's own diagnostic log. It does not belong inside the user's repo. Move these to:

- `~/.factory-obs/dispatchers/<project-hash>/` or
- `~/Library/Logs/factory-obs/<project-name>/` (macOS) / `~/.local/state/factory-obs/<project-name>/` (Linux)

Only application event JSONL (the `events-*.jsonl` that the OTel collector tails) should live in `.factory/logs/`.

### Option B: dispatcher only writes when .factory/ is a confirmed worktree

The dispatcher should verify `.factory/` is a git worktree on `factory-artifacts` before opening any log file. If not, it should write to a fallback path outside the repo and emit a warning.

```python
def resolve_logs_dir(project_root):
[...]

Issue #205, opened 2026-06-23
bug(factory-health): 'git worktree add .factory factory-artifacts' silently mounts at .factory/.factory/ when racing with dispatcher


arcavenai:
## Summary

When `/vsdd-factory:factory-health` runs `git worktree add .factory factory-artifacts` against a project where the dispatcher is concurrently recreating `.factory/logs/`, the worktree silently ends up mounted at `.factory/.factory/` (one level too deep) instead of `.factory/`. Git reports success; the skill's downstream validations do not detect the nested mount.

This is **possibly related to #130** (which fixed the recursive shadow caused by dispatcher cwd resolution), but the symptom here is different: there is no `.factory/.factory/logs/` shadow — the *worktree itself* is one level too deep.

## Repro

1. Clean project, `factory-artifacts` orphan branch exists.
2. Trigger the dispatcher to create `.factory/logs/` (e.g., via `factory-obs register` + any hook firing).
3. Move `.factory/` aside (`mv .factory .factory.bak`).
4. Run `git worktree add .factory factory-artifacts`.
5. Inspect:
   - `git worktree list` reports `/project/.factory/.factory  <sha>  [factory-artifacts]`
   - `.factory/.git` points to `/project/.git/worktrees/-factory/`
   - But `.factory/` actually contains a nested `.factory/` worktree, not the worktree root.

In my run, `git worktree add .factory factory-artifacts` printed:
```
Preparing worktree (checking out 'factory-artifacts')
HEAD is now at 0bb23d7 chore: initialize factory-artifacts orphan branch
```
— a clean success. But the resulting layout was:
```
.factory/              ← plain dir (recreated by dispatcher between mv and add)
.factory/.factory/     ← the actual worktree
.factory/logs/         ← dispatcher logs from the race window
```

## Why this slips past validation

The skill's step 3 (`cd .factory && git branch --show-current`) returns `develop`, not `factory-artifacts`, because `.factory/` is now a regular directory containing the worktree, not the worktree itself. The skill interprets this as "worktree on wrong branch" and (per the documented recovery) would `git worktree remove .factory --force` — but that fails because `.factory/` is not a worktree; only `.factory/.factory/` is.

## Possible root cause

[...]

Issue #203, opened 2026-06-23
bug(onboarding): /onboard-observability creates .factory/logs/ in a plain dir, blocking later /factory-health worktree mount


arcavenai:
## Summary

`/vsdd-factory:onboard-observability` and `/vsdd-factory:factory-health` have a hard ordering dependency that neither skill enforces or documents. Running onboard-observability first on a fresh project puts the project in a state that factory-health cannot auto-repair without destructive operations on user data.

## Repro (vsdd-factory@1.0.0-rc.21)

Fresh project, no `.factory/`, no `factory-artifacts` orphan branch.

1. Invoke `/vsdd-factory:onboard-observability`
2. The skill calls `factory-obs register` which (transitively, via the running dispatcher hook) creates `.factory/logs/dispatcher-internal-2026-06-23.jsonl` inside a **plain directory** `.factory/`.
3. Invoke `/vsdd-factory:factory-health`
4. The skill's step 2 tries `git worktree add .factory factory-artifacts`, which requires `.factory/` to be empty or absent. It is not — it contains live dispatcher logs.

The factory-health skill has no branch for "plain `.factory/` directory containing only dispatcher logs" — its only options are "missing → create" and "mounted → check branch." The result is a stuck state where the operator must manually preserve logs, remove the dir, then mount.

## Why this is a problem

Onboarding observability is the **first** thing a new factory project does (and is encouraged in the onboard-observability skill's "When to use" section: "Brand new project that just had the vsdd-factory plugin installed"). Factory-health is the **prerequisite** for the rest of the pipeline. Today these two skills race and the documented first-time-use sequence fails.

### Option A: ordering — onboard-observability requires factory-health first

`/vsdd-factory:onboard-observability` should detect missing `factory-artifacts` worktree and either:
- Refuse with a clear "run /vsdd-factory:factory-health first" message, or
- Invoke factory-health implicitly before proceeding.

### Option B: factory-obs register defers log creation

`factory-obs register` should only register the project's path in the watch list. It should **not** trigger the dispatcher to start writing to `.factory/logs/` until `.factory/` is confirmed to be a `factory-artifacts` git worktree. The dispatcher could resolve `.factory/` via `git worktree list` rather than path-only.

### Option C: factory-health preserves pre-existing `.factory/logs/`

[...]

PR #737, opened 2026-07-20
fix(skills): assert .factory mounts at repo root; gate onboard-observability on the mount

arcavenai:
## Why

`/factory-health` mounts `.factory/` with a bare `git worktree add .factory factory-artifacts` and never checks where it landed. If `.factory/` already exists as a plain **non-empty** directory, the bare command fails on current git (`fatal: '.factory' already exists`; verified 2.50/2.55) — and a nested mount at `.factory/.factory` has been observed once from that state (#205; mechanism unconfirmed — candidates are a nested add during error recovery or a racing re-create mid-mount). Under the nested layout the whole check silently misbehaves: step 3's `cd .factory && git branch --show-current` reads the **parent** branch (so a corrupt layout looks healthy), and the documented recovery `git worktree remove .factory --force` can't fix it because `.factory` isn't the worktree — the worktree is one level down.

The plain-directory precondition isn't hypothetical: `/onboard-observability` creates one. Its "When to use" says "brand new project that just had the vsdd-factory plugin installed," but its Prerequisites require a `.factory/` ancestor — a contradiction. Run onboarding first and `factory-obs register` drops a plain `.factory/logs/`, which is exactly the plain dir that makes the bare mount fail (`already exists`) when `/factory-health` runs later — the state from which #205's nested layout was observed.

This adds the missing post-mount assertion and a plain-dir guard to `/factory-health`, and gives `/onboard-observability` an explicit ordering prerequisite so it refuses to run before the mount exists.

## What changed

Class 0 — skill instruction text only; no engine/hook code.

- **`skills/factory-health/SKILL.md`** step 2:
  - Pre-mount **plain-dir guard**: if `.factory` exists but isn't a worktree, `mv` it aside before `git worktree add`, with a commented restore step for its contents (mirrors the mv-aside idiom already used by the sibling `/factory-worktree-health` skill).
  - Post-mount **assertion**: `[ "$(git -C .factory rev-parse --show-toplevel)" = "$(git rev-parse --show-toplevel)/.factory" ]`, aborting on mismatch.
  - On failure, names the **nested-mount** shape and explicitly says NOT to run `git worktree remove .factory --force` — instead remove the real nested worktree (`.factory/.factory`) and re-run from the guard.
- **`skills/onboard-observability/SKILL.md`**:
  - Prerequisite 1 now asserts `.factory/` is a **mounted worktree at the repo root** (same `rev-parse --show-toplevel` check) and, on failure, refuses with "run `/factory-health` first" rather than creating `.factory/`.
  - "When to use" no longer claims "brand new project that just had the plugin installed"; it states onboarding runs **after** `/factory-health` has mounted `.factory/`, resolving the contradiction with the prerequisite.
- **New suite `tests/factory-mount-guards.bats`** (9 tests): 5 content-contract greps (RED before these edits, GREEN after) asserting the assertion / nested-mount warning / plain-dir guard / onboard prerequisite / corrected "When to use" text are present; plus 4 execution tests that fabricate real git layouts (healthy, nested `.factory/.factory`, plain `.factory/logs`, absent) and prove the assertion command distinguishes them.

## Test evidence

New suite `tests/factory-mount-guards.bats`, run against this worktree (bats 1.x, git 2.50.1).

RED (before the skill edits — the 5 content-contract tests fail; the 4 execution tests already pass because they validate git logic, not prose):
```
not ok 1 factory-health step 2 has a post-mount assertion on .factory toplevel (#205)
not ok 2 factory-health warns against blind 'worktree remove --force' on a nested mount (#205)
not ok 3 factory-health guards against a pre-existing plain .factory directory before mounting (#205/#203)
not ok 4 onboard-observability refuses unless .factory is a mounted worktree (#203)
not ok 5 onboard-observability 'When to use' no longer claims brand-new-just-installed (#203)
ok 6 mount assertion PASSES for a healthy repo-root .factory worktree
ok 7 mount assertion FAILS for a nested .factory/.factory mount (#205 shape)
ok 8 mount assertion FAILS for a plain .factory/logs dir (onboard-first, #203 shape)
ok 9 mount assertion FAILS when .factory is absent
```

GREEN (after the skill edits — all 9 pass):
```
ok 1 .. ok 9   (9 tests, 0 failures)
```

No regressions: `tests/skills.bats` (incl. the onboard-observability structure + model-invocable checks) and `tests/factory-lock-status.bats` (which greps `factory-health/SKILL.md`) both still pass; `tests/check-bats-orphans.sh` passes.

**One honest caveat on the execution tests.** The `#205` report states that the recovery `git worktree remove .factory --force` *fails* under a nested mount. On git 2.50.1 that command returned exit 0 in my repro (it prunes the plain-dir entry but does not touch the real nested worktree, so it's ineffective rather than erroring). The tests therefore do NOT assert that removal errors — they assert the two things that reproduce deterministically: the mount assertion detects the nested layout, and the real worktree is at `.factory/.factory` (proving the blind removal targets the wrong path). The fix text warns against the blind removal on the "wrong path" grounds, which holds regardless of git's exit code.

## Reviewer note — scope and #203/#206 mapping

Refs, not Closes. The **root cause** of the nested-mount race — the dispatcher mounting `.factory` while a plain dir exists — lives in the dispatcher and is tracked separately as #206 (its own PR). This PR is the skill-text defense-in-depth: it makes `/factory-health` detect and refuse the corrupt layout, and stops `/onboard-observability` from *creating* the plain dir that provokes it. #203's full closure (the onboarding-ordering contradiction) depends on #206 landing the dispatcher-side ordering guarantee; this PR removes the skill-level contradiction and the foot-gun but does not by itself close the race. Hence both trailers are `Refs:`.

Refs: #205
Refs: #203

### DIFF
--- a/plugins/vsdd-factory/skills/factory-health/SKILL.md
+++ b/plugins/vsdd-factory/skills/factory-health/SKILL.md
@@ -38,10 +38,10 @@ git worktree list | grep -F '.factory'
   running `/onboard-observability` before this check), `git worktree add`
   fails on current git with `fatal: '.factory' already exists` when the
   directory is non-empty (verified on git 2.50/2.55; an empty one mounts
-  cleanly). Nested mounts at `.factory/.factory` — a corrupt layout — have
-  also been observed from this state (#205): a nested path passed while
-  recovering from the already-exists error, or a process re-creating
-  `.factory` mid-setup, produces one. Move the plain directory aside first
+  cleanly). A nested mount at `.factory/.factory` — a corrupt layout — has
+  also been observed once from this state (#205; mechanism unconfirmed —
+  candidates are a nested add path during error recovery or a process
+  re-creating `.factory` mid-setup). Move the plain directory aside first
   (mirrors `/factory-worktree-health`), mount, then restore its contents:
   ```bash
   if [ -e .factory ] && ! git -C .factory rev-parse --git-dir >/dev/null 2>&1; then
@@ -59,9 +59,9 @@ git worktree list | grep -F '.factory'
     || echo "ABORT: .factory did not mount at the repo root"
   ```
   If the assertion fails, the worktree is almost certainly **nested** at
-  `.factory/.factory` (a nested add path used while recovering from
-  `already exists`, or a racing process re-creating `.factory` mid-mount —
-  see #205). Do NOT run `git worktree remove .factory --force` — that
+  `.factory/.factory` (#205 — observed once, mechanism unconfirmed;
+  candidates include a nested add during error recovery or a racing
+  re-create mid-mount). Do NOT run `git worktree remove .factory --force` — that
   removal targets the wrong path and cannot fix a nested mount. Instead remove
   the *actual* nested worktree and re-run this check from the plain-dir guard:
   ```bash
--- a/plugins/vsdd-factory/skills/factory-health/SKILL.md
+++ b/plugins/vsdd-factory/skills/factory-health/SKILL.md
@@ -32,11 +32,40 @@ git worktree list | grep -F '.factory'
 ```
 
 - **If missing**: Mount it.
+
+  First guard against a pre-existing **plain** `.factory/` directory. If
+  `.factory` exists but is NOT a worktree (e.g. a bare `.factory/logs/` left by
+  running `/onboard-observability` before this check), `git worktree add`
+  mounts the worktree *nested* at `.factory/.factory` — a corrupt layout. Move
+  the plain directory aside first (mirrors `/factory-worktree-health`), mount,
+  then restore its contents:
   ```bash
+  if [ -e .factory ] && ! git -C .factory rev-parse --git-dir >/dev/null 2>&1; then
+    mv .factory .factory-backup-$(date +%s)   # plain dir, not a worktree
+  fi
   git worktree add .factory factory-artifacts
+  # If a backup was made, restore any contents (e.g. logs/) into the worktree:
+  #   cp -R .factory-backup-*/. .factory/ 2>/dev/null || true
+  #   rm -rf .factory-backup-*
   ```
 
-- **If mounted but pointing to wrong branch**: Remove and remount.
+  Then **assert the mount landed at the repo root**, not nested:
+  ```bash
+  [ "$(git -C .factory rev-parse --show-toplevel)" = "$(git rev-parse --show-toplevel)/.factory" ] \
+    || echo "ABORT: .factory did not mount at the repo root"
+  ```
+  If the assertion fails, the worktree is almost certainly **nested** at
+  `.factory/.factory` (git chose that path because `.factory` already existed as
+  a plain directory). Do NOT run `git worktree remove .factory --force` — that
+  removal targets the wrong path and cannot fix a nested mount. Instead remove
+  the *actual* nested worktree and re-run this check from the plain-dir guard:
+  ```bash
+  git worktree remove .factory/.factory --force
+  rm -rf .factory        # now a leftover plain dir; the guard above re-handles it
+  ```
+
+- **If mounted but pointing to wrong branch**: Remove and remount, then re-run
+  the same post-mount assertion above.
   ```bash
   git worktree remove .factory --force
   git worktree add .factory factory-artifacts
--- a/plugins/vsdd-factory/skills/factory-health/SKILL.md
+++ b/plugins/vsdd-factory/skills/factory-health/SKILL.md
@@ -36,9 +36,13 @@ git worktree list | grep -F '.factory'
   First guard against a pre-existing **plain** `.factory/` directory. If
   `.factory` exists but is NOT a worktree (e.g. a bare `.factory/logs/` left by
   running `/onboard-observability` before this check), `git worktree add`
-  mounts the worktree *nested* at `.factory/.factory` — a corrupt layout. Move
-  the plain directory aside first (mirrors `/factory-worktree-health`), mount,
-  then restore its contents:
+  fails on current git with `fatal: '.factory' already exists` when the
+  directory is non-empty (verified on git 2.50/2.55; an empty one mounts
+  cleanly). Nested mounts at `.factory/.factory` — a corrupt layout — have
+  also been observed from this state (#205): a nested path passed while
+  recovering from the already-exists error, or a process re-creating
+  `.factory` mid-setup, produces one. Move the plain directory aside first
+  (mirrors `/factory-worktree-health`), mount, then restore its contents:
   ```bash
   if [ -e .factory ] && ! git -C .factory rev-parse --git-dir >/dev/null 2>&1; then
     mv .factory .factory-backup-$(date +%s)   # plain dir, not a worktree
@@ -55,8 +59,9 @@ git worktree list | grep -F '.factory'
     || echo "ABORT: .factory did not mount at the repo root"
   ```
   If the assertion fails, the worktree is almost certainly **nested** at
-  `.factory/.factory` (git chose that path because `.factory` already existed as
-  a plain directory). Do NOT run `git worktree remove .factory --force` — that
+  `.factory/.factory` (a nested add path used while recovering from
+  `already exists`, or a racing process re-creating `.factory` mid-mount —
+  see #205). Do NOT run `git worktree remove .factory --force` — that
   removal targets the wrong path and cannot fix a nested mount. Instead remove
   the *actual* nested worktree and re-run this check from the plain-dir guard:
   ```bash
--- a/plugins/vsdd-factory/skills/factory-health/SKILL.md
+++ b/plugins/vsdd-factory/skills/factory-health/SKILL.md
@@ -41,32 +41,65 @@ git worktree list | grep -F '.factory'
   cleanly). A nested mount at `.factory/.factory` — a corrupt layout — has
   also been observed once from this state (#205; mechanism unconfirmed —
   candidates are a nested add path during error recovery or a process
-  re-creating `.factory` mid-setup). Move the plain directory aside first
-  (mirrors `/factory-worktree-health`), mount, then restore its contents:
+  re-creating `.factory` mid-setup).
+
+  Note `rev-parse --git-dir` CANNOT detect the plain-dir case — for an
+  in-repo plain directory it walks up, finds the parent repo's `.git`, and
+  exits 0, exactly as it does for a real worktree. The guard therefore
+  compares canonicalized (`pwd -P`, symlink-safe) worktree toplevels: a
+  healthy mount's toplevel resolves to `<repo-root>/.factory`, a plain dir's
+  resolves to the parent repo root. Run as one block — guard, prune stale
+  registrations, mount, restore, assert:
   ```bash
-  if [ -e .factory ] && ! git -C .factory rev-parse --git-dir >/dev/null 2>&1; then
-    mv .factory .factory-backup-$(date +%s)   # plain dir, not a worktree
+  canon() { (cd "$1" 2>/dev/null && pwd -P); }    # symlink-safe path compare
+  repo_root="$(canon "$(git rev-parse --show-toplevel)")"
+
+  # Nested corrupt layout (#205)? Positively confirmed only — route to the
+  # recovery block below rather than mounting over it.
+  if [ -e .factory/.factory/.git ]; then
+    echo "ABORT: nested .factory/.factory mount detected — run the #205 recovery block" >&2
+    exit 1
   fi
+
+  # Plain-dir guard: move aside anything that is not a repo-root worktree.
+  if [ -e .factory ] && \
+     [ "$(canon "$(git -C .factory rev-parse --show-toplevel 2>/dev/null || echo /nonexistent)")" != "$repo_root/.factory" ]; then
+    mv .factory ".factory-backup-$(date +%s)"   # plain dir, not a root worktree
+  fi
+
+  git worktree prune   # drop stale registrations left by earlier bad recoveries
   git worktree add .factory factory-artifacts
-  # If a backup was made, restore any contents (e.g. logs/) into the worktree:
-  #   cp -R .factory-backup-*/. .factory/ 2>/dev/null || true
-  #   rm -rf .factory-backup-*
+
+  # Restore any backed-up contents (e.g. logs/) into the fresh worktree.
+  for b in .factory-backup-*; do
+    [ -e "$b" ] || continue
+    cp -R "$b/." .factory/
+    rm -rf "$b"
+  done
+
+  # Post-mount assertion — hard stop; later checks must not run on a bad mount.
+  [ "$(canon "$(git -C .factory rev-parse --show-toplevel 2>/dev/null || echo /nonexistent)")" = "$repo_root/.factory" ] \
+    || { echo "ABORT: .factory did not mount at the repo root" >&2; exit 1; }
   ```
 
-  Then **assert the mount landed at the repo root**, not nested:
+  **#205 recovery block** — only for a positively confirmed nested mount
+  (`.factory/.factory/.git` exists). Do NOT run
+  `git worktree remove .factory --force` — that removal targets the wrong
+  path and cannot fix a nested mount. And do not delete anything on a mere
+  assertion failure: if the layout is not the known nested shape, inspect
+  `git worktree list` manually instead. The wrapper directory is moved
+  aside, not deleted, so its contents (e.g. logs/) are restored by the
+  mount block's backup-restore step on re-run:
   ```bash
-  [ "$(git -C .factory rev-parse --show-toplevel)" = "$(git rev-parse --show-toplevel)/.factory" ] \
-    || echo "ABORT: .factory did not mount at the repo root"
-  ```
-  If the assertion fails, the worktree is almost certainly **nested** at
-  `.factory/.factory` (#205 — observed once, mechanism unconfirmed;
-  candidates include a nested add during error recovery or a racing
-  re-create mid-mount). Do NOT run `git worktree remove .factory --force` — that
-  removal targets the wrong path and cannot fix a nested mount. Instead remove
-  the *actual* nested worktree and re-run this check from the plain-dir guard:
-  ```bash
-  git worktree remove .factory/.factory --force
-  rm -rf .factory        # now a leftover plain dir; the guard above re-handles it
+  if [ -e .factory/.factory/.git ]; then
+    git worktree remove .factory/.factory --force
+    mv .factory ".factory-backup-$(date +%s)"   # preserve wrapper contents
+    git worktree prune                          # drop the dangling registration
+    # Now re-run the mount block above; it mounts clean and restores the backup.
+  else
+    echo "ABORT: not the known nested shape — inspect 'git worktree list' manually" >&2
+    exit 1
+  fi
   ```
 
 - **If mounted but pointing to wrong branch**: Remove and remount, then re-run
--- a/plugins/vsdd-factory/skills/onboard-observability/SKILL.md
+++ b/plugins/vsdd-factory/skills/onboard-observability/SKILL.md
@@ -24,9 +24,10 @@ Before any other action, say verbatim:
 1. `.factory/` must be a **mounted factory-artifacts worktree at the repo root** —
    not merely a directory that happens to exist. This ordering matters: if you
    run this skill first, `factory-obs register` would create a plain
-   `.factory/logs/` directory, and a later `/factory-health` would then mount
-   the worktree *nested* at `.factory/.factory` (a corrupt layout). Assert the
-   mount before doing anything:
+   `.factory/logs/` directory, and a later `/factory-health` mount then fails
+   (`fatal: '.factory' already exists` on current git) — with nested
+   `.factory/.factory` mounts observed from botched recoveries of that state
+   (#205). Assert the mount before doing anything:
    ```bash
    [ "$(git -C .factory rev-parse --show-toplevel 2>/dev/null)" = "$(git rev-parse --show-toplevel)/.factory" ]
    ```
@@ -132,7 +133,7 @@ This skill **does not**:
 ## When to use
 
 Run this **after** `/factory-health` has mounted `.factory/` (see Prerequisites) —
-never before, or the mount lands nested.
+never before, or the leftover plain directory blocks the mount.
 
 - **Newly set-up project** whose `.factory/` worktree is already mounted and which you now want wired into observability.
 - **Existing project** you've been working in but haven't connected to your running stack yet (e.g., you set up the stack from a different project and want this one's events in the same Grafana).
--- a/plugins/vsdd-factory/skills/onboard-observability/SKILL.md
+++ b/plugins/vsdd-factory/skills/onboard-observability/SKILL.md
@@ -27,13 +27,14 @@ Before any other action, say verbatim:
    `.factory/logs/` directory, and a later `/factory-health` mount then fails
    (`fatal: '.factory' already exists` on current git) — with nested
    `.factory/.factory` mounts observed from botched recoveries of that state
-   (#205). Assert the mount before doing anything:
+   (#205). Assert the mount before doing anything — the snippet halts, it does
+   not merely warn (paths canonicalized so symlinked checkouts don't false-fail):
    ```bash
-   [ "$(git -C .factory rev-parse --show-toplevel 2>/dev/null)" = "$(git rev-parse --show-toplevel)/.factory" ]
+   canon() { (cd "$1" 2>/dev/null && pwd -P); }
+   [ "$(canon "$(git -C .factory rev-parse --show-toplevel 2>/dev/null || echo /nonexistent)")" = "$(canon "$(git rev-parse --show-toplevel)")/.factory" ] \
+     || { echo ".factory/ is not a mounted worktree — run /factory-health first to set it up, then re-run this skill." >&2; exit 1; }
    ```
-   If this fails, abort with: **"`.factory/` is not a mounted worktree — run
-   `/factory-health` first to set it up, then re-run this skill."** Do not create
-   `.factory/` here.
+   Do not create `.factory/` here.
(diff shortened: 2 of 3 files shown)


arcaven:
The guard, assertion, and onboarding-ordering prerequisite all look sound, and it's nice that the mount check turns out symlink-safe as written. One soft thing worth a second look for the record: the "Why" says a pre-existing plain `.factory/` makes `git worktree add .factory factory-artifacts` mount nested at `.factory/.factory`, but on git 2.55.0 that bare command doesn't nest — an empty `.factory` mounts cleanly and a non-empty one (the onboard `.factory/logs/` case) errors with `fatal: '.factory' already exists`; the only way I could produce a nested mount was passing `.factory/.factory` explicitly, which is what the four execution tests fabricate rather than reproduce from the documented command. The fix still stands on its own (the mv-aside cleanly handles that real "already exists" case), so this is really just about version-qualifying or lightly rescoping the mechanism note so the motivation matches what the command actually does.

arcavenai:
Verified on git 2.50.1 here — same result as your 2.55.0: empty plain `.factory` mounts cleanly, non-empty fails `fatal: '.factory' already exists`, and the bare documented command never nests. Rescoped in d4f15031: both skills' mechanism prose now says the plain-dir state *blocks* the mount on current git, with nesting attributed to where it actually comes from (nested add paths used in recovery, or a process re-creating `.factory` mid-mount — the #205 observation), and the bats header now states explicitly that the execution tests fabricate the nested layout rather than reproduce it from the bare command. Guard, assertion, and recovery are unchanged — as you said, the mv-aside is exactly what the real already-exists case needs.

arcaven:
Thanks for the quick rescope — the SKILL.md prose and the bats header now match what the command actually does, and the FABRICATE note lines up with the explicit `.factory/.factory` path the execution tests pass.

Two leftovers that may be worth the same sweep: the PR description still carries the pre-rescope mechanism in its first and third paragraphs ("git mounts the worktree *nested* at `.factory/.factory`" / "exactly the plain dir that triggers the nested mount"), and the bats header's #203 paragraph still says the plain dir is "exactly the plain-dir that makes the #205 nested mount happen", twelve lines under the corrected #205 paragraph.

One softer point on the replacement attribution: #205's repro has no recovery step in it, and its author noted he couldn't reproduce the nesting in isolation outside the dispatcher race window — so "a nested path passed while recovering" and "a process re-creating `.factory` mid-setup, produces one" may both read as firmer than #205 supports; "observed once, mechanism still unexplained" might be the safer shape.

Separately, the red `bats-full-suite (linux)` looks like infrastructure rather than your change — the job's only annotation is `System.IO.IOException: No space left on device` on the runner's own diag log, it died at 5m40s where #738/#740 took ~18m45s on the same base, and the logs come back `BlobNotFound` because the worker crashed before upload. A re-run should clear it.

arcavenai:
All three leftovers actioned in `bf0f13e7` + a body edit:

- **PR body paras 1 & 3** rescoped to the corrected mechanism (bare mount fails `already exists` on non-empty plain dir; nested layout attributed to #205 as observed-once).
- **bats header #203 paragraph** now says the plain dir "makes the bare mount fail (`already exists`)" instead of "makes the #205 nested mount happen".
- **Attribution softened** in both the header and SKILL.md to "observed once, mechanism unconfirmed" with the candidate paths listed as candidates rather than asserted mechanisms — agreed that #205's own reporter couldn't reproduce the nesting in isolation, so the firmer phrasing overclaimed.

Guard/assertion/recovery unchanged; 9/9 bats green locally. The push re-triggers CI, which should also clear the `No space left on device` runner red — thanks for pulling the annotation, agreed it's infrastructure.

Zious11 (changes requested):
# PR #737 Review — fix(skills): assert .factory mounts at repo root; gate onboard-observability on the mount

**Reviewer:** pr-reviewer (fresh-eyes, different-model cognitive diversity)
**Verdict:** **REQUEST_CHANGES**
**Class:** Fix-PR / skill instruction text + new bats test (held to full production-grade bar)

---

## Summary

The PR correctly diagnoses a real problem (bare `git worktree add .factory` with no landing-site
check; `/onboard-observability` creating a plain `.factory/` that later blocks or nests the mount)
and the `onboard-observability` half of the fix is sound. However, the primary remediation in
`factory-health/SKILL.md` — the pre-mount plain-directory guard — **does not fire in the exact
precondition it targets**, because its detection test succeeds for a plain in-repo directory. In
addition, the nested-mount recovery instruction carries a `rm -rf .factory` that can destroy a
healthy worktree on a false-positive assertion. These are blocking under the production-grade bar.

The single linux bats flake is *plausibly* attributable to worktree test isolation (see F4), but I
cannot confirm it from the summarized `.bats` body — flagged as needs-verification, not confirmed.

---

## Findings

### F1 — BLOCKING / HIGH — `correctness`
**factory-health/SKILL.md, plain-dir guard**

```bash
if [ -e .factory ] && ! git -C .factory rev-parse --git-dir >/dev/null 2>&1; then
  mv .factory .factory-backup-$(date +%s)   # plain dir, not a worktree
fi
```

`git -C .factory rev-parse --git-dir` is **not** a valid "is this a worktree" test here. When
`.factory` is a plain directory sitting inside the project repo — which is the precondition the
guard exists for (a bare `.factory/logs/` left by `/onboard-observability` / `factory-obs
register`) — git discovers the *parent* repo's `.git` by walking up the tree and exits 0. The
condition `! git ... rev-parse --git-dir` therefore evaluates false, the `&&` short-circuits, and
`mv .factory .factory-backup-*` **never runs**.

Net effect: the guard is a no-op in exactly the case it is meant to handle. Control then falls
through to `git worktree add .factory factory-artifacts`, which fails on the non-empty plain dir
(`fatal: '.factory' already exists`, acknowledged in the diff text itself) or nests — i.e. the PR
does not prevent the failure it claims to fix.

**Failure scenario:** repo with a leftover plain `.factory/logs/` directory → run
`/factory-health` → guard's rev-parse succeeds via parent-repo discovery → mv skipped → `git
worktree add .factory` fails / nests. Same outcome as pre-fix.

**Fix:** use a test that distinguishes a mounted worktree from a plain in-repo subdir. The PR
*already has the right check* in the post-mount assertion and in `onboard-observability` —
`--show-toplevel` equality:
```bash
if [ -e .factory ] && \
   [ "$(git -C .factory rev-parse --show-toplevel 2>/dev/null)" != "$(git rev-parse --show-toplevel)/.factory" ]; then
  mv .factory .factory-backup-$(date +%s)
fi
```
Standardize on `--show-toplevel` for both the guard and the assertions.

---

### F2 — BLOCKING / HIGH — `data-loss`
**factory-health/SKILL.md, nested-mount recovery**

```bash
git worktree remove .factory/.factory --force
rm -rf .factory        # now a leftover plain dir; the guard above re-handles it
```

Two compounding risks:

1. **False-positive assertion → destroys a healthy worktree.** The assertion that gates this
   recovery is a raw string comparison of two `--show-toplevel` outputs (see F7). On macOS a
   symlinked TMPDIR (`/tmp` → `/private/tmp`) or a trailing-slash / canonicalization mismatch can
   make a *healthy* mount fail the equality check. The recovery then runs `rm -rf .factory` against
   a live, correctly-mounted worktree, deleting its working tree. This is unrecoverable file loss.

2. **"the guard above re-handles it" is false** given F1. After `rm -rf .factory`, the guard does
   not reclaim the state (F1 makes it a no-op), and if `git worktree remove` did not actually
   remove the registration (the author's own caveat notes git 2.50.1 returns exit 0 while only
   pruning the plain-dir entry), a dangling worktree registration remains in the parent repo's
   `.git/worktrees`, needing a manual `git worktree prune`.

**Fix:** gate `rm -rf` behind an explicit confirmation that the path is not a live worktree; use
`-ef`/`realpath` comparison (F7); and add `git worktree prune` after removal. Do not rely on the
F1 guard to clean up.

---

### F3 — SUGGESTION / MEDIUM — `correctness`
**factory-health/SKILL.md, post-mount assertion**

```bash
[ "$(git -C .factory rev-parse --show-toplevel)" = "$(git rev-parse --show-toplevel)/.factory" ] \
  || echo "ABORT: .factory did not mount at the repo root"
```

The word "ABORT" implies halting, but `echo` prints and execution continues into the following
blocks. If run as a script the mount-failure path proceeds regardless. Should be:
```bash
|| { echo "ABORT: .factory did not mount at the repo root" >&2; exit 1; }
```
(`onboard-observability` handles this correctly by instructing the agent to abort; `factory-health`
should match that rigor mechanically.)

---

### F4 — SUGGESTION / MEDIUM — `test-isolation` (needs-verification; candidate CI-flake cause)
**tests/factory-mount-guards.bats**

The 4 execution tests create real git worktrees via `git worktree add`, share `$REPO` from
`_setup_repo`, and teardown runs `git worktree prune` + `rm -rf $WORK`. If `_setup_repo` does not
allocate a **unique per-test (or per-file) `mktemp -d`**, or if `git worktree prune` / `rm -rf`
operate on any shared or ambient repo, then:
- bats file-level parallelism (`--jobs`) across suites, or any shared temp namespace, can race
  worktree add/prune between concurrent tests;
- worst case, a worktree operation executed with CWD inside the actual vsdd-factory checkout could
  touch the real repo's worktree list.

This is the **most plausible source of the single linux flake** among the PR's changes (the 5
content-contract grep tests cannot flake). I cannot confirm from the summarized `.bats` body —
**verify** that: (a) each test uses its own `mktemp -d`; (b) all worktrees are created under that
unique dir; (c) teardown scopes `prune`/`rm` to that dir only and never to a shared/real repo;
(d) tests never `git worktree add` with CWD inside the project repo. If any is not guaranteed,
tighten isolation.

---

### F5 — SUGGESTION / MEDIUM — `usability`
**factory-health/SKILL.md, commented-out restore**

```bash
# If a backup was made, restore any contents (e.g. logs/) into the worktree:
#   cp -R .factory-backup-*/. .factory/ 2>/dev/null || true
#   rm -rf .factory-backup-*
```

Left as a comment, an agent executing the block literally will: (a) not restore observability logs
into the new worktree, silently stranding them; and (b) leave `.factory-backup-<ts>` dirs
accumulating with no cleanup. Either make the restore an executed, conditional step (only when a
backup was actually created) or state explicitly that the operator must run it and why it is
guarded. As written it is ambiguous and risks silent data continuity loss.

---

### F6 — NIT / LOW — `coherence`
**tests/factory-mount-guards.bats, setup()**

The setup fabricates `factory-artifacts` via the OLD `git checkout --orphan` dance. This overlaps
PR #735, which removes that dance. Not a correctness break — the test builds a self-contained
scratch repo, so it does not depend on the plugin's canonical recipe — but it is fixture
duplication/divergence that will rot. Prefer referencing the canonical helper (once #735 lands) or
add a comment noting the intentional standalone fabrication so a future reader does not "fix" it
into the removed recipe.

---

### F7 — ADVISORY / LOW — `correctness`
**Both files, path-equality assertions**

All landing-site checks compare `--show-toplevel` outputs as raw strings. Symlinked TMPDIR
(macOS `/tmp` → `/private/tmp`), trailing slashes, or case-insensitive filesystems can produce
false mismatches — which both weakens the runtime assertions and could itself flake the execution
tests. Prefer `[ "$a" -ef "$b" ]` or compare `realpath`-normalized paths. This compounds F2/F3.

---

- Traced the guard's `rev-parse --git-dir` semantics against the stated precondition (plain
  in-repo `.factory/`) → confirmed the guard short-circuits and never moves the dir (F1).
- Confirmed `onboard-observability`'s `--show-toplevel` assertion **is** correct and *does* detect
  the plain-dir case — which is what exposes the inconsistency with `factory-health`'s weaker
  `--git-dir` guard. The right check already exists in the PR; the guard should adopt it.
- Confirmed the `2>/dev/null` in the onboard assertion (Q4) is correct: it makes the LHS empty on a
  missing/non-worktree `.factory`, so the comparison fails and the abort fires — it hides nothing
  the agent needs.
- Assessed the CI flake hypothesis: content-contract greps cannot flake; execution tests create
  real worktrees, so isolation is the plausible vector (F4) — flagged needs-verification rather
  than confirmed, since the `.bats` body is summarized.

## Verdict

**REQUEST_CHANGES.** F1 (guard defeated in its target case) and F2 (`rm -rf` data-loss on
false-positive) are individually blocking. F3–F5 should be fixed in the same pass. Once F1 adopts
the `--show-toplevel` test, the guard, the assertion, and `onboard-observability` all use one
consistent, correct check.

Zious11 (changes requested):
# PR #737 Re-Review — head bf0f13e7

**Title:** fix(skills): assert .factory mounts at repo root; gate onboard-observability on the mount
**Repo:** drbothen/vsdd-factory
**Prior review:** 2026-07-22, prior head ~e828-era (F1–F5)

## Verdict: REQUEST_CHANGES

APPROVE requires BOTH HIGH findings (F1, F2) to be structurally FIXED. F1 is
UNADDRESSED (the flagged flawed line is present verbatim in this head) and F2 is
only PARTIAL (destructive `rm -rf .factory` + missing `git worktree prune`
persist). Verdict criteria not met.

---

## Per-finding closure status

### F1 (HIGH) — plain-dir guard uses `rev-parse --git-dir`, a no-op in its target case
**Status: UNADDRESSED**

The PR added a *post-mount* assertion using `--show-toplevel`, but the
**pre-mount plain-dir guard** — the exact thing F1 flagged — still uses the
flawed `rev-parse --git-dir` test, unchanged:

```bash
if [ -e .factory ] && ! git -C .factory rev-parse --git-dir >/dev/null 2>&1; then
  mv .factory .factory-backup-$(date +%s)   # plain dir, not a worktree
fi
git worktree add .factory factory-artifacts
```

Failure scenario (the guard's stated target case — a plain `.factory/logs/`
left by `/onboard-observability`): `.factory` is a subdirectory *inside* the
repo working tree. `git -C .factory rev-parse --git-dir` therefore walks up,
discovers the parent repo's `.git`, prints it, and **exits 0**. The negation
`! git ... rev-parse --git-dir` is false, so the guard body never runs and the
plain dir is NOT moved aside. `git worktree add .factory factory-artifacts`
then fails with `fatal: '.factory' already exists` — precisely the bug the PR
claims to fix. `rev-parse --git-dir` returns exit 0 for BOTH a real worktree
AND a plain in-repo dir, so it can never distinguish them.

The fix suggested in the prior review (use `--show-toplevel` equality, e.g.
`[ "$(git -C .factory rev-parse --show-toplevel 2>/dev/null)" != "$(git rev-parse --show-toplevel)/.factory" ]`)
was applied only to the separate post-mount assertion, not to the guard. The
guard remains inert.

Note: the new content-contract test (test 3) only greps for the word `plain`
in the SKILL — it does NOT execute the guard, so it cannot detect that the
guard is inert. The 4 execution tests validate the `--show-toplevel` assertion
logic, never the `rev-parse --git-dir` guard. Passing tests here do not
constitute a load-bearing closure (TD-VSDD-059).

### F2 (HIGH) — destructive `rm -rf .factory` recovery, string-comparison fragility, dangling registration
**Status: UNADDRESSED / PARTIAL (not a clean FIX)**

New nested-mount recovery block:

```bash
git worktree remove .factory/.factory --force
rm -rf .factory        # now a leftover plain dir; the guard above re-handles it
```

Improvement: the `rm -rf .factory` is now reached only after the post-mount
assertion fails, rather than being gated inline on a comparison. But the
structural risk F2 named persists:

1. **String-comparison fragility still gates a destructive op.** The gating
   assertion is itself a raw string equality
   `[ "$(git -C .factory rev-parse --show-toplevel)" = "$(git rev-parse --show-toplevel)/.factory" ]`.
   On a symlink/path-canonicalization mismatch this can FALSE-FAIL on a
   *healthy* repo-root worktree, routing a human/agent into the recovery block
   where `rm -rf .factory` then deletes the contents of a live worktree. The
   recovery has no positive confirmation that the layout is actually nested
   (e.g. `test -e .factory/.factory/.git`) before the destructive `rm`. The
   test suite even acknowledges this canonicalization hazard by doing
   `pwd -P` (/var→/private/var) in setup — the production skill has no such
   safeguard.

2. **Dangling registration / missing `git worktree prune` is UNADDRESSED.**
   The recovery contains no `git worktree prune`. If `.factory` itself carries
   a stale worktree registration (the corrupt state #205 describes),
   `rm -rf .factory` leaves a dangling entry in `.git/worktrees` and the next
   `git worktree add .factory` can fail with "already registered". F2 called
   this out explicitly; there is no prune step anywhere in the diff.

3. The trailing comment "the guard above re-handles it" is misleading given
   F1: the guard is inert, so re-running does not mv-aside — the mount only
   succeeds incidentally because `.factory` was just `rm`'d to absent.

### F3 (MEDIUM) — `|| echo ABORT` doesn't halt
**Status: UNADDRESSED**

```bash
[ "$(git -C .factory rev-parse --show-toplevel)" = "$(git rev-parse --show-toplevel)/.factory" ] \
  || echo "ABORT: .factory did not mount at the repo root"
```

Still `|| echo`. This prints and continues; step 3 (`cd .factory && git branch
--show-current`) then reads the parent branch of a nested/plain layout and the
check reports healthy — the exact silent-misbehavior the PR set out to prevent.
Needs `{ echo "ABORT..."; exit 1; }` (or equivalent hard stop).

### F4 (MEDIUM) — worktree tests sharing $REPO as flake source
**Status: FIXED**

`setup()` now creates a fresh `WORK="$(mktemp -d)"` / `REPO="$WORK/repo"` per
test (bats runs `setup()` before each test), and `teardown()` runs
`git -C "$REPO" worktree prune` + `rm -rf "$WORK"`. Each execution test mounts
its own `.factory` in an isolated repo, so the healthy / nested / plain-dir /
absent tests no longer contend for a shared `$REPO`. Load-bearing structural
change; closure accepted.

### F5 (MEDIUM) — backup-restore is comment-only
**Status: UNADDRESSED**

```bash
git worktree add .factory factory-artifacts
# If a backup was made, restore any contents (e.g. logs/) into the worktree:
#   cp -R .factory-backup-*/. .factory/ 2>/dev/null || true
#   rm -rf .factory-backup-*
```

Restore remains commented out. Any data moved to `.factory-backup-<ts>` by the
guard (if the guard ever fired) is silently orphaned; the user is left with a
`.factory-backup-*` dir and no automatic restoration.

---

## New findings at this head

- **N1 (LOW):** Recovery `rm -rf .factory` lacks a positive nested-layout
  confirmation. Add `test -e .factory/.factory/.git` (or a worktree-list check)
  before the destructive `rm`, and add `git worktree prune` after, so a
  false-failed assertion cannot destroy a healthy worktree (ties to F2).
- **N2 (LOW):** onboard-observability Prerequisite 1 presents the mount check
  as a bare `[ ... ]` with prose "abort with ..." but no `|| { echo ...; exit 1; }`.
  Same halt-doesn't-halt shape as F3; as an executable snippet it does not
  enforce the abort.

## Checklist notes

- Diff coherence: all changes relate to the #205/#203 mount-ordering hazard. OK.
- Description accuracy: the "What changed" body claims a working plain-dir
  guard; the shipped guard is inert (F1). Description overstates the fix.
- Test coverage: content-contract tests grep for prose only and cannot detect
  the inert guard or the non-halting abort; execution tests validate
  `--show-toplevel` logic but not the `rev-parse --git-dir` guard or the
  `rm -rf` recovery path. The two HIGH defects are untested.
- Diff size: 205/-3 — reasonable.

## Summary table

| ID | Sev | Status |
|----|-----|--------|
| F1 | HIGH | UNADDRESSED |
| F2 | HIGH | UNADDRESSED / PARTIAL |
| F3 | MEDIUM | UNADDRESSED |
| F4 | MEDIUM | FIXED |
| F5 | MEDIUM | UNADDRESSED |
| N1 | LOW | new |
| N2 | LOW | new |

**REQUEST_CHANGES** — both HIGH findings unresolved.

arcavenai:
All findings addressed at 98ddd188:
- **F1 — guard is no longer inert**: `rev-parse --git-dir` is gone. The plain-dir guard compares canonicalized (`pwd -P`) worktree toplevels — a root mount resolves to `<repo-root>/.factory`, a plain in-repo dir resolves to the parent root — so the mv-aside actually fires in the #203 onboard-first case. A nested `.factory/.factory/.git` layout is detected up front and routed to recovery instead of mounted over.
- **F2 + N1 — `rm -rf` is gone from the skill entirely**: the recovery block requires the positively-confirmed nested shape (`test -e .factory/.factory/.git`), removes only the nested worktree, **moves the wrapper aside instead of deleting it** (its contents are restored by the mount block's backup-restore on re-run), runs `git worktree prune` for the dangling registration, and refuses with a manual-inspection message on any other layout. The mount block also prunes before adding, so a prior bad recovery's stale registration can't fail the re-add. The misleading "guard above re-handles it" comment is gone with the block it described.
- **F3 + N2 — hard stops**: post-mount assertion and the onboard-observability prerequisite both `{ echo …; exit 1; }` (onboard's check also canonicalized so symlinked checkouts don't false-fail into the destructive path — the hazard named in F2.1).
- **F5 — restore is executable**, not commented out; backups are consumed and removed.
- **Tests**: new extract-and-execute suite runs the EXACT fenced blocks shipped in the SKILLs against fabricated layouts — clean mount, plain-dir guard + content restore, nested refuse→recover→re-run heal (wrapper logs survive), recovery refusal on a healthy mount, stale-registration prune, onboard prereq halt/pass. 15/15 green; the two HIGH defects are now load-bearing-tested (TD-VSDD-059).

Ready for re-review.

Zious11 (approved):
# PR #737 Re-Review — head 98ddd188

**Verdict: APPROVE**

All seven findings from the previous REQUEST_CHANGES round (F1–F5, N1, N2) are addressed at this head. One new LOW finding (NF1) is pre-existing and not blocking.

## F1 (HIGH) — ADDRESSED

`rev-parse --git-dir` is gone. The plain-dir guard now uses a `canon()` function (`cd "$dir" 2>/dev/null && pwd -P`) to compare canonicalized `--show-toplevel` outputs:

```bash
if [ -e .factory ] && \
   [ "$(canon "$(git -C .factory rev-parse --show-toplevel 2>/dev/null || echo /nonexistent)")" != "$repo_root/.factory" ]; then
  mv .factory ".factory-backup-$(date +%s)"
fi
```

For a plain `.factory/logs` directory (the #203 onboard-first case), `git -C .factory rev-parse --show-toplevel` walks up and returns the parent repo root — not `<repo>/.factory` — so the `!=` comparison is true and `mv` fires. This is genuinely load-bearing: the new "shipped mount block: plain .factory/logs dir is guarded, mounted over, contents restored" bats test extracts the EXACT fenced snippet from SKILL.md via `_extract_block` and executes it against a fabricated layout. A regression to the inert `rev-parse --git-dir` guard would fail that test.

## F2 + N1 (HIGH + LOW) — ADDRESSED

`rm -rf` is absent from the skill entirely. The recovery block now:

1. Requires positive `[ -e .factory/.factory/.git ]` confirmation before acting (N1 closure).
2. Removes only the nested worktree with `git worktree remove .factory/.factory --force`.
3. Moves the wrapper aside with `mv .factory ".factory-backup-$(date +%s)"` — non-destructive; contents are restored by the mount block's backup-restore loop on re-run.
4. Runs `git worktree prune` to drop the dangling registration.
5. Refuses with `{ echo "ABORT: not the known nested shape — inspect 'git worktree list' manually" >&2; exit 1; }` on any unrecognized layout.

The mount block also runs `git worktree prune` before `git worktree add`, so a stale registration from a prior bad recovery cannot fail the re-add. The post-mount assertion uses `canon()` throughout — no raw string comparison on potentially symlinked paths, eliminating the false-positive-assertion risk that was the core of F2.1.

## F3 (MEDIUM) — ADDRESSED

Post-mount assertion now hard-stops:

```bash
[ "$(canon "$(git -C .factory rev-parse --show-toplevel 2>/dev/null || echo /nonexistent)")" = "$repo_root/.factory" ] \
  || { echo "ABORT: .factory did not mount at the repo root" >&2; exit 1; }
```

Subsequent steps cannot execute on a bad mount.

## F4 (MEDIUM) — CONFIRMED STILL FIXED

`setup()` allocates `WORK="$(mktemp -d)"` with `WORK="$(cd "$WORK" && pwd -P)"` (macOS /var→/private/var resolved) and `REPO="$WORK/repo"` per bats test invocation. `teardown()` scopes `git worktree prune` and removal to that unique `$WORK`. No shared state between tests.

## F5 (MEDIUM) — ADDRESSED

Backup-restore is now an executable loop, not commented-out prose:

```bash
for b in .factory-backup-*; do
  [ -e "$b" ] || continue
  cp -R "$b/." .factory/
  # removes the backup after restore
done
```

Backups are consumed and removed. The `[ -e "$b" ] || continue` guard handles the no-backup case correctly (unmatched glob expands to the literal string; `-e` fails; loop body is skipped).

## N2 (LOW) — ADDRESSED

The onboard-observability prerequisite now hard-stops with `canon()` on both sides:

```bash
canon() { (cd "$1" 2>/dev/null && pwd -P); }
[ "$(canon "$(git -C .factory rev-parse --show-toplevel 2>/dev/null || echo /nonexistent)")" = "$(canon "$(git rev-parse --show-toplevel)")/.factory" ] \
  || { echo ".factory/ is not a mounted worktree — run /factory-health first to set it up, then re-run this skill." >&2; exit 1; }
```

Symlinked checkouts don't false-fail into the abort path.

## NF1 (LOW — pre-existing, not blocking, root fix in #206)

`git worktree list | grep -F '.factory'` is a substring match. When `.factory/.factory` is registered in the worktree list, `grep -F '.factory'` matches it (`.factory` is a substring of `.factory/.factory`) and the skill routes to the "If mounted" branch — which has no nested-layout detection. `git -C .factory branch --show-current` then reads the parent repo's branch (plain `.factory` dir causes git to walk up), which matches the current branch, so the skill exits reporting healthy. This is the "silently misbehaves" scenario the PR's Why section describes. The PR adds nested detection only in the "If missing" path (which catches the sub-case where the nested worktree registration was also lost). This gap pre-dates this PR; the root fix is the dispatcher-side ordering guarantee in #206. Not blocking.

## Tests

The suite grew from 9 to 15 tests by adding 6 shipped-block execution tests. These use `_extract_block` to extract the EXACT fenced bash from SKILL.md and execute it against fabricated git repos (isolated per-test scratch repos via `mktemp -d`). TD-VSDD-059 compliant: both HIGH defects are now load-bearing-tested — a regression to the inert `rev-parse --git-dir` guard or to a destructive recovery block would fail the corresponding execution tests, not merely a content grep.

PR #738, opened 2026-07-20
fix(dispatcher): gate .factory/logs creation on a mounted worktree — stop racing bootstrap

arcavenai:
## Why

The dispatcher's internal log does an unconditional `fs::create_dir_all(<root>/.factory/logs)` on every write, and the dispatcher fires on every tool use. During `/factory-health` first-time setup that means the dispatcher continuously re-plants a plain `.factory/` directory in the window between "checked missing" and `git worktree add .factory factory-artifacts` — and git, finding `.factory` already present, mounts the worktree **nested** at `.factory/.factory` (#206; this is the shared root cause of the #203/#205 bootstrap-failure cluster). The skill-side guards for the symptom half ship separately (companion skills PR adds the mount assertion); this PR removes the process that keeps recreating the conflicting state.

## What changed

- `crates/factory-dispatcher/src/log_dir.rs` — new pure helper `factory_mount_ready(log_dir)`: a `.factory`-parented log dir is writable only when that `.factory` carries a `.git` entry. A `git worktree add` mount has a `.git` *file*; a plain checkout of the artifact branch has a `.git` *dir*; both count. A log dir whose parent is not named `.factory` (a `VSDD_LOG_DIR` / `FACTORY_ROOT` override pointing elsewhere) is always writable — it cannot collide with the mount. An override deliberately pointing at an *unmounted* `.factory/logs` is gated like the resolved shape, since creating the plain dir would cause the same nested mount regardless of who asked; this corner is documented on the helper.
- `crates/factory-dispatcher/src/internal_log.rs` — `write_inner` consults the gate before `create_dir_all`. Suppression is a skip, not an error: a once-per-session stderr notice (shared `AtomicBool` across clones, same pattern as the dedup set) explains the state and points at `/factory-health` or `VSDD_LOG_DIR`. Events during the bootstrap window remain observable via `VSDD_SINK_FILE`.
- Resolution levels A–G themselves are unchanged — this gates the *write*, not the *resolution*, so the log location documented in project docs stays `.factory/logs/dispatcher-internal-YYYY-MM-DD.jsonl`.

## Behavior

| Log-dir provenance | `.factory` state | Before | After |
|---|---|---|---|
| resolved (level C–G) | absent | created as plain dir (the race) | skip + warn once |
| resolved (level C–G) | plain dir, no `.git` (onboard-first state) | `logs/` created inside (blocks the mount) | skip + warn once |
| resolved (level C–G) | mounted worktree (`.git` file) or checkout (`.git` dir) | write | write (unchanged) |
| explicit `VSDD_LOG_DIR` / `FACTORY_ROOT` (level A/B) | any | write | write (unchanged — the gate never overrides the override) |

## Test evidence

- 5 new unit tests on the pure gate (`log_dir.rs`): absent / plain-dir / `.git`-file mount / `.git`-dir checkout / non-`.factory` override.
- 3 new write-path tests (`internal_log.rs`): `.factory` absent → not created; plain `.factory` → no `logs/` child; mounted `.factory` → JSONL written exactly as before.
- Two existing fixtures fabricated plain `.factory` dirs and asserted writes landed — the exact behavior this issue reports as the defect. Updated to the mount shape (`.git` file) their surrounding comments already assumed: `tests/internal_log_integration.rs` (mkdir-p semantics still exercised — `logs/` left absent) and `tests/issue_130_log_dir_shadow.rs` (its shadow assertion is unchanged and still passes).
- `cargo test -p factory-dispatcher`: all 33 test binaries pass. `cargo test --workspace --all-targets`: green except the two pre-existing `validate-state-structure` tests that read the live artifact state (banner drift recorded in the project's own state frontmatter; unrelated to this diff). `cargo fmt --check --all` and `cargo clippy -p factory-dispatcher --all-targets -- -D warnings` clean.

## Notes

- `bin/factory-obs` (`_generate_override`) also does a `mkdir -p "$root/.factory/logs"` outside any worktree check — a second, non-recurring creation site. Left out of this PR to keep it single-concern; the onboard-observability ordering guard in the companion skills PR covers its practical path.
- Operator-level effect lands at the next release: the cached plugin ships a prebuilt dispatcher, so develop-branch changes don't alter running installs until an rc is cut.

## Review round: override exemption (second commit)

As first shipped, the gate also suppressed writes when the log dir came from an **explicit `VSDD_LOG_DIR` / `FACTORY_ROOT` override** pointing at a `.factory/logs`-shaped path — and the first real consumer of that corner turned out to be this repo's own bats harness, which points `VSDD_LOG_DIR` at scratch `.factory/logs` fixtures. 16 tests across `regression-v1.0`, `check-harness-version`, `validate-heavy-op-delegation`, and the trajectory-tail advisory suites failed on empty internal logs (caught in review — thank you). `InternalLog` now takes `with_mount_gate(bool)`; `main.rs` disables the gate whenever either override env var is set and non-empty, so an explicit operator choice is honored verbatim. All four suite groups re-run green locally against the rebuilt release dispatcher; a new unit test (`gate_bypassed_for_explicit_override`) locks the bypass. The remaining red on this PR's first CI run also included the batch-wide `sprint-state-format` def-b failure, which is #724's known issue (fix in flight on a sibling PR) — unrelated to this diff.

Refs: #206
Refs: #203

### DIFF
--- a/crates/factory-dispatcher/src/internal_log.rs
+++ b/crates/factory-dispatcher/src/internal_log.rs
@@ -246,19 +246,36 @@ pub struct InternalLog {
     /// once per dispatcher session instead of once per event. Shared across
     /// clones like `seen_errors`.
     gate_warned: std::sync::Arc<std::sync::atomic::AtomicBool>,
+    /// Whether the #206 mount gate applies. `true` for resolved (level C–G)
+    /// log dirs; `false` when the operator explicitly chose the location via
+    /// `VSDD_LOG_DIR` / `FACTORY_ROOT` (level A/B) — an explicit override
+    /// must not itself be overridden, even when it points at a `.factory/logs`
+    /// path (the bats harness does exactly that with scratch fixtures).
+    mount_gated: bool,
 }
 
 impl InternalLog {
     /// Build a writer rooted at `log_dir`. The directory is NOT created
-    /// eagerly; `write` will `mkdir -p` on first use — and only once the
-    /// `.factory` parent is a mounted worktree (see
-    /// [`crate::log_dir::factory_mount_ready`], issue #206).
+    /// eagerly; `write` will `mkdir -p` on first use — and, by default, only
+    /// once the `.factory` parent is a mounted worktree (see
+    /// [`crate::log_dir::factory_mount_ready`], issue #206). Callers whose
+    /// `log_dir` came from an explicit operator override should disable the
+    /// gate via [`InternalLog::with_mount_gate`].
     pub fn new(log_dir: PathBuf) -> Self {
         Self {
             log_dir,
             seen_errors: std::sync::Arc::new(Mutex::new(HashSet::new())),
             gate_warned: std::sync::Arc::new(std::sync::atomic::AtomicBool::new(false)),
+            mount_gated: true,
         }
+    }
+
+    /// Set whether the #206 mount gate applies. Pass `false` when the log
+    /// dir came from an explicit `VSDD_LOG_DIR` / `FACTORY_ROOT` override.
+    #[must_use]
+    pub fn with_mount_gate(mut self, gated: bool) -> Self {
+        self.mount_gated = gated;
+        self
     }
 
     /// Best-effort append. Never panics, never propagates errors. On
@@ -301,11 +318,13 @@ impl InternalLog {
         // Issue #206: never create `.factory/logs` ahead of the worktree
         // mount. The dispatcher fires on every tool use, so an unconditional
         // mkdir here continuously recreated a plain `.factory/` during
-        // `/factory-health` bootstrap, forcing the later `git worktree add`
-        // to nest at `.factory/.factory` (#203/#205). Suppression is a skip,
-        // not an error — events during the bootstrap window are still
-        // observable via `VSDD_SINK_FILE`.
-        if !crate::log_dir::factory_mount_ready(&self.log_dir) {
+        // `/factory-health` bootstrap — blocking the later `git worktree add`
+        // (`fatal: '.factory' already exists`) and feeding the #203/#205
+        // bootstrap-failure cluster. Suppression is a skip, not an error —
+        // events during the bootstrap window are still observable via
+        // `VSDD_SINK_FILE`. Explicit VSDD_LOG_DIR / FACTORY_ROOT overrides
+        // are exempt (`mount_gated == false`): the operator chose the path.
+        if self.mount_gated && !crate::log_dir::factory_mount_ready(&self.log_dir) {
             if !self
                 .gate_warned
                 .swap(true, std::sync::atomic::Ordering::Relaxed)
@@ -588,6 +607,29 @@ mod tests {
         assert!(
             !factory.join("logs").exists(),
             "plain .factory dir must not gain logs/ — that is the bootstrap race"
+        );
+    }
+
+    /// Issue #206: an explicit override (`with_mount_gate(false)`, the
+    /// VSDD_LOG_DIR / FACTORY_ROOT path) writes even into a plain `.factory`
+    /// with no `.git` — the operator chose the location; the gate must not
+    /// override the override.
+    #[test]
+    fn gate_bypassed_for_explicit_override() {
+        let dir = tempfile::tempdir().unwrap();
+        let factory = dir.path().join(".factory");
+        stdfs::create_dir_all(&factory).unwrap();
+        let log = InternalLog::new(factory.join("logs")).with_mount_gate(false);
+        let ts = Local.with_ymd_and_hms(2026, 1, 15, 9, 30, 0).unwrap();
+
+        log.write(&InternalEvent::with_ts(DISPATCHER_STARTED, ts));
+
+        let expected = factory
+            .join("logs")
+            .join(format!("{FILENAME_PREFIX}2026-01-15{FILENAME_SUFFIX}"));
+        assert!(
+            expected.exists(),
+            "explicit override must write regardless of mount state"
         );
     }
 
--- a/crates/factory-dispatcher/src/internal_log.rs
+++ b/crates/factory-dispatcher/src/internal_log.rs
@@ -246,11 +246,13 @@ pub struct InternalLog {
     /// once per dispatcher session instead of once per event. Shared across
     /// clones like `seen_errors`.
     gate_warned: std::sync::Arc<std::sync::atomic::AtomicBool>,
-    /// Whether the #206 mount gate applies. `true` for resolved (level C–G)
-    /// log dirs; `false` when the operator explicitly chose the location via
-    /// `VSDD_LOG_DIR` / `FACTORY_ROOT` (level A/B) — an explicit override
-    /// must not itself be overridden, even when it points at a `.factory/logs`
-    /// path (the bats harness does exactly that with scratch fixtures).
+    /// Whether the #206 mount gate applies. `true` for resolved (level B–G)
+    /// log dirs; `false` only when the operator explicitly chose the location
+    /// via `VSDD_LOG_DIR` (level A) — an explicit override must not itself be
+    /// overridden, even when it points at a `.factory/logs` path (the bats
+    /// harness does exactly that with scratch fixtures). `FACTORY_ROOT`
+    /// (level B) is deliberately NOT exempt: it resolves to
+    /// `$FACTORY_ROOT/logs`, the racing shape the gate holds back.
     mount_gated: bool,
 }
 
@@ -270,8 +272,9 @@ impl InternalLog {
         }
     }
 
-    /// Set whether the #206 mount gate applies. Pass `false` when the log
-    /// dir came from an explicit `VSDD_LOG_DIR` / `FACTORY_ROOT` override.
+    /// Set whether the #206 mount gate applies. Pass `false` only when the
+    /// log dir came from an explicit `VSDD_LOG_DIR` (level A) override — see
+    /// `crate::log_dir::mount_gate_exempt`.
     #[must_use]
     pub fn with_mount_gate(mut self, gated: bool) -> Self {
         self.mount_gated = gated;
@@ -322,8 +325,8 @@ impl InternalLog {
         // (`fatal: '.factory' already exists`) and feeding the #203/#205
         // bootstrap-failure cluster. Suppression is a skip, not an error —
         // events during the bootstrap window are still observable via
-        // `VSDD_SINK_FILE`. Explicit VSDD_LOG_DIR / FACTORY_ROOT overrides
-        // are exempt (`mount_gated == false`): the operator chose the path.
+        // `VSDD_SINK_FILE`. Only an explicit VSDD_LOG_DIR override is exempt
+        // (`mount_gated == false`): the operator chose that exact path.
         if self.mount_gated && !crate::log_dir::factory_mount_ready(&self.log_dir) {
             if !self
                 .gate_warned
@@ -610,10 +613,10 @@ mod tests {
         );
     }
 
-    /// Issue #206: an explicit override (`with_mount_gate(false)`, the
-    /// VSDD_LOG_DIR / FACTORY_ROOT path) writes even into a plain `.factory`
+    /// Issue #206: an explicit override (`with_mount_gate(false)`, reached
+    /// only via VSDD_LOG_DIR — level A) writes even into a plain `.factory`
     /// with no `.git` — the operator chose the location; the gate must not
-    /// override the override.
+    /// override the override. FACTORY_ROOT does NOT reach this path (#738).
     #[test]
     fn gate_bypassed_for_explicit_override() {
         let dir = tempfile::tempdir().unwrap();
--- a/crates/factory-dispatcher/src/internal_log.rs
+++ b/crates/factory-dispatcher/src/internal_log.rs
@@ -242,15 +242,22 @@ pub struct InternalLog {
     /// Shared via `Arc` so every `clone()` of `InternalLog` participates in the
     /// same dedup window (one process invocation = one dispatcher session).
     seen_errors: std::sync::Arc<Mutex<HashSet<u64>>>,
+    /// One-shot flag so the mount-gate suppression (issue #206) warns exactly
+    /// once per dispatcher session instead of once per event. Shared across
+    /// clones like `seen_errors`.
+    gate_warned: std::sync::Arc<std::sync::atomic::AtomicBool>,
 }
 
 impl InternalLog {
     /// Build a writer rooted at `log_dir`. The directory is NOT created
-    /// eagerly; `write` will `mkdir -p` on first use.
+    /// eagerly; `write` will `mkdir -p` on first use — and only once the
+    /// `.factory` parent is a mounted worktree (see
+    /// [`crate::log_dir::factory_mount_ready`], issue #206).
     pub fn new(log_dir: PathBuf) -> Self {
         Self {
             log_dir,
             seen_errors: std::sync::Arc::new(Mutex::new(HashSet::new())),
+            gate_warned: std::sync::Arc::new(std::sync::atomic::AtomicBool::new(false)),
         }
     }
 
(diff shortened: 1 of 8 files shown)


arcaven:
Heads-up that the `bats-full-suite (linux)` red here looks different from the batch-wide `sprint-state-format` drift on the sibling PRs — this run shows 10 failing suites, and 9 of them (`regression-v1.0`, `validate-heavy-op-delegation`, `check-harness-version`, the index-advisory suites) trip on `[ -n "$log" ]`, i.e. the dispatcher's internal log came back empty. That reads like the new `factory_mount_ready` gate suppressing the write in the bats harness, where `.factory` isn't a `.git`-bearing worktree — the two Rust integration fixtures were updated to fabricate the mount, but the shell regression suite that asserts "dispatcher writes an event for every invocation" wasn't, so it now fails. Might be worth a local `tests/run-all.sh` against the gate: either the harness needs the same mount fabrication, or the gate is suppressing logging in more states than the bootstrap race it targets — the default diagnostic trail would go silent on any non-worktree `.factory`, not just during `/factory-health`. (Separately, since `bin/factory-obs`'s `mkdir -p .factory/logs` is left in scope by design, `Refs: #206` may fit better than `Closes:`.)

arcavenai:
Confirmed and fixed in dfc89a6b — your diagnosis was exact, with one refinement on the mechanism: the failing suites don't rely on default resolution at all; they set `VSDD_LOG_DIR` explicitly at scratch `.factory/logs` fixtures, and the gate was suppressing the **explicit override** (the corner the first commit had 'accepted and documented' — its first real consumer turned out to be your own harness, which settles the design question). `InternalLog` now carries `with_mount_gate(bool)` and `main.rs` disables the gate whenever `VSDD_LOG_DIR`/`FACTORY_ROOT` is set and non-empty; resolved level C–G paths stay gated. Re-ran all four failing suite groups locally against the rebuilt release dispatcher: regression-v1.0 9/9, check-harness-version green (incl. TC-ENV-001), validate-heavy-op-delegation green, trajectory-tail 67/67; crate suite 33/33 binaries, fmt/clippy clean. Also took the trailer suggestion — `Refs: #206` now, since `bin/factory-obs`'s mkdir site stays in scope. (The 17th red in that run was the batch-wide def-b #724 failure, unrelated.)

arcaven:
The `with_mount_gate(bool)` split reads cleanly, and the `main.rs` predicate matches the level-A/B `filter(|s| !s.is_empty())` condition exactly, so the bypass set and the override set can't drift apart. I also went looking for whether the escape hatch could re-open the bootstrap race and couldn't find a live path — the only production `export VSDD_LOG_DIR` is `bin/factory-dashboard`, whose children are shell helpers rather than the dispatcher, and `FACTORY_ROOT` is never exported outside one test; the `git worktree add` in `factory-health` runs with neither set.

One soft follow-on, on the default path rather than this diff: `docs/guide/migrating-from-0.79.md` still has two checklist items that read an empty `dispatcher-internal.jsonl` as "dispatcher not firing" (the "Dispatcher log has entries" item around line 130, and the "Dispatcher not firing / hooks not running" troubleshooting entry around line 222). After this change an unmounted `.factory` produces exactly that symptom for a different reason, and the once-per-session stderr notice may not surface to an operator working through the checklist. A line in each pointing at `/factory-health` may save a wrong diagnosis.

Zious11 (changes requested):
# PR Review — #738 fix(dispatcher): gate .factory/logs creation on a mounted worktree — stop racing bootstrap

**Verdict:** REQUEST_CHANGES
**Base:** develop | **Class:** fix-PR (backlog remediation) | **Refs:** #206, #203

## What I verified

The pure gate (`factory_mount_ready`) in `log_dir.rs` is correct and cleanly handles the described cases (absent, plain-dir, `.git`-file mount, `.git`-dir checkout, non-`.factory` override path). The builder/`with_mount_gate` pattern is sound and thread-safe: `mount_gated` is write-only before `Arc` construction, read-only after; `gate_warned` as `Arc<AtomicBool>` with `swap(true, Relaxed)` is a correct once-per-session print guard. The behavior table rows 1–4 are implemented correctly in the gate logic itself.

## Findings

### [HIGH] `FACTORY_ROOT` in the override bypass reopens the #206 race — `main.rs`

The bypass in `main.rs` disables the mount gate whenever **either** `VSDD_LOG_DIR` **or** `FACTORY_ROOT` is set and non-empty:

```rust
let explicit_override = ["VSDD_LOG_DIR", "FACTORY_ROOT"]
    .iter()
    .any(|k| std::env::var(k).is_ok_and(|v| !v.is_empty()));
InternalLog::new(resolve_log_dir()).with_mount_gate(!explicit_override)
```

The two env vars are not equivalent for this gate:
- `VSDD_LOG_DIR` (level A) points *directly* at a log dir that may legitimately live outside any `.factory`. Bypassing the gate for it is defensible.
- `FACTORY_ROOT` (level B) is a **root**, and `resolve_log_dir()` derives `$FACTORY_ROOT/.factory/logs` from it — i.e., the resulting path is parented on `.factory`, which is exactly the shape `factory_mount_ready` guards. The gate would handle a `FACTORY_ROOT`-derived path correctly on its own.

**Failure scenario:** an operator runs first-time `/factory-health` bootstrap with `FACTORY_ROOT` exported. The gate is disabled; the dispatcher resumes unconditional `create_dir_all($FACTORY_ROOT/.factory/logs)` on every tool use; a plain `.factory` is re-planted before `git worktree add` runs; git mounts nested at `.factory/.factory` — the #206 root cause this PR set out to eliminate.

The bypass should be narrowed to `VSDD_LOG_DIR` only. The stated fixture problem ("this repo's bats harness points `VSDD_LOG_DIR` at scratch `.factory/logs`") only requires the `VSDD_LOG_DIR` exemption. If the harness also relies on `FACTORY_ROOT` scratch fixtures, the production-grade fix is to give those fixtures the `.git` mount shape (as was already done for the two updated fixtures) rather than blanket-bypass the gate.

### [MEDIUM] `bin/factory-obs` retains the same unconditional `mkdir -p` with no tracked follow-up
The Notes section acknowledges `factory-obs` was left out to keep the PR single-concern. Under the project's production-grade default, a bare "left out" is a defer-pattern smell unless anchored to a concrete tracked follow-up. Please open a tracking issue for the `factory-obs` guard and reference it before merge.

### [LOW] Test coverage cements the blanket bypass rather than distinguishing the two env vars
`gate_bypassed_for_explicit_override` locks the bypass for the combined case but does not assert that a `FACTORY_ROOT`-derived path (`$FACTORY_ROOT/.factory/logs`) remains gated. If the bypass is narrowed per the HIGH finding, add a test verifying a `FACTORY_ROOT`-derived path is still subject to the mount check.

### [LOW] Gate trusts any `.factory/.git` entry — possible false "ready"
`factory_mount_ready` returns `true` whenever `parent.join(".git").exists()`. A dangling `.git` file from a de-mounted worktree satisfies the gate even with no live mount. Impact is benign (logs land rather than racing), but worth a doc comment noting the invariant being relied on.

### [INFO] Thread-safety and builder pattern are sound
`with_mount_gate` takes `mut self` and is applied pre-`Arc`, before any sharing; `mount_gated` is read-only thereafter. `gate_warned` `AtomicBool::swap(true, Relaxed)` is correct for a once-per-session print guard. No race.

## Summary
The pure gate logic and the `InternalLog` mechanism are correct and well-tested. The blocking concern is in `main.rs`: folding `FACTORY_ROOT` into the bypass disables the gate for paths that still resolve under `.factory`, which can reintroduce the #206 nesting race during a `FACTORY_ROOT`-configured first-time bootstrap — undercutting the PR's own goal. Narrow the bypass to `VSDD_LOG_DIR`, verify the bats harness doesn't rely on `FACTORY_ROOT` scratch fixtures for this, anchor the deferred `factory-obs` guard to a tracked issue, and add a test asserting `FACTORY_ROOT`-derived paths remain gated.

arcavenai:
Good catch on the operator-facing diagnosis path — folded into this PR as `a9dfdcca` since the symptom is created by this change: both migration-guide items (the "Dispatcher log has entries" checklist item and the "Dispatcher not firing" troubleshooting entry) now say an empty `dispatcher-internal.jsonl` can mean "`.factory` not yet mounted" and point at `/vsdd-factory:factory-health` before the dead-dispatcher diagnosis. Thanks also for the escape-hatch race audit — the `bin/factory-dashboard` observation matches our read: its children are shell helpers, so the override exemption never reaches a dispatcher process during mount setup.

Zious11 (changes requested):
# PR #738 Re-Review — head `a9dfdcca`

**PR:** fix(dispatcher): gate .factory/logs creation on a mounted worktree — stop racing bootstrap
**Repo:** drbothen/vsdd-factory
**Head reviewed:** `a9dfdcca7fa3a22db0c1f7af391f415f33c3deca`
**Reviewer:** pr-reviewer (fresh-eyes re-review after prior REQUEST_CHANGES)
**Verdict: REQUEST_CHANGES** — the HIGH finding remains UNADDRESSED.

---

## Verdict criterion

APPROVE only if the HIGH finding is FIXED. The HIGH finding is **UNADDRESSED**
(the `FACTORY_ROOT` bypass was retained, not narrowed to `VSDD_LOG_DIR`), so the
verdict is **REQUEST_CHANGES**.

Since the prior review, the only new commit at this head (`a9dfdcca`) is a
docs-only change to the migration guide. No code changed. None of the four prior
findings were addressed.

---

## Per-finding closure status

### HIGH — narrow bypass to `VSDD_LOG_DIR` only; keep `FACTORY_ROOT`-derived paths gated — **UNADDRESSED**

`crates/factory-dispatcher/src/main.rs` at head:

```rust
let explicit_override = ["VSDD_LOG_DIR", "FACTORY_ROOT"]
    .iter()
    .any(|k| std::env::var(k).is_ok_and(|v| !v.is_empty()));
let internal_log =
    Arc::new(InternalLog::new(resolve_log_dir()).with_mount_gate(!explicit_override));
```

`FACTORY_ROOT` is still in the bypass array. The recommended fix was not applied.

Why this is a real, still-open defect — grounded in `log_dir.rs` resolution:

- **Level A (`VSDD_LOG_DIR`)** returns the value directly. It is a deliberate
  per-invocation diagnostic override and is the *only* consumer that needs the
  bypass — the repo's own bats harness points `VSDD_LOG_DIR` at scratch
  `.factory/logs` fixtures, which is the documented justification for the
  second commit.
- **Level B (`FACTORY_ROOT`)** returns `$FACTORY_ROOT/logs` (only `/logs` is
  appended — NOT `/.factory/logs`). By the resolution contract `FACTORY_ROOT`
  names the `.factory` directory itself, so the conventional
  `FACTORY_ROOT=<repo>/.factory` yields a log dir whose parent basename is
  `.factory` — exactly the shape `factory_mount_ready` exists to hold back.

Failure scenario: an operator with `FACTORY_ROOT=<repo>/.factory` exported (the
population most likely to be mid-setup) runs first-time `/factory-health`
bootstrap. `explicit_override == true` → `with_mount_gate(false)` → gate
skipped. The dispatcher's per-tool-use `mkdir -p $FACTORY_ROOT/logs` re-plants a
plain `.factory/` in the window before `git worktree add .factory
factory-artifacts`, and git mounts nested at `.factory/.factory` — the #206 race
this PR set out to close, reopened for the FACTORY_ROOT-configured path.

There is no demonstrated need to bypass for `FACTORY_ROOT`: the only cited
consumer (bats) uses `VSDD_LOG_DIR`. Keeping `FACTORY_ROOT` gated costs nothing
in the non-conventional case — if `FACTORY_ROOT` points somewhere NOT named
`.factory`, `factory_mount_ready` already returns `true` and never fires. So
gating `FACTORY_ROOT` only ever engages in precisely the racing case.

Fix: `let explicit_override = std::env::var("VSDD_LOG_DIR").is_ok_and(|v| !v.is_empty());`
(drop `FACTORY_ROOT` from the bypass set), and correct the `main.rs` comment and
the `log_dir.rs` helper doc, which currently claim both level A and B are
exempted. If any harness relies on a `FACTORY_ROOT` scratch fixture, give that
fixture the `.git` mount shape (as was already done for the two updated
fixtures) rather than blanket-bypassing the gate.

### MEDIUM — `bin/factory-obs` unconditional mkdir, no tracked follow-up — **UNADDRESSED**

The PR "Notes" section still acknowledges `bin/factory-obs` (`_generate_override`)
does `mkdir -p "$root/.factory/logs"` outside any worktree check — a second
creation site with the same race potential — and defers it, citing "the
onboard-observability ordering guard in the companion skills PR covers its
practical path." No issue number is referenced and no tracked follow-up appears
in this diff. Under the production-grade default this second creation site
should be fixed in-scope or, at minimum, anchored to a real tracked issue ID; a
prose "covered elsewhere" claim without an anchor is a defer-pattern smell. Not
itself the merge blocker, but still open.

### LOW — test cements the blanket bypass — **UNADDRESSED**

`gate_bypassed_for_explicit_override` (internal_log.rs) asserts
`with_mount_gate(false)` writes into a plain `.factory` regardless of mount
state. While the HIGH finding stands, `main.rs` routes `FACTORY_ROOT` into that
same `with_mount_gate(false)` path, so the suite locks in the over-broad bypass.
There is still no test exercising the `main.rs` env-var branch. When HIGH is
fixed, add coverage asserting a `FACTORY_ROOT`-derived (level B) path stays
gated while a `VSDD_LOG_DIR` (level A) path bypasses.

### LOW — gate trusts any `.factory/.git` entry — **UNADDRESSED**

`factory_mount_ready` returns `parent.join(".git").exists()`. Any `.git` entry —
including a dangling/garbage `.git` file with no valid `gitdir:` pointer —
satisfies the gate. A more robust check would confirm the `.git` file contains a
`gitdir:` line (or that the `.git` dir is a real git dir). Low severity: the
false-shape is unlikely and the failure direction (allowing a write) is the
benign one. Unchanged at this head.

---

## New findings at head `a9dfdcca`

The head commit is docs-only (`docs/guide/migrating-from-0.79.md`): it adds
guidance that a missing/empty internal log can mean an unmounted `.factory`
(run `/vsdd-factory:factory-health`) rather than a broken dispatcher. Accurate
and consistent with the code behavior. No new code-level defects. The
`eprintln!` once-per-session notice follows the file's existing best-effort
stderr diagnostic pattern (matching the dedup path) and is not a new
`println!`-convention violation.

The pure gate design (`factory_mount_ready`, write-path gating, the
`with_mount_gate` builder, the absent/plain/mounted/checkout/override test
matrix) remains correct and well-tested — the thread-safety of `mount_gated`
(write-only pre-`Arc`) and `gate_warned` (`AtomicBool::swap(true, Relaxed)`) is
sound.

---

## Summary

| Finding | Severity | Status |
|---|---|---|
| Narrow bypass to VSDD_LOG_DIR; keep FACTORY_ROOT gated | HIGH | UNADDRESSED |
| factory-obs unconditional mkdir, no tracked follow-up | MEDIUM | UNADDRESSED |
| Test cements blanket bypass | LOW | UNADDRESSED |
| Gate trusts any .factory/.git entry | LOW | UNADDRESSED |

The one merge-blocking issue is that the second-commit override bypass was not
narrowed: `FACTORY_ROOT` still disables the gate and, in its conventional
`.factory`-named form, reopens the #206 race this PR exists to close. This is a
structural gap, not a paper-fix — the code is simply unchanged from the prior
review at the load-bearing line. **REQUEST_CHANGES.**

arcavenai:
All findings addressed at b5c67c12:
- **HIGH — bypass narrowed to `VSDD_LOG_DIR` only**: `FACTORY_ROOT` is out of the exemption. The rule lives in a new unit-testable helper `log_dir::mount_gate_exempt(vsdd_log_dir)`; `main.rs` wires it from the env. Comments/docs in `main.rs`, `log_dir.rs`, and `internal_log.rs` corrected to level-A-only, with the level-B rationale (`$FACTORY_ROOT/logs` under a conventional `FACTORY_ROOT=<repo>/.factory` is exactly the racing shape) stated where the old claim was.
- **LOW — env-branch coverage**: `mount_gate_exempt` tests plus a composition test proving the level-B resolution (`FACTORY_ROOT=<x>/.factory` → `<x>/.factory/logs` via `resolve_log_dir_from_params`) stays gated while `.factory` is plain and is admitted once the worktree mount shape exists.
- **LOW — `.git` marker validated**: `factory_mount_ready` accepts a `.git` *directory* (plain checkout) or a `.git` *file carrying a `gitdir:` pointer*; garbage/empty `.git` files keep the gate closed. New tests for both rejection cases.
- **MEDIUM — second creation site fixed in-scope** (not deferred): `bin/factory-obs` `_generate_override` now pre-creates `logs/` only when `.factory` is mount-shaped, otherwise warns and skips (the override mount line is still written for when the mount lands). Two bats regression tests cover the unmounted-skip and mounted-create paths.

`cargo fmt`/`clippy -D warnings`/workspace tests green; obs bats 21/21 + 81/81. Ready for re-review.

Zious11 (approved):
**Verdict: APPROVE**

All four findings from the prior review are addressed.

**HIGH — FACTORY_ROOT bypass narrowed to VSDD_LOG_DIR only — ADDRESSED**

`main.rs` now calls `mount_gate_exempt(std::env::var("VSDD_LOG_DIR").ok().as_deref())` — `FACTORY_ROOT` is not in the bypass. The composition test `test_factory_root_derived_path_stays_gated_until_mounted` closes the exact failure scenario: with `VSDD_LOG_DIR` unset and `FACTORY_ROOT=<repo>/.factory`, the gate holds the write until the `.git` worktree file exists.

**MEDIUM — `bin/factory-obs` unconditional mkdir — ADDRESSED (fixed in-scope)**

`_generate_override` gates `mkdir -p` on `.git` entry presence; warns and skips otherwise. Two bats regression tests cover both paths.

**LOW — Test coverage cements blanket bypass — ADDRESSED**
`mount_gate_exempt_tests` has dedicated coverage; composition test locks the FACTORY_ROOT-stays-gated invariant.

**LOW — Gate trusts any `.factory/.git` entry — ADDRESSED**
`factory_mount_ready` validates `gitdir:` prefix; garbage/empty file cases covered by new tests.

One soft follow-on: the shell gate in `factory-obs` uses existence (`-e`) rather than validating `gitdir:` content, unlike the Rust gate. Difference is limited to a garbage `.git` file — benign false-positive direction — not a blocker.

Gate logic, test coverage, and Rust conventions are all sound.

Issue #212, opened 2026-06-23
bug(state-manager): transient API connection errors require manual orchestrator retry


arcavenai:
## Summary

When `state-manager` (or any sub-agent) fails mid-execution with a transient API error like `Connection closed mid-response`, the failure is reported to the orchestrator as a generic "the response above may be incomplete" string. There is no structured failure signal, no automatic retry, and no way for the orchestrator to distinguish a transient API failure from a permanent agent error.

The orchestrator must manually:

1. Detect the failure (by inspecting the expected output files / STATE.md state).
2. Compose a fresh retry prompt with all the same context.
3. Re-dispatch as a brand-new `Agent` call.

This is error-prone and wastes tokens — the retry prompt has to re-include all the content the failed dispatch already received.

## Reproduction

ftc-blue pass-8 finalization (2026-06-23):

1. Orchestrator dispatched state-manager with two tasks: persist pass-8 report (~150 lines verbatim) and update STATE.md.
2. Agent returned `API Error: Connection closed mid-response. The response above may be incomplete.` after ~345 seconds.
3. Orchestrator inspected `.factory/cycles/cycle-001/adversarial/p1d-pass-8.md` (not created) and STATE.md (unchanged) to confirm failure.
4. Orchestrator composed a retry prompt (~3000 tokens of duplicated context) and dispatched a fresh state-manager.
5. Retry succeeded.

Net cost: ~5k extra tokens + manual recovery overhead.

## Proposed change

Two options:

**Option A — Auto-retry with idempotency.**

The orchestrator's `Agent` tool, or the agent dispatch framework, should auto-retry on `API Error: Connection closed mid-response` (or other transient errors) up to N times (e.g., 3) before surfacing the failure to the orchestrator. Retries must be idempotent — for state-manager especially, this means the prompt should be safe to re-run.

**Option B — Structured failure surface.**

Wrap the agent return value in a structured failure type:

```
{
  status: "succeeded" | "failed_transient" | "failed_permanent",
  reason?: string,
  output: string,
  partial?: bool
}
```

The orchestrator can then make informed decisions: retry on `failed_transient`, escalate on `failed_permanent`, audit on `partial`.

Option A is preferred for transient errors; Option B for everything else.

## Applies to

- Agent dispatch framework / `Agent` tool internals
- `agents/state-manager.md` — ensure idempotency of common operations (re-running a write to the same path with the same content should be safe)
- `agents/orchestrator/` — failure-handling protocol

[...]

PR #740, opened 2026-07-20
docs(agents): state-manager idempotency contract + orchestrator transient-dispatch recovery protocol

arcavenai:
## Why

Issue #212 reports that when a sub-agent dies mid-flight on a transient upstream
error (`Connection closed mid-response`), the orchestrator gets a raw error
string with no structured signal — and has to manually detect whether work
landed, then re-compose and re-dispatch. Its core asks are Claude Code harness
concerns: auto-retry on transient errors inside the Agent tool, and a structured
failure enum (`succeeded | failed_transient | failed_permanent | partial`)
wrapping the agent return value. Neither is fixable in this repo — there is no
Agent-tool internals or dispatch-framework surface here to change, so this PR
does not attempt them and deliberately adds no retry machinery.

What this repo *does* own is the two acceptance criteria that are documentation:
criterion #3 ("common state-manager operations are documented as idempotent")
and the orchestrator-side failure-handling protocol (the issue's own recommended
diagnostic behavior, `agents/orchestrator/`). This PR lands exactly those two,
as prose guidance, and stops there. When the harness-level enum/auto-retry ship,
the diagnostic protocol here is the fallback the orchestrator uses in the
meantime and the behavior it falls back to for any error the harness doesn't
auto-handle.

## What changed

- `agents/state-manager.md` — new "Idempotency" section after Failure &
  Escalation. States that its common operations are safe to re-run: STATE.md
  updates are set-to-value (not append), artifact writes overwrite the same
  path with the same content, `.factory/` structure creation is create-if-absent,
  and `factory-artifacts` commits are safe to re-attempt (already-committed
  content yields "nothing to commit"). Names the one non-idempotent class
  explicitly — append-style logs (burst-log.md, convergence-trajectory.md rows)
  — and tells the agent to read-then-skip on re-dispatch so a retry does not
  double an entry. This is the honest shape: "idempotent" is not blanket-true
  for this agent, and pretending it is would set up exactly the double-append
  the issue worries about.

- `agents/orchestrator/orchestrator.md` — new "Transient Dispatch Failure
  Recovery" subsection under Failure & Escalation. Codifies the diagnostic
  response the issue's reporter performed by hand: do not treat the raw error
  as success or hard failure; inspect the expected output paths to determine
  what actually landed; then decide (nothing → re-dispatch; everything →
  treat as success, don't re-run; partial → re-dispatch idempotent agents
  whole, or the un-landed portion otherwise). Bounds retries at the existing
  3rd-retry escalation ceiling (reusing the Human Notification table's "Agent
  timeout — 3rd retry" rather than inventing a new limit), and cross-references
  state-manager's Idempotency section so "safe to re-run" has a concrete anchor.

Total ~60 lines of markdown across the two files. No code, no workflow changes,
no new hooks.

## Note on testing

This is prose operating guidance for agents; there is no runtime surface in the
repo to exercise it. The existing `tests/permissions.bats` structural checks
(Tool Access / Failure & Escalation / Remember sections must be present; no
shell in coding-profile agents) still pass 68/68 after the edits, but no test
asserts the *content* of a recovery protocol — that is behavioral guidance a
live orchestrator follows, not a statically-verifiable contract. The harness-level
pieces (#212's auto-retry + structured enum) remain the reporter's to build
where the Agent tool lives.

Refs: #212

### DIFF
--- a/plugins/vsdd-factory/agents/orchestrator/orchestrator.md
+++ b/plugins/vsdd-factory/agents/orchestrator/orchestrator.md
@@ -434,7 +434,10 @@ a hard failure. Diagnose actual state first:
    (e.g. the report file was created; STATE.md shows the phase transition). You
    have read access — use it. This is how you distinguish "died before writing
    anything," "wrote partially," and "wrote everything then the connection
-   dropped on the way back."
+   dropped on the way back." Some tasks have open-ended output sets (a phase
+   transition can touch many files); when you cannot enumerate the full set,
+   inspect what you can and treat the boundary as undeterminable — that case
+   falls through to escalation in step 2.
 2. **Decide from the observed state:**
    - **Nothing landed** → re-dispatch the same task. Agents whose common
      operations are idempotent (state-manager — see its Idempotency section) are
@@ -447,8 +450,10 @@ a hard failure. Diagnose actual state first:
      only the portion that did not land, or escalate if you cannot determine the
      boundary.
 3. **Bound the retries.** Apply the same ceiling as other failures: after the
-   3rd retry without a clean result, escalate to the human per the Human
-   Notification table (Agent timeout — 3rd retry).
+   3rd retry without a clean result, escalate to the human per this file's
+   Failure & Escalation Level 2 ("after 3 retries"). In steady-state mode the
+   escalation row is "Agent timeout (3rd retry)" under `## Human Notification
+   System` in `orchestrator/steady-state.md`.
 
 Prefer diagnosis over blind retry: re-dispatching a task whose work already
 landed wastes tokens and can corrupt append-style state, while assuming success
--- a/plugins/vsdd-factory/agents/orchestrator/orchestrator.md
+++ b/plugins/vsdd-factory/agents/orchestrator/orchestrator.md
@@ -420,6 +420,40 @@ confirmed codifications in `cycles/<cycle>/lessons.md` with `[codified]` tag.
 - **Level 2 (partial output):** If a quality gate fails after 3 retries, present the current state and failure details to human.
 - **Level 3 (escalate):** If a critical prerequisite is missing (no repo, no worktree, model unavailable), stop and report to human immediately.
 
+### Transient Dispatch Failure Recovery
+
+Sometimes a dispatched agent returns a raw transport error rather than a clean
+result — e.g. `API Error: Connection closed mid-response. The response above may
+be incomplete.` This is a transient upstream failure, not an agent logic error,
+and the failed agent may have completed all, some, or none of its work before
+the connection dropped. Do NOT treat the raw error string as either success or
+a hard failure. Diagnose actual state first:
+
+1. **Inspect the expected outputs.** For every file the dispatched task was to
+   produce or modify, check whether it exists and reflects the task
+   (e.g. the report file was created; STATE.md shows the phase transition). You
+   have read access — use it. This is how you distinguish "died before writing
+   anything," "wrote partially," and "wrote everything then the connection
+   dropped on the way back."
+2. **Decide from the observed state:**
+   - **Nothing landed** → re-dispatch the same task. Agents whose common
+     operations are idempotent (state-manager — see its Idempotency section) are
+     safe to re-run verbatim.
+   - **Everything landed** → treat as success; the error was on the return trip
+     only. Do not re-dispatch (a needless re-run risks doubling append-style
+     logs).
+   - **Partial** → for idempotent agents, re-dispatch the whole task (the
+     re-run converges the remaining writes). For non-idempotent work, re-dispatch
+     only the portion that did not land, or escalate if you cannot determine the
+     boundary.
+3. **Bound the retries.** Apply the same ceiling as other failures: after the
+   3rd retry without a clean result, escalate to the human per the Human
+   Notification table (Agent timeout — 3rd retry).
+
+Prefer diagnosis over blind retry: re-dispatching a task whose work already
+landed wastes tokens and can corrupt append-style state, while assuming success
+on a task that died silently loses the work entirely.
+
 ## Remember
 
 **You are a COORDINATOR, not a doer. Every substantive task is delegated via
--- a/plugins/vsdd-factory/agents/state-manager.md
+++ b/plugins/vsdd-factory/agents/state-manager.md
@@ -474,6 +474,32 @@ Do this at every phase gate after writing artifacts. You own the commit — no n
 - **Level 2 (partial output):** If a worktree precondition check fails, report the specific error and recovery command without proceeding.
 - **Level 3 (escalate):** If .factory/ is missing or corrupted and cannot be recovered via worktree commands, stop and report to orchestrator.
 
+## Idempotency
+
+Your common operations are safe to re-run. If the orchestrator re-dispatches you
+after a transient failure (e.g. a dropped connection mid-response left it unsure
+whether your writes landed), re-executing the same task against the same paths
+produces the same end state — it does not duplicate or corrupt what already
+persisted:
+
+- **STATE.md updates** are set-to-value, not append. Phase transitions, file
+  sizes, finding counts, and gate verdicts overwrite the field to the value the
+  task specifies; applying the same transition twice leaves STATE.md identical
+  to applying it once.
+- **Artifact persistence** (writing a report or cycle file to a given absolute
+  path) overwrites that path with the provided content; re-running writes the
+  same bytes.
+- **`.factory/` structure creation** is create-if-absent; re-running skips any
+  directory that already exists.
+- **Git commits** to `factory-artifacts` are safe to re-attempt: if the content
+  is already committed, `git add -A && git commit` reports "nothing to commit"
+  rather than producing a duplicate.
+
+The one non-idempotent class is **append-style logs** — burst-log.md entries and
+convergence-trajectory.md per-pass rows. Re-running a task that appends can
+double the entry. When re-dispatched for an append task, first read the target
+file and skip the append if that entry is already present.
+
 ## Templates
 
 - Pipeline state: `../../templates/state-template.md`
--- a/plugins/vsdd-factory/agents/state-manager.md
+++ b/plugins/vsdd-factory/agents/state-manager.md
@@ -495,10 +495,14 @@ persisted:
   is already committed, `git add -A && git commit` reports "nothing to commit"
   rather than producing a duplicate.
 
-The one non-idempotent class is **append-style logs** — burst-log.md entries and
-convergence-trajectory.md per-pass rows. Re-running a task that appends can
-double the entry. When re-dispatched for an append task, first read the target
-file and skip the append if that entry is already present.
+The one non-idempotent class is **append-style records** — burst-log.md
+entries, convergence-trajectory.md per-pass rows, and the STATE.md **Current
+Phase Steps** row that the Burst-complete protocol above appends on the same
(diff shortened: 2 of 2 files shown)


arcaven:
One small thing on the state-manager Idempotency section — it reads cleanly, and scoping the non-idempotent class to the append logs is the right call. The one spot that might be worth a word: the "STATE.md updates are set-to-value, not append" line, since the burst-complete protocol just above (`state-manager.md:134`) also appends a row to STATE.md's "Current Phase Steps (keep last 5)" on the same event that writes burst-log.md — so a verbatim re-dispatch could double that row too (until the 5-row window evicts it). Might be worth folding Current Phase Steps into the same read-then-skip note as burst-log, so the "safe to re-run" framing stays airtight. Minor, and the eviction window mostly covers it either way.

arcavenai:
Good catch — folded in 8c6799b0: the non-idempotent caveat now names all three append-style records (burst-log.md entries, convergence-trajectory.md rows, and the STATE.md Current Phase Steps row that the Burst-complete protocol appends on the same event), with a note that the keep-last-5 window eventually evicts a duplicate but a verbatim re-run doubles the row until it does. The read-then-skip guidance now covers STATE.md sections as well as the cycle files.

Zious11 (changes requested):
# PR #740 Review — docs(agents): state-manager idempotency contract + orchestrator transient-dispatch recovery

**Verdict: REQUEST_CHANGES**

**Reviewer standard applied:** docs-only PR → documented contracts must match *actual* agent behavior (technical-writer standard: describe current behavior, never aspirational). Findings below were verified against the actual agent files that ship in this repo (`plugins/vsdd-factory/agents/state-manager.md` and `plugins/vsdd-factory/agents/orchestrator/*`), which are outside the `.factory/` information wall.

## Summary

The intent is sound and the scope is honest (it explicitly declines the harness-level asks in issue #212 and lands only what this repo owns). The prose is clear and well-organized. However, three of the concrete factual claims do not match the documented behavior of the agents they describe, and two of those undercut the very safety property the sections exist to establish. Because a docs-only PR's entire value is accuracy, and because all fixes are small in-scope edits, this is REQUEST_CHANGES rather than APPROVE.

## Findings

### [HIGH] Non-idempotent enumeration omits `lessons.md` (and other append-style artifacts) — the safety warning is incomplete

`state-manager.md` — new "Idempotency" section, closing paragraph.

The section states: "The one non-idempotent class is **append-style records** — burst-log.md entries, convergence-trajectory.md per-pass rows, and the STATE.md Current Phase Steps row..."

The state-manager agent file's own update-event list (lines 134–136) enumerates **four** append events on the same trigger surface:

1. Burst complete → append `burst-log.md`  ✅ listed
2. Current Phase Steps row  ✅ listed
3. Adversary pass → append `convergence-trajectory.md`  ✅ listed
4. **Lesson learned → "Append to `cycles/<cycle>/lessons.md`"**  ❌ **omitted**

`lessons.md` is a high-frequency append (`L-EDP1-NNN` entries) and is explicitly documented as append in the same file. The 4 INDEX files and `decision-log.md` also receive appended/added rows. A re-dispatched lesson-capture task would double the lesson entry — the exact corruption this section warns against — with no warning here. Because the paragraph is phrased as a closed set ("The **one** non-idempotent class is ... A, B, C"), a reader treats it as exhaustive.

**Suggestion:** Add `lessons.md` to the enumeration, and either enumerate the INDEX/decision-log appends or replace the closed list with "including but not limited to" plus a pointer to the agent's update-event list as the authoritative source.

### [MEDIUM] STATE.md "set-to-value, not append" lead clause overclaims and is contradicted by the same section

`state-manager.md` — new "Idempotency" section, first bullet.

Bullet 1 opens: "**STATE.md updates** are set-to-value, not append." The section's own final paragraph then carves out "the STATE.md **Current Phase Steps** row" as append-style. A reader who acts on the general lead clause alone (STATE.md is set-to-value → safe to re-run) would wrongly re-dispatch a burst-complete task and double the Current Phase Steps row. The specific fields the bullet then lists (phase transitions, file sizes, finding counts, gate verdicts) are indeed set-to-value, so the fix is to scope the lead clause.

**Suggestion:** Change the lead to "**Most STATE.md fields** are set-to-value, not append" (or "The STATE.md fields below are set-to-value"), so the general claim no longer collides with the documented Current Phase Steps exception.

### [MEDIUM] `git add -A` idempotency claim ignores broad-staging of unintended / concurrent changes

`state-manager.md` — new "Idempotency" section, git-commits bullet.

The bullet claims commits "are safe to re-attempt: if the content is already committed, `git add -A && git commit` reports 'nothing to commit' rather than producing a duplicate." The `git add -A` the agent actually uses stages **every** change in the worktree, not just the task's intended files. The "nothing to commit" guarantee only holds when the tree is clean relative to HEAD. In the failure scenario this section is written for — a dropped connection where a dead agent may have written partial/unintended files, or where concurrent bursts have touched the tree — a verbatim re-run can sweep unrelated changes into the commit. That is not idempotent; it is "idempotent only if the working tree is otherwise clean."

**Suggestion:** Qualify the claim: idempotent *when the worktree contains only this task's changes*; note that `git add -A` stages all pending changes, so a re-dispatch should verify the tree scope (or the burst protocol/`state-burst` skill should own the staging boundary).

### [MEDIUM] Fragile, imprecise cross-reference to the "Human Notification table"

`orchestrator.md` — new "Transient Dispatch Failure Recovery" subsection, step 3.

The text says "escalate to the human per the Human Notification table (Agent timeout — 3rd retry)." Verification: the target exists but **in a sibling file** — `orchestrator/steady-state.md`, under the heading `## Human Notification System`, row `| Agent timeout (3rd retry) | ESCALATION | Any phase |`. Three drift points:

- **No file pointer.** A reader inside `orchestrator.md` finds no "Human Notification table" in that file; the reference is unqualified.
- **Wording drift.** Cited as "Human Notification **table** (Agent timeout **—** 3rd retry)"; actual is heading "Human Notification **System**" and row "Agent timeout **(**3rd retry**)**". This is the exact drift risk called out in the review brief.
- **Scope mismatch.** The table lives in the steady-state sequence; the new recovery protocol is written as general dispatch guidance for any mode. Citing a steady-state-scoped table from a general protocol may mislead in non-steady-state runs.

**Suggestion:** Point explicitly to `steady-state.md` "Human Notification System" and match the row label verbatim ("Agent timeout (3rd retry)"), or, if the escalation ceiling is meant to be mode-general, reference the orchestrator's own `Failure & Escalation` Level 2 ("after 3 retries") which lives in the same file.

### [LOW] "Inspect the expected outputs" assumes an enumerable output set

`orchestrator.md` — recovery subsection, step 1.

Phase transitions can touch many files and agent outputs are often open-ended, so "for every file the dispatched task was to produce or modify" is not always knowable to the orchestrator. This is acceptable as prose because step 2/step 3 degrade gracefully ("escalate if you cannot determine the boundary"), but the instruction slightly overstates the orchestrator's ability to know the full output set a priori.

**Suggestion:** Add a half-sentence acknowledging outputs may be open-ended and that partial-knowledge cases fall through to escalation.

### [LOW] "You have read access — use it" is unqualified

`orchestrator.md` — recovery subsection, step 1.

For the orchestrator specifically this is accurate (it has read access), so the risk is low. Noting only that the instruction is stated as a blanket property; info-asymmetry agents (e.g. holdout-evaluator) are separate agents and unaffected, so no change is strictly required.

### [NIT] Placement of the Idempotency section

Placing "Idempotency" immediately after "Failure & Escalation" is reasonable — idempotency is recovery-adjacent and is consumed by the orchestrator's recovery protocol. No change needed; flagged only because the review brief asked.

- Confirmed the orchestrator `Failure & Escalation` ceiling ("after 3 retries") exists (orchestrator.md L412) and that the named notification row exists in steady-state.md (L136) — so the cross-reference is fragile, not broken.
- Confirmed `lessons.md` is documented as an append operation in state-manager.md (L136), establishing the HIGH enumeration gap.
- Confirmed state-manager uses `git add -A` in its commit blocks (L413/423/455), establishing the broad-staging concern.
- Confirmed the STATE.md Current Phase Steps row is keep-last-5 append (L134, L111), establishing the internal-contradiction concern in bullet 1.
- Confirmed structural sections (Tool Access / Failure & Escalation / Remember) are preserved, consistent with the reported 68/68 permissions.bats pass.

## Path to APPROVE

All findings are small in-scope prose edits (add `lessons.md` to the list; scope the two overclaim clauses; fix the cross-reference file pointer + wording). No code, no test, no scope expansion required. Re-request review after the edits.

arcavenai:
All requested changes addressed at 9679811d:
- **Set-to-value claim scoped**: "Most STATE.md fields" with the Current Phase Steps row named as the exception.
- **`git add -A` qualified**: applies only when the worktree contains just this task's changes, with a `git -C .factory status --porcelain` pre-check and escalation otherwise.
- **Append class opened**: "including but not limited to" + the concrete artifact list + an authoritative pointer to the agent's own update-event list.
- **Orchestrator steps 1/3**: step 1 covers open-ended outputs; step 3 routes to Failure & Escalation Level 2 and names the steady-state 'Agent timeout (3rd retry)' row.

(No dismissal rights on this repo, so flagging here instead: ready for re-review.)

Zious11 (approved):
## Re-review at 9679811d — APPROVE

All four CHANGES_REQUESTED findings are addressed:

**[HIGH] lessons.md omission — closed.** The idempotency section now uses "including but not limited to" (open list, not the prior closed set) and explicitly names `cycles/<cycle>/lessons.md` L-entries. It also adds an authoritative pointer: "treat every append trigger in this agent's own update-event list as non-idempotent." Cross-checked the pointer against the full update-event list (state-manager.md lines 131–138): all six numbered append events are covered — four by explicit naming (burst-log.md, convergence-trajectory.md, lessons.md, Current Phase Steps), two by the pointer (blocking-issues-resolved.md, session-checkpoints.md). The decision-log.md and 4 INDEX files are named explicitly even though they are not in the numbered event list.

**[MEDIUM] STATE.md "set-to-value" overclaim — closed.** Changed to "**Most STATE.md fields** are set-to-value, not append" with a parenthetical naming the Current Phase Steps exception inline.

**[MEDIUM] git add -A boundary — closed.** Now explicitly qualified: "safe to re-attempt *when the worktree contains only this task's changes*," with a `git -C .factory status --porcelain` pre-check directive and an escalation path if the tree is not clean. The prior unconditional "nothing to commit" claim is gone.

**[MEDIUM] Cross-reference drift — closed.** Step 3 now reads: "escalate to the human per this file's Failure & Escalation Level 2 ('after 3 retries'). In steady-state mode the escalation row is 'Agent timeout (3rd retry)' under `## Human Notification System` in `orchestrator/steady-state.md`." File pointer, heading label, and row label are all verbatim-correct.

**[LOW] Open-ended output sets — closed.** Step 1 now includes: "Some tasks have open-ended output sets (a phase transition can touch many files); when you cannot enumerate the full set, inspect what you can and treat the boundary as undeterminable — that case falls through to escalation in step 2."

---

One new non-blocking observation: the "Partial" case in orchestrator step 2 says "for idempotent agents, re-dispatch the whole task." State-manager falls into this bucket by implication, which is functionally correct — its own read-then-skip guards make the whole-task re-dispatch safe for partials. But the explanatory "(state-manager — see its Idempotency section)" pointer only appears in the "Nothing landed" bullet above it, not in the partial-case text. The combined protocol is correct end-to-end; a future pass could surface the nuance explicitly in the partial-case branch as well.

TD-VSDD-053 and D-417(b) checks: clean.

PR #768, opened 2026-08-04
ci: pin workflow actions to release commit shas

arcaven:
A tag ref like actions/checkout@v6 is mutable: whoever controls that tag controls code that runs with the workflow's token. Pinning to release commit SHAs removes that surface, and semgrep (config=auto) stops flagging the workflows: a full scan over .github/workflows at this branch reports 0 findings across 82 rules.

This replaces the 22 tag refs across ci.yml, release.yml, and semgrep.yml (counted with `grep -rhoE 'uses:\s*\S+' .github/workflows/*.yml` at develop 948f0fb1). Each pin is the commit SHA of the action's current latest release, resolved via `git ls-remote --tags` (peeled where annotated), with the version in a trailing comment so Renovate/Dependabot can keep tracking it:

| action | was | now |
|---|---|---|
| actions/checkout | v6 | 3d3c42e5 (v7.0.1) |
| actions/upload-artifact | v4 | 043fb46d (v7.0.1) |
| actions/download-artifact | v4 | 3e5f45b2 (v8.0.1) |
| softprops/action-gh-release | v2 | 3d0d9888 (v3.0.2) |
| dtolnay/rust-toolchain | 1.95.0 | 2c7215f1 (master) + explicit `toolchain: 1.95.0` input |

dtolnay/rust-toolchain publishes no releases and normally derives the toolchain from the ref name. At the 1.95.0 branch head its action.yml has no toolchain input, so a bare SHA pin would break selection; at master (2c7215f1) toolchain is a required input. The four sites gain `toolchain: 1.95.0`, so the installed toolchain is unchanged.

One caveat to weigh: checkout v6 to v7, upload-artifact v4 to v7, download-artifact v4 to v8, and gh-release v2 to v3 are major bumps to current releases. ci.yml and semgrep.yml are exercised by this PR's own checks; release.yml only runs on a release, so its upload/download/gh-release sites are untested here. Happy to drop those three to the latest release within their current majors instead (v4.6.2 / v4.3.0 / v2.6.2) if you'd rather separate behavior changes from pinning.

Local gates at develop 948f0fb1: cargo fmt --check, clippy -D warnings, cargo test --workspace (two pre-existing failures in validate-state-structure that read the live .factory/STATE.md banner count; they fail identically on clean develop and your CI is green at the same SHA), hooks.bats + bin.bats, and SEMGREP_BASELINE_REF=upstream/develop semgrep ci with 0 findings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
   validate:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Mount factory artifacts
         run: |
@@ -111,12 +111,13 @@ jobs:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Install Rust toolchain
         # Pinned in rust-toolchain.toml — keep this version in sync with that file.
-        uses: dtolnay/rust-toolchain@1.95.0
+        uses: dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c # master 2026-08-04
         with:
+          toolchain: 1.95.0
           components: rustfmt, clippy
           targets: wasm32-wasip1
 
@@ -323,7 +324,7 @@ jobs:
     # knows exactly which side to refresh.
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Install yq + python3-yaml
         run: |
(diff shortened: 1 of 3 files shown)


arcaven:
Reviewing my own PR after 39 days, since nothing has run on it and I would rather refresh it than leave it sitting. Three things, all measured against develop `ebd16f79` today.

**Still needed.** 28 of the 35 `uses:` refs across the three workflows on develop are still mutable tag refs (17 `actions/checkout@v6`, 6 `dtolnay/rust-toolchain@1.95.0`, 2 upload-artifact, 2 download-artifact, 1 gh-release), so nothing has superseded this.

**But the coverage has drifted, and the body no longer describes a merged tree.** develop's `ci.yml` has grown from 14 `uses:` at my branch point to 23. GitHub still reports `MERGEABLE / CLEAN` after 41 commits, which is the part worth flagging: a clean merge would pin the refs I touched and silently leave the roughly 9 that arrived since, so the headline "22 tag refs" and the "0 findings across 82 rules" semgrep result are both accurate as captured and stale as a description of the result. I would rather rebase and re-measure than have you merge a half-done pin.

**No CI has ever run on it.** Zero workflow runs and zero check runs on head `81b6267f`. The pattern across the fork suggests why: #729, #737, #738 and #740 are all `ArcavenAE` fork PRs and all got 14 checks, and #729 touches a workflow file too, so it is not the workflow-file edit. The difference is the author. Those four are `arcavenai`, which has merged PRs here; this one is `arcaven`, which has none. That looks like an Actions approval gate scoped to first-time contributors, in which case it is waiting on a maintainer and no notification says so. Worth knowing independently of this PR, since it means any future PR from my human account is invisible to CI by default.

Also worth weighing before a rebase: the pins were resolved to then-current releases on 2026-08-04 and carry major bumps (checkout v6 to v7, upload-artifact v4 to v7, download-artifact v4 to v8, gh-release v2 to v3). If you would rather have the pins without the major bumps, say so and I will pin each action at the SHA of the version currently in use instead, which makes this a pure supply-chain change with no behavior delta.

Happy to rebase, re-resolve and re-measure on a word from you. Equally happy for you to close it if you would rather do the pinning in-pipeline.

arcaven:
Correction to my third point above. I attributed the missing CI to a first-time-contributor gate that `arcavenai` had already cleared by merging. The run timestamps say otherwise, and the difference matters for what the fix is.

All four `arcavenai` PRs were opened on 2026-07-20. Their first workflow runs fired on 2026-07-22 and 2026-07-23, two to three days later, and #737, #738 and #740 all started within one second of each other at 23:17:04-05Z on the 22nd. Automatic dispatch on `pull_request` fires in seconds, not days, and never lands three separate PRs on the same second. That is someone clicking approve on a batch.

So the gate is not one an account ages out of. Neither `arcaven` nor `arcavenai` is a member here, and every fork PR from an outside contributor needs a maintainer click before any check runs, each time, on each push. #768 has simply never had that click, and the same will be true of the rebase I offered.

That makes it less about this PR than about the contribution path: if you would rather not click per push, granting one of the accounts triage or write would let CI run on its own. Either way is fine by me; the useful part is that it is a standing cost rather than a one-time initiation, and nothing in the UI says so from the contributor's side.

PR #774, opened 2026-08-09
fix(dispatcher): raise WASM fuel cap 10M→20M + fuel-vs-epoch block_reason disambiguation

Zious11:
## Fix: ADR-042 v1.2 / D-964 — WASM Fuel Cap + Disambiguation

**Authority:** ADR-042 v1.2, ADR-040 v1.1, D-964(b)/(h)
**Severity:** HIGH — silent hook validation chain failure under production load

### Problem

The WASM plugin fuel cap was set to 10M units. Measured fuel-exhaustion events reached 838→1,226 in a single session across **35 distinct plugins**, silently disabling the hook validation chain on large `.factory/` writes.

Root cause: worst-case payload (ARCH-INDEX + 50 KB `last_assistant_message`, 377,109 bytes) consumed **10,406,058 fuel** — exceeding the 10M cap by ~4%. The failure surfaced to agents as an ambiguous `fail-closed: plugin timed out`, indistinguishable from a real wall-clock timeout.

Two distinct fixes and one load-bearing test guard in this PR:

### Commit 1: `182dfc68` — Fuel cap 10M → 20M

- `InvokeLimits::default().fuel_cap`: 10M → 20M (`crates/factory-dispatcher/src/invoke.rs`)
- `RegistryDefaults::default().fuel_cap`: 10M → 20M (`crates/factory-dispatcher/src/registry.rs`)
- Existing registry test updated to match the new default
- Two new regression tests: `invoke_limits_default_fuel_cap_is_20m` and `fuel_workload_exhausts_10m_completes_at_20m` (WAT arithmetic loop @ ~10.8M fuel verifies both the exhaustion at 10M and success at 20M)
- Comment updates in `validate-index-cite-refresh` Cargo.toml + lib.rs and `validate-state-structure` lib.rs: update stale "10M budget" references to "20M" (ADR-042 §Decision 1)

### Commit 2: `7cbb9232` — Fuel-vs-epoch `block_reason` disambiguation

- `extract_reason_from_outcome()` in `crates/factory-dispatcher/src/main.rs`: split the single `PluginResult::Timeout` arm into two arms by `TimeoutCause`
  - `TimeoutCause::Fuel` → `"fail-closed: fuel exhausted ({N} units consumed; raise fuel_cap or reduce payload size)"`
  - `TimeoutCause::Epoch` → unchanged `"fail-closed: plugin timed out"`
- Three new regression tests in `tests_extract_reason_cause_distinction`: fuel arm, epoch arm (confirms no regression), crash arm (confirms no regression)
- Imports `TimeoutCause` in `main.rs`

### Commit 3: `4bd626c5` — Cross-field sync guard for ADR-042 fuel cap

- `fuel_cap_defaults_stay_in_sync` test in `crates/factory-dispatcher/src/registry.rs`
- Pins a named constant `ADR_042_FUEL_CAP: u64 = 20_000_000` and asserts **both** `InvokeLimits::default().fuel_cap` and `RegistryDefaults::default().fuel_cap` against it
- Rationale: the two structs (`InvokeLimits` and `RegistryDefaults`) carry independent `fuel_cap` fields that must always agree. Prior to this commit, only a prose comment bound them — a future change to one field would silently diverge without any test failure. Pinning against a named constant (rather than asserting bare equality) also catches co-drift — the case where both constants are reset to the same wrong value, which is the realistic regression when a cap change is reverted wholesale
- Closes a `pr-reviewer` NIT in-scope per TD-VSDD-060 (sibling-site sweep) rather than deferring it
- Load-bearing character verified by perturbation: test passes on correct code, FAILS when one constant is set to `10_000_000`, and passes again after revert. The perturbation was not committed — verified via `git show --stat` (38 insertions, 0 deletions in one file)

### Why This Matters

Landing on `develop` is the prerequisite for a later cross-compile release that rebuilds the bundled binaries under `plugins/vsdd-factory/hooks/dispatcher/bin/` and the operator cache at `1.0.0-rc.23`. The binaries currently embed the 10M cap.

### Testing

- [x] All existing tests pass (`cargo test --workspace --all-targets`)
- [x] New test `invoke_limits_default_fuel_cap_is_20m`: asserts default is 20M
- [x] New test `fuel_workload_exhausts_10m_completes_at_20m`: WAT loop at ~10.8M fuel — fails at 10M cap (Timeout::Fuel), passes at 20M cap (Ok)
- [x] New tests `fuel_timeout_reason_identifies_fuel_exhaustion`, `epoch_timeout_reason_is_unaffected`, `crash_reason_is_unaffected`: full coverage of the three `extract_reason_from_outcome` arms
- [x] New test `fuel_cap_defaults_stay_in_sync`: pins `ADR_042_FUEL_CAP = 20_000_000` and asserts both `InvokeLimits::default().fuel_cap` and `RegistryDefaults::default().fuel_cap` against it — catches independent-field divergence that bare equality cannot detect

### Security Note

This is a WASM sandbox resource-limit change. The fuel cap raise increases the maximum CPU-equivalent work a single plugin invocation can perform before being terminated. Security review (CWE-400/CWE-770 DoS surface) is requested.

### DIFF
--- a/crates/factory-dispatcher/src/executor.rs
+++ b/crates/factory-dispatcher/src/executor.rs
@@ -685,6 +685,7 @@ fn emit_lifecycle(
             stderr,
             elapsed_ms,
             fuel_consumed,
+            fuel_cap: _,
         } => {
             let cause_str = match cause {
                 TimeoutCause::Epoch => "epoch",
@@ -748,6 +749,7 @@ fn emit_lifecycle(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::invoke::DEFAULT_FUEL_CAP;
 
     // ── CRIT-PR59-001 regression tests: advisory-block gate ──────────────────
 
@@ -840,6 +842,7 @@ mod tests {
             stderr: String::new(),
             elapsed_ms: 5_000,
             fuel_consumed: 0,
+            fuel_cap: DEFAULT_FUEL_CAP,
         };
         assert!(!plugin_requests_block(&r));
     }
@@ -891,6 +894,7 @@ mod tests {
             stderr: String::new(),
             elapsed_ms: 5_000,
             fuel_consumed: 0,
+            fuel_cap: DEFAULT_FUEL_CAP,
         };
         assert!(
             plugin_fail_closed(&r, OnError::Block),
@@ -901,11 +905,13 @@ mod tests {
     /// Timeout + on_error=Continue → NOT fail-closed.
     #[test]
     fn fail_closed_timeout_with_on_error_continue_is_open() {
+        // fuel_consumed == fuel_cap on Trap::OutOfFuel (remaining=0 → cap.saturating_sub(0)=cap).
         let r = PluginResult::Timeout {
             cause: TimeoutCause::Fuel,
             stderr: String::new(),
             elapsed_ms: 5_000,
-            fuel_consumed: 1_000_000_000,
+            fuel_consumed: DEFAULT_FUEL_CAP,
+            fuel_cap: DEFAULT_FUEL_CAP,
         };
         assert!(
             !plugin_fail_closed(&r, OnError::Continue),
--- a/crates/factory-dispatcher/src/executor.rs
+++ b/crates/factory-dispatcher/src/executor.rs
@@ -685,7 +685,7 @@ fn emit_lifecycle(
             stderr,
             elapsed_ms,
             fuel_consumed,
-            fuel_cap: _,
+            fuel_cap,
         } => {
             let cause_str = match cause {
                 TimeoutCause::Epoch => "epoch",
@@ -704,6 +704,7 @@ fn emit_lifecycle(
                         "stderr".to_string(),
                         serde_json::Value::String(stderr.clone()),
                     ),
+                    ("fuel_cap".to_string(), serde_json::Value::from(*fuel_cap)),
                 ],
             )
         }
@@ -932,6 +933,70 @@ mod tests {
         assert!(
             !plugin_fail_closed(&r, OnError::Block),
             "Ok result + on_error=Block must NOT trigger fail-closed (advisory path handles Ok)"
+        );
+    }
+
+    // ── emit_lifecycle telemetry field coverage ────────────────────────────
+
+    /// `plugin.timeout` events must carry both `fuel_cap` and `fuel_consumed`
+    /// so operators can see which budget was hit without opening the full trace.
+    /// CLAUDE.md Factory Hook Diagnostics Step 2 sends operators to grep
+    /// `.factory/logs/dispatcher-internal-*.jsonl` for exactly these fields.
+    #[test]
+    fn emit_lifecycle_timeout_carries_fuel_cap_and_consumed() {
+        let dir = tempfile::tempdir().expect("tempdir");
+        let log = InternalLog::new(dir.path().join("logs"));
+        let base_ctx = crate::host::HostContext::new("test-plugin", "0.1.0", "sess-1", "trace-1");
+        let entry = RegistryEntry {
+            name: "test-plugin".to_string(),
+            event: "PreToolUse".to_string(),
+            tool: None,
+            plugin: std::path::PathBuf::from("test.wasm"),
+            priority: None,
+            enabled: true,
+            timeout_ms: None,
+            fuel_cap: None,
+            on_error: None,
+            capabilities: None,
+            config: toml::Value::Table(toml::Table::new()),
+            async_flag: false,
+            needs_context: vec![],
+        };
+        let cap: u64 = DEFAULT_FUEL_CAP;
+        let consumed: u64 = 12_345_678;
+        let result = PluginResult::Timeout {
+            cause: TimeoutCause::Fuel,
+            stderr: String::new(),
+            elapsed_ms: 10,
+            fuel_consumed: consumed,
+            fuel_cap: cap,
+        };
+        emit_lifecycle(&log, &base_ctx, &entry, &result);
+
+        // Read back the JSONL and verify both fields are present.
+        let log_dir = dir.path().join("logs");
+        let files: Vec<_> = std::fs::read_dir(&log_dir)
+            .expect("log dir must exist after write")
+            .map(|e| e.expect("dir entry").path())
+            .collect();
+        assert_eq!(files.len(), 1, "expected exactly one log file");
+        let content = std::fs::read_to_string(&files[0]).expect("read log file");
+        let event: serde_json::Value =
+            serde_json::from_str(content.trim_end()).expect("log line must be valid JSON");
+        assert_eq!(
+            event["type"].as_str(),
+            Some(PLUGIN_TIMEOUT),
+            "event type must be plugin.timeout"
+        );
+        assert_eq!(
+            event["fuel_consumed"].as_u64(),
+            Some(consumed),
+            "plugin.timeout event must carry fuel_consumed"
+        );
+        assert_eq!(
+            event["fuel_cap"].as_u64(),
+            Some(cap),
+            "plugin.timeout event must carry fuel_cap (CLAUDE.md Diagnostics Step 2)"
         );
     }
 }
--- a/crates/factory-dispatcher/src/invoke.rs
+++ b/crates/factory-dispatcher/src/invoke.rs
@@ -268,7 +268,11 @@ impl Default for InvokeLimits {
     fn default() -> Self {
         Self {
             timeout_ms: 5_000,
-            fuel_cap: 10_000_000,
+            // ADR-042 §Decision 1: raised 10M → 20M (measurement-validated).
+            // Worst-case legacy-bash-adapter payload consumed 10,406,058 fuel
+            // (ARCH-INDEX + 50 KB last_assistant_message, 377,109 payload bytes),
+            // exhausting the former 10M cap. 20M leaves ~9.6M headroom (92% margin).
+            fuel_cap: 20_000_000,
         }
     }
 }
@@ -1055,6 +1059,81 @@ mod tests {
 
     fn bare_ctx() -> HostContext {
         HostContext::new("plugin", "0.0.1", "sess", "trace")
+    }
+
+    // ADR-042 §Decision 1: global fuel cap raised 10M → 20M (measurement-validated).
+    // 838 fuel-exhaustion events across 35 plugins confirmed by perf-engineer;
+    // worst-case payload (ARCH-INDEX + 50 KB assistant message) measured 10,406,058 fuel.
+    #[test]
+    fn invoke_limits_default_fuel_cap_is_20m() {
+        assert_eq!(InvokeLimits::default().fuel_cap, 20_000_000);
+    }
+
+    // Demonstrates that the 10M→20M raise resolves the class of exhaustions seen in
+    // production. The WAT loop runs 1,200,000 iterations at ~9 WASM instructions each
+    // (≈10.8M fuel) — above the former 10M cap, below the new 20M cap.
+    // Calibrated against the perf-engineer model:
+    //   fuel = 29,452 + 27.514 × payload_bytes (R²=0.9999999).
+    // At 377,109 bytes (ARCH-INDEX + 50 KB assistant msg): 10,406,058 fuel.
+    #[test]
+    fn fuel_workload_exhausts_10m_completes_at_20m() {
+        let engine = build_engine().unwrap();
+        // Arithmetic loop: 1,200,000 iterations x ~9 instructions = ~10.8M fuel.
+        let module = compile(
+            &engine,
+            r#"
+            (module
+              (memory (export "memory") 1)
+              (func (export "_start")
+                (local $i i32)
+                (local.set $i (i32.const 0))
+                (block $done
+                  (loop $l
+                    (local.set $i (i32.add (local.get $i) (i32.const 1)))
+                    (br_if $done (i32.ge_u (local.get $i) (i32.const 1200000)))
+                    (br $l)))))
+            "#,
+        );
+
+        // At the former 10M cap: fuel exhaustion.
+        let res_10m = invoke_plugin(
+            &engine,
+            &module,
+            bare_ctx(),
+            b"",
+            InvokeLimits {
+                timeout_ms: 30_000,
+                fuel_cap: 10_000_000,
+            },
+        )
+        .unwrap();
+        assert!(
+            matches!(
+                res_10m,
+                PluginResult::Timeout {
+                    cause: TimeoutCause::Fuel,
+                    ..
+                }
+            ),
+            "expected Timeout{{Fuel}} at 10M cap, got {res_10m:?}"
+        );
+
+        // At the new 20M cap: completes successfully.
+        let res_20m = invoke_plugin(
+            &engine,
+            &module,
+            bare_ctx(),
+            b"",
+            InvokeLimits {
+                timeout_ms: 30_000,
+                fuel_cap: 20_000_000,
+            },
+        )
+        .unwrap();
+        assert!(
+            matches!(res_20m, PluginResult::Ok { exit_code: 0, .. }),
+            "expected Ok(exit_code=0) at 20M cap, got {res_20m:?}"
+        );
     }
 
     #[test]
--- a/crates/factory-dispatcher/src/invoke.rs
+++ b/crates/factory-dispatcher/src/invoke.rs
@@ -226,7 +226,18 @@ pub enum PluginResult {
         /// a partial message because the plugin was interrupted.
         stderr: String,
         elapsed_ms: u64,
+        /// Instructions executed before the interrupt. Useful telemetry for
+        /// the event log. Note: on `Trap::OutOfFuel`, `fuel_consumed_from_store`
+        /// returns `cap.saturating_sub(remaining)` where `remaining == 0`, so
+        /// `fuel_consumed == fuel_cap` on every fuel trap — it does not convey
+        /// demand independently of the cap. Use `fuel_cap` for the operator
+        /// message when communicating what limit was hit.
         fuel_consumed: u64,
+        /// The configured fuel budget at the time of invocation. Always
+        /// reflects the actual cap set on the wasmtime Store, independent of
+        /// whether `get_fuel()` succeeds. Use this field — not `fuel_consumed`
+        /// — when constructing operator-facing messages about which cap to raise.
+        fuel_cap: u64,
     },
     Crashed {
         trap_string: String,
@@ -409,6 +420,7 @@ pub fn invoke_plugin(
             &stderr,
             elapsed_ms,
             fuel_consumed,
+            limits.fuel_cap,
         ),
     }
 }
@@ -419,6 +431,7 @@ fn classify_trap(
     stderr: &MemoryOutputPipe,
     elapsed_ms: u64,
     fuel_consumed: u64,
+    fuel_cap: u64,
 ) -> Result<PluginResult, InvokeError> {
     let stderr_text = stderr_to_string(stderr);
     // WASI `exit(n)` propagates as an `I32Exit` in wasmtime-wasi's
@@ -441,12 +454,14 @@ fn classify_trap(
                 stderr: stderr_text,
                 elapsed_ms,
                 fuel_consumed,
+                fuel_cap,
             },
             Trap::OutOfFuel => PluginResult::Timeout {
                 cause: TimeoutCause::Fuel,
                 stderr: stderr_text,
                 elapsed_ms,
                 fuel_consumed,
+                fuel_cap,
             },
             other => PluginResult::Crashed {
                 trap_string: other.to_string(),
--- a/crates/factory-dispatcher/src/invoke.rs
+++ b/crates/factory-dispatcher/src/invoke.rs
@@ -255,6 +255,18 @@ pub enum TimeoutCause {
     Fuel,
 }
 
+/// ADR-042 §Decision 1: global fuel cap raised 10M → 20M (measurement-validated).
+/// Worst-case legacy-bash-adapter payload consumed 10,406,058 fuel
+/// (ARCH-INDEX + 50 KB last_assistant_message, 377,109 payload bytes),
+/// exhausting the former 10M cap. 20M leaves ~9.6M headroom (92% margin).
+///
+/// This constant is the single source of truth for the ADR-042 fuel cap.
+/// Both `InvokeLimits::default()` and `RegistryDefaults::default()` reference it,
+/// so any future deliberate cap change is made in one place and propagates
+/// atomically to both — structural enforcement rather than test-only enforcement.
+/// Any change to this value requires updating the ADR-042 §Decision 1 rationale.
+pub const DEFAULT_FUEL_CAP: u64 = 20_000_000;
+
 /// Per-invocation budget. Defaults live in
 /// `RegistryDefaults`; callers usually get these from a
 /// `RegistryEntry` with fallback.
@@ -268,11 +280,7 @@ impl Default for InvokeLimits {
     fn default() -> Self {
         Self {
             timeout_ms: 5_000,
-            // ADR-042 §Decision 1: raised 10M → 20M (measurement-validated).
-            // Worst-case legacy-bash-adapter payload consumed 10,406,058 fuel
-            // (ARCH-INDEX + 50 KB last_assistant_message, 377,109 payload bytes),
-            // exhausting the former 10M cap. 20M leaves ~9.6M headroom (92% margin).
-            fuel_cap: 20_000_000,
+            fuel_cap: DEFAULT_FUEL_CAP,
         }
     }
 }
@@ -1066,19 +1074,21 @@ mod tests {
     // worst-case payload (ARCH-INDEX + 50 KB assistant message) measured 10,406,058 fuel.
     #[test]
     fn invoke_limits_default_fuel_cap_is_20m() {
-        assert_eq!(InvokeLimits::default().fuel_cap, 20_000_000);
+        assert_eq!(InvokeLimits::default().fuel_cap, DEFAULT_FUEL_CAP);
     }
 
-    // Demonstrates that the 10M→20M raise resolves the class of exhaustions seen in
-    // production. The WAT loop runs 1,200,000 iterations at ~9 WASM instructions each
-    // (≈10.8M fuel) — above the former 10M cap, below the new 20M cap.
-    // Calibrated against the perf-engineer model:
-    //   fuel = 29,452 + 27.514 × payload_bytes (R²=0.9999999).
-    // At 377,109 bytes (ARCH-INDEX + 50 KB assistant msg): 10,406,058 fuel.
+    // Proves the dispatcher enforces the fuel ceiling at the boundary: a synthetic WAT
+    // workload of ~10.8M fuel crosses from trapped to completing when the ceiling
+    // moves from 10M to 20M. This is a ceiling-enforcement test, not a production
+    // model: the WAT loop runs with an empty payload and no host calls, so it does
+    // not reflect the regression formula measured by the perf-engineer
+    // (fuel = 29,452 + 27.514 × payload_bytes). The ~10.8M estimate is based on
+    // the loop structure (1,200,000 iterations × ~9 WASM instructions each); the
+    // actual observed value is asserted in the 20M branch below.
     #[test]
     fn fuel_workload_exhausts_10m_completes_at_20m() {
         let engine = build_engine().unwrap();
-        // Arithmetic loop: 1,200,000 iterations x ~9 instructions = ~10.8M fuel.
+        // Arithmetic loop: 1,200,000 iterations x ~9 instructions ≈ 10.8M fuel.
         let module = compile(
             &engine,
             r#"
@@ -1126,14 +1136,31 @@ mod tests {
             b"",
             InvokeLimits {
                 timeout_ms: 30_000,
-                fuel_cap: 20_000_000,
+                fuel_cap: DEFAULT_FUEL_CAP,
             },
         )
         .unwrap();
-        assert!(
-            matches!(res_20m, PluginResult::Ok { exit_code: 0, .. }),
-            "expected Ok(exit_code=0) at 20M cap, got {res_20m:?}"
-        );
+        // Assert both the exit code and the observed fuel so a future wasmtime
+        // accounting change fails with a message stating the actual cost rather
+        // than an opaque pattern-match mismatch.
+        match res_20m {
+            PluginResult::Ok {
+                exit_code,
+                fuel_consumed,
+                ..
+            } => {
+                assert_eq!(exit_code, 0, "expected exit_code=0 at 20M cap");
+                assert!(
+                    fuel_consumed > 10_000_000,
+                    "fuel_consumed should be > 10M (workload is ~10.8M), got {fuel_consumed}"
+                );
+                assert!(
+                    fuel_consumed < DEFAULT_FUEL_CAP,
+                    "fuel_consumed should be < cap ({DEFAULT_FUEL_CAP}), got {fuel_consumed}"
+                );
+            }
+            other => panic!("expected Ok(exit_code=0) at 20M cap, got {other:?}"),
+        }
     }
 
     #[test]
--- a/crates/factory-dispatcher/src/invoke.rs
+++ b/crates/factory-dispatcher/src/invoke.rs
@@ -208,6 +208,12 @@ pub fn dispatch_postcompact() {
 /// and have to re-run with a manual capture to find out why a plugin
 /// exited 1. Field added in v1.0.0-beta.4 after the S-2.7 dogfood loop
 /// ran into exactly that diagnostic gap.
+/// `#[non_exhaustive]` allows adding new variants (e.g. a future `Abandoned`
+/// discriminant for S-1.6 async partition semantics) without a breaking
+/// change for external match sites. Follows the `ResolverError` precedent
+/// in `resolver.rs` (~L78): applied at the enum level so external crates must
+/// include a wildcard arm; in-crate matches are unaffected.
+#[non_exhaustive]
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub enum PluginResult {
     Ok {
--- a/crates/factory-dispatcher/src/invoke.rs
+++ b/crates/factory-dispatcher/src/invoke.rs
@@ -208,12 +208,6 @@ pub fn dispatch_postcompact() {
 /// and have to re-run with a manual capture to find out why a plugin
 /// exited 1. Field added in v1.0.0-beta.4 after the S-2.7 dogfood loop
 /// ran into exactly that diagnostic gap.
-/// `#[non_exhaustive]` allows adding new variants (e.g. a future `Abandoned`
-/// discriminant for S-1.6 async partition semantics) without a breaking
-/// change for external match sites. Follows the `ResolverError` precedent
-/// in `resolver.rs` (~L78): applied at the enum level so external crates must
-/// include a wildcard arm; in-crate matches are unaffected.
-#[non_exhaustive]
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub enum PluginResult {
     Ok {
--- a/crates/factory-dispatcher/src/lib.rs
+++ b/crates/factory-dispatcher/src/lib.rs
@@ -55,7 +55,7 @@ pub use internal_log::{
     PLUGIN_LOADED, PLUGIN_TIMEOUT,
 };
 pub use invoke::{
-    EventType, InvokeError, InvokeLimits, PluginResult, StoreData, TimeoutCause,
+    DEFAULT_FUEL_CAP, EventType, InvokeError, InvokeLimits, PluginResult, StoreData, TimeoutCause,
     dispatch_postcompact, dispatch_precompact, invoke_plugin,
 };
 pub use partition::{PluginPartition, partition_plugins};
--- a/crates/factory-dispatcher/src/main.rs
+++ b/crates/factory-dispatcher/src/main.rs
@@ -832,14 +832,27 @@ fn extract_reason_from_outcome(result: &PluginResult) -> Option<String> {
         // reduce payload); epoch/wall-clock is a transient compute failure
         // (investigate slow script). Conflating them forces operators to grep
         // the internal log for the trace UUID to determine cause — TD #71 pattern.
+        //
+        // Fuel arm uses the FUEL_EXHAUSTED: stable prefix so consumers can match
+        // a fixed token without breaking when the message wording changes. This
+        // mirrors the pattern used elsewhere in the codebase (e.g.
+        // MULTI_COMMIT_CHAIN_NOT_ALLOWED). The epoch arm string "fail-closed:
+        // plugin timed out" is unchanged — existing operator runbooks match it
+        // and changing it would break them.
+        //
+        // On Trap::OutOfFuel, wasmtime's remaining-fuel counter reaches zero, so
+        // fuel_consumed_from_store() == fuel_cap (cap.saturating_sub(0) = cap).
+        // The interpolated value therefore represents the configured cap, not a
+        // measured cost. The message is framed accordingly ("fuel cap of N units")
+        // rather than "N units consumed" which would imply a metered demand figure.
         PluginResult::Crashed { .. } => Some("fail-closed: plugin crashed".to_owned()),
         PluginResult::Timeout {
             cause: TimeoutCause::Fuel,
             fuel_consumed,
             ..
         } => Some(format!(
-            "fail-closed: fuel exhausted ({fuel_consumed} units consumed; \
-             raise fuel_cap or reduce payload size)"
+            "FUEL_EXHAUSTED: fuel cap of {fuel_consumed} units exhausted; \
+             raise fuel_cap or reduce payload size"
         )),
         PluginResult::Timeout {
             cause: TimeoutCause::Epoch,
@@ -1161,33 +1174,47 @@ mod tests_extract_reason_cause_distinction {
     #![allow(clippy::expect_used, clippy::unwrap_used)]
     use factory_dispatcher::invoke::{PluginResult, TimeoutCause};
 
-    // RED before fix: fuel timeout must produce a fuel-specific reason, not "plugin timed out".
+    // Fuel timeout must produce a fuel-specific reason, not "plugin timed out".
+    //
+    // The test input uses fuel_consumed == 20_000_000 to reflect the real invariant:
+    // on Trap::OutOfFuel, wasmtime's remaining-fuel counter is zero, so
+    // fuel_consumed_from_store() == fuel_cap (cap.saturating_sub(0) = cap).
+    // A value like 10_500_000 that differs from any plausible cap is unreachable
+    // in production and would make assertions on the interpolated value misleading.
     #[test]
     fn fuel_timeout_reason_identifies_fuel_exhaustion() {
+        use factory_dispatcher::invoke::DEFAULT_FUEL_CAP;
         let result = PluginResult::Timeout {
             cause: TimeoutCause::Fuel,
             stderr: String::new(),
             elapsed_ms: 5,
-            fuel_consumed: 10_500_000,
+            // consumed == cap on fuel trap (see invariant comment above).
+            fuel_consumed: DEFAULT_FUEL_CAP,
         };
         let reason = super::extract_reason_from_outcome(&result);
         let text = reason
             .as_deref()
             .expect("fuel timeout must produce Some reason");
+        // Stable prefix: consumers match FUEL_EXHAUSTED: to distinguish fuel from epoch.
         assert!(
-            text.contains("fuel exhausted"),
-            "fuel timeout block_reason must contain 'fuel exhausted', got: {text:?}"
+            text.starts_with("FUEL_EXHAUSTED:"),
+            "fuel timeout block_reason must start with 'FUEL_EXHAUSTED:', got: {text:?}"
+        );
+        assert!(
+            text.contains("exhausted"),
+            "fuel timeout block_reason must contain 'exhausted', got: {text:?}"
         );
         assert!(
             !text.contains("timed out"),
             "fuel timeout block_reason must NOT say 'timed out' — \
              operators need distinct messages to choose the right remedy; got: {text:?}"
         );
-        // Consumed-fuel figure surfaces for scale assessment.
+        // The interpolated value is the cap (consumed == cap on fuel trap).
+        // Assert against the actual cap constant so this tracks deliberate changes.
         assert!(
-            text.contains("10500000"),
-            "fuel timeout block_reason should include the consumed-fuel figure \
-             (10500000) so operators can assess scale without log grep; got: {text:?}"
+            text.contains(&DEFAULT_FUEL_CAP.to_string()),
+            "fuel timeout block_reason should include the cap value ({DEFAULT_FUEL_CAP}) \
+             so operators know which cap to raise; got: {text:?}"
         );
     }
 
--- a/crates/factory-dispatcher/src/main.rs
+++ b/crates/factory-dispatcher/src/main.rs
@@ -848,10 +848,10 @@ fn extract_reason_from_outcome(result: &PluginResult) -> Option<String> {
         PluginResult::Crashed { .. } => Some("fail-closed: plugin crashed".to_owned()),
         PluginResult::Timeout {
             cause: TimeoutCause::Fuel,
-            fuel_consumed,
+            fuel_cap,
             ..
         } => Some(format!(
-            "FUEL_EXHAUSTED: fuel cap of {fuel_consumed} units exhausted; \
+            "FUEL_EXHAUSTED: fuel cap of {fuel_cap} units exhausted; \
              raise fuel_cap or reduce payload size"
         )),
         PluginResult::Timeout {
@@ -1190,6 +1190,8 @@ mod tests_extract_reason_cause_distinction {
             elapsed_ms: 5,
             // consumed == cap on fuel trap (see invariant comment above).
             fuel_consumed: DEFAULT_FUEL_CAP,
+            // fuel_cap is the configured cap; message interpolates this field.
+            fuel_cap: DEFAULT_FUEL_CAP,
         };
         let reason = super::extract_reason_from_outcome(&result);
         let text = reason
@@ -1218,15 +1220,54 @@ mod tests_extract_reason_cause_distinction {
         );
     }
 
+    // Proves the fix is load-bearing, not cosmetic: on the `Err(_)` path of
+    // `fuel_consumed_from_store` (when `store.get_fuel()` returns an error),
+    // `fuel_consumed` is 0 but the cap is still known via `fuel_cap`.  Before
+    // the `fuel_cap` field was added, the operator-facing message read
+    // "fuel cap of 0 units exhausted" — a false statement.  This test asserts
+    // that the correct configured cap is reported regardless of whether
+    // `fuel_consumed` carries a meaningful value.
+    #[test]
+    fn fuel_timeout_with_zero_consumed_still_reports_cap() {
+        use factory_dispatcher::invoke::DEFAULT_FUEL_CAP;
+        // Simulates the Err(_) arm of fuel_consumed_from_store: consumed=0, cap=DEFAULT.
+        let result = PluginResult::Timeout {
+            cause: TimeoutCause::Fuel,
+            stderr: String::new(),
+            elapsed_ms: 0,
+            fuel_consumed: 0,
+            fuel_cap: DEFAULT_FUEL_CAP,
+        };
+        let reason = super::extract_reason_from_outcome(&result);
+        let text = reason
+            .as_deref()
+            .expect("fuel timeout must produce Some reason");
+        assert!(
+            text.starts_with("FUEL_EXHAUSTED:"),
+            "message must carry FUEL_EXHAUSTED: prefix even when fuel_consumed=0; got: {text:?}"
+        );
+        assert!(
+            text.contains(&DEFAULT_FUEL_CAP.to_string()),
+            "message must report the configured cap ({DEFAULT_FUEL_CAP}), \
+             not fuel_consumed (0); got: {text:?}"
+        );
+        assert!(
+            !text.contains(": 0 "),
+            "message must NOT interpolate fuel_consumed=0 as the cap; got: {text:?}"
+        );
+    }
+
     // Epoch timeout must still produce the original "plugin timed out" sentinel —
     // changing this arm would break existing operator runbooks for slow scripts.
     #[test]
     fn epoch_timeout_reason_is_unaffected() {
+        use factory_dispatcher::invoke::DEFAULT_FUEL_CAP;
         let result = PluginResult::Timeout {
             cause: TimeoutCause::Epoch,
             stderr: String::new(),
             elapsed_ms: 5_001,
             fuel_consumed: 2_000_000,
+            fuel_cap: DEFAULT_FUEL_CAP,
         };
         let reason = super::extract_reason_from_outcome(&result);
         assert_eq!(
--- a/crates/factory-dispatcher/src/main.rs
+++ b/crates/factory-dispatcher/src/main.rs
@@ -790,8 +790,6 @@ fn extract_block_info(outcomes: &[PluginOutcome]) -> (String, String) {
             PluginResult::Crashed { .. } | PluginResult::Timeout { .. } => {
                 outcome.on_error == OnError::Block
             }
-            // Future variants are not fail-closed unless explicitly handled.
-            _ => false,
         };
 
         if is_blocking {
@@ -860,8 +858,6 @@ fn extract_reason_from_outcome(result: &PluginResult) -> Option<String> {
             cause: TimeoutCause::Epoch,
             ..
         } => Some("fail-closed: plugin timed out".to_owned()),
-        // Future variants: no reason available without opening the internal log.
-        _ => None,
     }
 }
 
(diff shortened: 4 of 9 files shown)


Zious11 (commented):
## Security Review — PR #774: WASM fuel cap 10M→20M + fuel-vs-epoch block_reason disambiguation

**Reviewed at:** `4bd626c5`  
**Verdict:** NO BLOCKERS. Two findings documented below; neither prevents merge.

---

### Scope

Two substantive changes:

1. `InvokeLimits::default().fuel_cap` and `RegistryDefaults::default().fuel_cap` raised from 10,000,000 → 20,000,000 (both in `invoke.rs` and `registry.rs`).
2. `extract_reason_from_outcome` in `main.rs` distinguishes fuel-exhaustion timeout from epoch (wall-clock) timeout in the operator-visible `block_reason` string, emitting the `fuel_consumed` count in the fuel-exhaustion path.

---

### Finding 1 — CWE-400 / CWE-770: Resource Exhaustion Surface Expansion from Fuel Cap Raise

**Severity: LOW**

Doubling the fuel cap from 10M to 20M increases the compute a single WASM plugin invocation can consume before the fuel trap fires. This expands the per-dispatch resource budget by 2×.

**Independent epoch (wall-clock) bound:** `invoke_plugin` at `invoke.rs:350–352` sets both the epoch deadline and the fuel cap before execution, in sequence and independently:

```
store.set_epoch_deadline(timeout_ms_to_epochs(limits.timeout_ms as u64));  // line 350
store.set_fuel(limits.fuel_cap)                                             // line 352
```

The epoch deadline is enforced by the `EpochTicker` — a background thread that increments the engine's epoch counter (confirmed by the test at `invoke.rs:1082–1084`: `let _ticker = EpochTicker::start(engine.clone())`). This is wasmtime's preemptive interrupt mechanism; it fires asynchronously without the WASM guest cooperating. The epoch timeout remains at `timeout_ms: 5_000` (5 seconds, unchanged). A plugin cannot exceed 5 seconds of wall-clock time regardless of how much fuel it would have remaining.

**Justification for 20M:** The diff provides:
- Empirical measurement: worst-case payload consumed 10,406,058 fuel (ARCH-INDEX + 50 KB assistant message, 377,109 bytes).
- Linear model: `fuel = 29,452 + 27.514 × payload_bytes` (R²=0.9999999), confirmed over 838 exhaustion events across 35 plugins.
- 20M provides ~9.6M headroom (92% margin above worst-case measured payload).
- The regression test `fuel_workload_exhausts_10m_completes_at_20m` validates that the 10.8M-fuel WAT loop fails under 10M and succeeds under 20M.

**Residual risk:** No per-plugin memory cap or I/O cap is visible in this diff. If a future plugin's fuel consumption grows significantly beyond 20M (e.g., due to larger `.factory/` artifacts or more complex logic), another exhaustion episode will recur. This is an operational sizing concern, not a security blocker. The measurement-validated approach used here is the correct pattern when it recurs.

**Conclusion:** The DoS surface expansion is real but bounded by the independent 5-second epoch timeout, and the 20M value is empirically grounded. Severity: LOW.

---

### Finding 2 — CWE-209 / CWE-497: Fuel Count Exposed in Operator-Visible block_reason

**Severity: INFO**

The new `block_reason` string for fuel exhaustion:

```
"fail-closed: fuel exhausted ({fuel_consumed} units consumed; raise fuel_cap or reduce payload size)"
```

exposes the raw `fuel_consumed` value. Given the near-perfect linear relationship `fuel ≈ 29,452 + 27.514 × payload_bytes`, this value is effectively a proxy for `.factory/` artifact payload size in bytes (precision ~1 byte).

This value is surfaced to:
- The LLM agent session (via stderr block_reason)
- The dispatcher internal JSONL log
- Terminal stderr output

**In the current deployment context:** vsdd-factory is an operator-facing, single-tenant tool. The agent consuming the block_reason already has direct read access to all `.factory/` artifacts — it can observe file sizes trivially by other means. There is no confidentiality boundary being crossed; no information is disclosed to a party that does not already have access.

**Caveat for future extension:** If this dispatcher is ever extended to a multi-tenant context (multiple isolated workspaces sharing a dispatcher process), fuel counts in block_reason would constitute a side channel on workspace artifact sizes. Operators extending vsdd-factory in that direction should treat `fuel_consumed` as sensitive and omit it from agent-visible output.

**Conclusion:** Not a blocker in the current single-tenant deployment context. INFO-level note for future multi-tenant extension.

---

### Fail-Closed Integrity — CONFIRMED SAFE

The `TimeoutCause::Fuel` and `TimeoutCause::Epoch` arms in `extract_reason_from_outcome` both return `Some(reason)`. Neither arm changes to `None`. The blocking logic downstream of this function is not modified. There is no fail-closed → fail-open conversion in this diff. Both timeout causes remain blocking.

---

### Additional Observations (no security impact)

- The `validate-index-cite-refresh` Cargo.toml comment update correctly notes that the regex crate's DFA compilation exhausts even the 20M cap — this is accurate and the hand-rolled scanning approach remains justified.
- The `fuel_cap_defaults_stay_in_sync` test in `registry.rs` pins both `InvokeLimits` and `RegistryDefaults` against the ADR-042 constant, preventing silent co-drift. This is a sound defensive testing pattern.
- The `validate-state-structure` comment change is cosmetic ("former 10M fuel budget"). No behavioral change.

---

### Summary

| Finding | CWE | Severity | Blocks merge? |
|---------|-----|----------|---------------|
| Fuel cap raise doubles per-dispatch compute budget | CWE-400, CWE-770 | LOW | No — epoch timeout is independently enforced; 20M is measurement-justified |
| `fuel_consumed` in block_reason is a payload-size proxy | CWE-209, CWE-497 | INFO | No — no confidentiality boundary crossed in single-tenant context |
| Fail-closed integrity of timeout arms | — | PASS | No change |

**The PR is safe to merge as-is.**

Zious11 (commented):
# Fresh-eyes review — PR #774 (head `4bd626c5`)

**Verdict: APPROVE-WITH-NOTES.** No blockers. The core change is correct and, importantly, *effective* — I verified the mechanism reaches production rather than only the unit tests (see "Verified as sound" below). Three MEDIUM findings concern claims made by comments and tests that the code does not actually deliver; two of them (MEDIUM-1, MEDIUM-2) are small in-scope edits I'd prefer to see landed before merge rather than deferred.

Scope of this review: `gh pr diff 774` only, plus targeted greps against the baseline tree to check three claims the diff makes but cannot self-evidence (registry override presence, `TimeoutCause` arity, `fuel_consumed` semantics). I did not read prior review reports or cycle artifacts.

---

## Verified as sound (not assumed)

These are the load-bearing correctness questions, and each checks out:

1. **The raise actually reaches all plugins.** `plugins/vsdd-factory/hooks-registry.toml` contains **zero** `fuel_cap` keys. Every hook entry therefore falls through `HookEntry::fuel_cap()` → `self.fuel_cap.unwrap_or(defaults.fuel_cap)` → the raised `RegistryDefaults::default()`. Had any entry pinned `fuel_cap = 10000000`, the raise would have been a no-op for that plugin. It is not.

2. **The `TimeoutCause` split genuinely discriminates.** `TimeoutCause` has exactly two variants (`Epoch`, `Fuel`). The two new arms are therefore exhaustive with no wildcard, meaning a future third variant would be a compile error rather than a silent fall-through to the wrong message. That is the right shape for this match.

3. **The epoch arm is genuinely unchanged.** `"fail-closed: plugin timed out"` is byte-identical to the pre-change string, and `epoch_timeout_reason_is_unaffected` asserts it with `assert_eq!` on the whole string rather than a `contains`. Same for the crash arm. These are real regression guards, not decoration.

4. **No information leak in the new message.** `fuel_consumed` is a `u64` instruction counter derived from `cap.saturating_sub(store.get_fuel())`. No payload content, path, or identity crosses into the agent-visible string.

5. **The 5 s epoch budget was not silently invalidated.** Raising the fuel ceiling to 20M while leaving `timeout_ms: 5_000` alone raises the question of whether fuel exhaustion now converts into epoch timeouts — which would misdirect operators to exactly the wrong remedy that commit 2 exists to prevent. Arithmetic says no: ~20M WASM instructions is tens of milliseconds of compiled execution, orders of magnitude inside 5 s. Fuel remains the binding constraint for compute-bound payloads. No action needed; recording it because the two commits interact and the interaction is not obvious.

6. **The comment sweep is complete.** Grepping the tree for `10M fuel` / numeric cap phrasings surfaces no sites the diff left behind, and no `.md` doc states the cap numerically, so there is no doc-drift tail.

7. **No collateral test breakage expected.** Integration tests pin explicit caps (`1_000_000_000`, `50_000_000`, `u64::MAX`), so they are insensitive to the default. `defaults_applied_when_missing` was the one default-dependent assertion, and the diff updates it.

---

## MEDIUM-1 — The interpolated fuel figure is tautological, so the stated benefit is not delivered

`main.rs`, fuel arm:

```rust
"fail-closed: fuel exhausted ({fuel_consumed} units consumed; raise fuel_cap or reduce payload size)"
```

`fuel_consumed` comes from `fuel_consumed_from_store(&store, limits.fuel_cap)`, which is `cap.saturating_sub(remaining)`. On a fuel trap, remaining is at/near zero, so **`fuel_consumed ≈ fuel_cap` by construction, every time.** In production this message will read `20000000 units consumed` on every fuel exhaustion, forever. It restates the cap.

The test comment claims the opposite:

> `// Consumed-fuel figure surfaces for scale assessment.`
> `"...so operators can assess scale without log grep"`

There is no scale to assess. The number an operator needs is *demand* (how much fuel the payload wanted) or the *payload size* — neither of which fuel accounting can report, because metering stops at the ceiling.

Compounding this, the test constructs a state `invoke_plugin` cannot produce:

```rust
cause: TimeoutCause::Fuel,
fuel_consumed: 10_500_000,   // > any cap that could have trapped it
```

Consumption cannot exceed the cap that trapped it. So the assertion `text.contains("10500000")` pins a formatting behavior against an unreachable input while the stated user-facing benefit goes unverified.

Suggested fix (small): interpolate the **cap** rather than the consumed count — `"fuel cap of {fuel_cap} units exhausted; raise fuel_cap or reduce payload size"` — or drop the number entirely and keep the remedy text. Either way, retire the "assess scale" rationale from the comment, since no variant of this message can support it.

This is the pattern the repo's own self-audit checklist names: a comment asserting a capability the behavior does not have.

## MEDIUM-2 — The sync invariant is asserted, not enforced; the test adds a third literal

The diff leaves **two independent `20_000_000` literals** in two files and introduces a **third** in the test:

- `invoke.rs` — `InvokeLimits::default().fuel_cap`
- `registry.rs` — `RegistryDefaults::default().fuel_cap`
- `registry.rs` tests — `const ADR_042_FUEL_CAP: u64 = 20_000_000`

The test's own comment states the problem accurately — "a comment cannot enforce the invariant" — and then declines the fix that would. A single `pub const DEFAULT_FUEL_CAP: u64 = 20_000_000` referenced by both `Default` impls makes drift *structurally impossible* rather than *test-detectable*, and collapses three edit sites to one. The pinning-vs-equality reasoning in the comment is sound as far as it goes, but it is reasoning about how to *detect* a class of defect that a shared constant *eliminates*.

Note also that `invoke_limits_default_fuel_cap_is_20m` (invoke.rs) and the first assertion of `fuel_cap_defaults_stay_in_sync` (registry.rs) are the same assertion written twice. With a shared constant, both tests reduce to one meaningful check.

I would not block merge on this — the guard does fail on drift, which is a real improvement over the prose comment alone. But "fix the class, not the instance" applies, and this is a ~10-minute edit.

## MEDIUM-3 — `fuel_workload_exhausts_10m_completes_at_20m` proves less than its comment claims, and is version-drift sensitive on the lower bound

Two distinct issues.

**(a) Overclaim.** The comment says the test "demonstrates that the 10M→20M raise resolves the class of exhaustions seen in production" and is "calibrated against the perf-engineer model: `fuel = 29,452 + 27.514 × payload_bytes`". The test drives a synthetic WAT arithmetic loop with an **empty payload** (`b""`). It exercises no payload, no host calls, no serialization — none of the machinery the regression model describes. What it actually proves is narrower and still worthwhile: *the dispatcher enforces the fuel ceiling at the boundary, and a ~10.8M-fuel workload crosses from trapped to completing when the ceiling moves 10M→20M.* That is a legitimate thing to test. The production claim would require driving a ~377 KB payload through the real plugin path. Recommend rewording the comment to what the test does, or adding the payload-driven case if the production claim is meant to be load-bearing.

**(b) Drift sensitivity.** The lower bound has only ~8% margin: the loop is estimated at ~10.8M fuel against a 10M cap, and the estimate ("1,200,000 iterations × ~9 instructions") is never asserted. If a wasmtime upgrade changes per-instruction fuel accounting by more than ~8% downward, `res_10m` becomes `Ok` and the test fails — loudly, which is better than silently, but it fails as an opaque assertion in an unrelated-looking test after a routine dependency bump. The 20M side has ~46% margin and is fine.

Cheap hardening: assert the observed `fuel_consumed` from the 20M run (e.g. `> 10_000_000` and `< 20_000_000`). That converts a future accounting change from a mysterious `matches!` failure into a message that states the actual cost, and it pins the calibration the comment currently only asserts in prose.

## LOW-1 — Comment strengthens an unmeasured claim

`validate-index-cite-refresh/Cargo.toml`:

```
-# WASM 10M fuel budget. The regex crate's DFA compilation exhausts the fuel
-# cap on first call.
+# WASM fuel budget. The regex crate's DFA compilation exhausts even a 20M
+# fuel cap on first call.
```

The original recorded an observation against a cap that existed. The replacement asserts a stronger claim — exhaustion at 20M — with nothing in the diff measuring it. It is probably true, but "probably true and now written down as fact" is how comments rot. The sibling edits in this same diff get the framing right (`validate-state-structure`: "exhausting the **former** 10M fuel budget"). Recommend matching that: "exhausted the former 10M cap on first call; not re-measured at 20M."

## LOW-2 — Coverage gap on the accessor production actually calls

The new sync test pins the two `Default` impls. The production path is neither of them directly: `executor.rs` builds `InvokeLimits { fuel_cap: entry.fuel_cap(&registry.defaults), .. }`, i.e. it goes through `HookEntry::fuel_cap()` and never through `InvokeLimits::default()`. So `InvokeLimits::default().fuel_cap` is a caller-convenience fallback, while `RegistryDefaults` is the load-bearing one — the reverse of the emphasis the diff's comments imply.

`defaults_applied_when_missing` already asserts `reg.hooks[0].timeout_ms(&reg.defaults)` and `reg.hooks[0].priority(&reg.defaults)`. Adding the parallel `assert_eq!(reg.hooks[0].fuel_cap(&reg.defaults), 20_000_000);` is one line and closes the chain from parsed TOML to effective per-hook cap — the thing operators care about. Low severity because `unwrap_or` is trivially correct and I confirmed no TOML overrides exist, so the chain holds by inspection today.

## NIT-1 — Operator docs not updated for the new sentinel

`CLAUDE.md` § Factory Hook Diagnostics has a "WASM fuel exhaustion on lessons.md" row — the exact place an operator lands after seeing this block — and it does not mention that `block_reason` now names fuel explicitly. Not required (no doc states the cap numerically, so there is no drift), and I recognize doc placement may belong on a different branch. Flagging because the diff's entire justification is operator ergonomics, and this is where the ergonomics are documented.

## NIT-2 — Prose sentinel versus this repo's code convention

The new reason is free prose with an embedded number. Elsewhere this project uses stable machine-greppable codes (`MULTI_COMMIT_CHAIN_NOT_ALLOWED`). Any consumer that today matches `"fail-closed: plugin timed out"` to detect timeouts will silently stop matching fuel cases — which is the intent, but it is an unversioned change to a semi-public string. A stable prefix (`FUEL_EXHAUSTED: ...`) would let consumers match a fixed token instead of variable prose. Worth considering; not worth blocking.

---

## Summary

| Severity | Count | Items |
|---|---|---|
| BLOCKER | 0 | — |
| MEDIUM | 3 | Tautological fuel figure + unreachable test input; sync invariant asserted not enforced (3 literals); WAT test overclaims and has ~8% lower-bound margin |
| LOW | 2 | Unmeasured 20M regex claim; no assertion on `HookEntry::fuel_cap()` production accessor |
| NIT | 2 | Operator-doc row not updated; prose sentinel vs. stable code convention |

The substance of both functional commits is right: the cap raise is measurement-motivated and verifiably reaches every plugin, and the fuel/epoch split is exhaustive with the pre-existing arms provably untouched. Every finding above is about a *claim* — a comment, a test rationale, or a stated invariant — outrunning what the code does. That is a narrower and more fixable problem than a defect in the change itself.

Recommend landing MEDIUM-1 and MEDIUM-2 in-scope before merge (both are single-digit-minute edits, and MEDIUM-1 removes a comment that asserts a benefit the mechanism cannot provide). MEDIUM-3(a) is a comment reword; MEDIUM-3(b), LOW-1, and LOW-2 are each one or two lines. None of them require rethinking the approach.

One transparency note on method: the greps backing items 1, 2, 4, 6, and 7 above ran against the repository baseline rather than a checkout of `4bd626c5`, so they establish the pre-change state and the surrounding structure the diff relies on — they are not a claim about post-merge tree contents.

Zious11 (commented):
## Security Re-Review — Delta `4bd626c5..ce0ebef1`

Scoped to the two commits that landed after review `4897650995` (anchored at `4bd626c5`):
- `4bdd1533` — prior-finding fixes (message reword, `DEFAULT_FUEL_CAP` constant, test hardening, comment corrections)
- `ce0ebef1` — `fuel_cap: u64` added to `PluginResult::Timeout`; `classify_trap` threads it; fuel-arm message interpolates `fuel_cap`

---

### Disposition of Prior Finding 2 — RETRACTED

**Prior classification:** CWE-209 / CWE-497 — operator-visible `block_reason` interpolated `fuel_consumed`, which via the perf-engineer regression formula (`fuel ≈ 29,452 + 27.514 × payload_bytes`, R²=0.9999999) acted as a proxy for `.factory/` artifact payload size.

**Finding 2 is retracted. The reasoning was unsound when written.**

The invariant to verify: on `Trap::OutOfFuel`, `wasmtime`'s remaining-fuel counter reaches zero before the trap is raised. `fuel_consumed_from_store` computes `cap.saturating_sub(remaining)`. With `remaining == 0`, this is `cap.saturating_sub(0) = cap`. Therefore `fuel_consumed ≡ fuel_cap` on every fuel trap — a constant, not a metered demand figure.

Verified in source (`crates/factory-dispatcher/src/invoke.rs`):

```rust
fn fuel_consumed_from_store(store: &Store<StoreData>, cap: u64) -> u64 {
    match store.get_fuel() {
        Ok(remaining) => cap.saturating_sub(remaining),
        Err(_) => 0,
    }
}
```

The regression formula relates *wasmtime fuel consumed during normal execution* to payload bytes. On a fuel trap, execution terminates when the counter hits zero — the observed `fuel_consumed` value is the cap, regardless of how much work the module actually performed before hitting the ceiling. Inferring payload size from the cap constant would tell an observer only what they already know from reading the public ADR-042 §Decision 1 rationale.

The delta makes this explicit: `fuel_cap` is now a separate named field on `PluginResult::Timeout`, the doc comment on `fuel_consumed` explains the invariant, and the operator-facing message says "fuel cap of N units exhausted" — not "N units consumed." This is the correct remediation. No second security classification is warranted.

---

### Prior Finding 1 — CARRIES FORWARD (code not in delta)

**CWE-400 / CWE-770** (MEDIUM): An operator raising `DEFAULT_FUEL_CAP` without an explicit ceiling is bounded only by the independent 5 s epoch deadline via `EpochTicker`. The delta does not touch `EpochTicker` configuration, `InvokeLimits::timeout_ms`, or the epoch-arm string. This constraint still holds and the bound still exists. No change in disposition.

---

### Assessment of Delta

**1. Disclosure surface — `fuel_cap` in operator message**

`fuel_cap` is `DEFAULT_FUEL_CAP = 20_000_000` (compile-time constant, `pub const` in `invoke.rs`, documented in ADR-042). No per-hook `fuel_cap` overrides exist in `plugins/vsdd-factory/hooks-registry.toml` — verified by grep, output empty.

In the hypothetical registry-override case: a per-hook value would be operator-configured, not payload-derived. An observer could infer *which* hook triggered by correlating the emitted cap to the registry — but that is operator-visible configuration, not user data or artifact content. This is not a finding at any severity.

The new message is a *smaller* disclosure surface than the prior design. The prior claim was that `fuel_consumed` could proxy payload size via the regression formula; that claim was wrong, but the intent of the message was to emit a metered figure. The new message emits only the configured constant. Disposition: **no finding**.

**2. Registry-override path**

Confirmed: no production hook overrides `fuel_cap`. The `RegistryEntry.fuel_cap` accessor falls back to `RegistryDefaults::default().fuel_cap` which is `crate::invoke::DEFAULT_FUEL_CAP`. If a future entry introduced a per-hook cap, the emitted value would reflect the operator's deliberate choice, not session state. **No finding.**

**3. `fuel_cap` in `pub enum` variant and event log**

`fuel_cap: u64` was added to `PluginResult::Timeout`. The `emit_lifecycle` function in `executor.rs` explicitly silences it with `fuel_cap: _` — the internal JSONL log (`dispatcher-internal-YYYY-MM-DD.jsonl`) records only `fuel_consumed` under the existing `fuel_consumed` key. The VSDD_SINK_FILE output captures plugin-emitted events via `host::emit_event`, which is a separate path and does not include `fuel_cap`. The field crosses the module boundary in the `pub enum` but is unused in any telemetry output. **No finding.**

**4. Fail-closed integrity**

`classify_trap` gained `fuel_cap: u64` as an additional pass-through parameter. Both the `Trap::Interrupt` arm (epoch) and the `Trap::OutOfFuel` arm (fuel) now carry `fuel_cap` into the `PluginResult::Timeout` variant. Neither arm changes whether the result triggers blocking. Blocking decisions live in `plugin_fail_closed()` and `plugin_requests_block()`, which match on `PluginResult::Crashed { .. } | PluginResult::Timeout { .. }` against `on_error == OnError::Block` — pattern unchanged by the delta. The epoch arm still returns `Some("fail-closed: plugin timed out")` byte-for-byte. **No regression in fail-closed integrity.**

**5. `FUEL_EXHAUSTED:` prefix — consumer compatibility (LOW)**

Prior behavior: both epoch and fuel timeouts produced `"fail-closed: plugin timed out"`.
Current behavior: fuel → `"FUEL_EXHAUSTED: fuel cap of N units exhausted; ..."`, epoch → `"fail-closed: plugin timed out"` (unchanged).

This is a **breaking change for any consumer that matched `"fail-closed: plugin timed out"` to detect fuel exhaustion events**. In practice, the existing message was misleading (identical to the epoch case), so any such consumer was already unreliable. The hook still terminates and the block still fires on `on_error=Block` — the security-relevant behavior is unchanged. The JSONL internal log continues to record `cause: "fuel"` for disambiguation.

**Severity: LOW / informational.** Operators with monitoring rules that match `"fail-closed: plugin timed out"` to detect fuel exhaustion should add a rule matching `"FUEL_EXHAUSTED:"`. Not a blocker.

---

### Verdict

**APPROVED** — no new blockers.

| Finding | CWE | Severity | Disposition |
|---|---|---|---|
| Prior F2: `fuel_consumed` as payload-size proxy | CWE-209 / CWE-497 | — | **RETRACTED** — `fuel_consumed ≡ fuel_cap` on `Trap::OutOfFuel`; the proxy never existed |
| Prior F1: unbounded fuel cap raise | CWE-400 / CWE-770 | MEDIUM | **Carries forward** — epoch deadline still bounds it; code not in delta |
| New: `FUEL_EXHAUSTED:` prefix breaks consumers matching old sentinel | — | LOW | Informational; monitoring rules should be updated |

No CWE-209/CWE-497 exposure exists in this delta. The prior review's second finding was based on a misreading of `fuel_consumed_from_store` semantics under `Trap::OutOfFuel`. This review supersedes Finding 2 in review `4897650995`.

Zious11 (commented):
## Follow-up review — delta `4bd626c5..ce0ebef1`

**Scope:** this review covers **only** the two commits added after the two reviews pinned to `4bd626c5`:

- `4bdd1533` — addresses prior-review MEDIUM-1/2/3 and LOW-1/2
- `ce0ebef1` — plumbs `fuel_cap` into `PluginResult::Timeout`

`git diff 4bd626c5..ce0ebef1` — 6 files, 170 insertions / 56 deletions. The three commits reviewed at `4bd626c5` are **not** re-reviewed here. `crates/hook-plugins/validate-index-cite-refresh/src/lib.rs` and `crates/hook-plugins/validate-state-structure/src/lib.rs` appear in the full PR diff but not in this delta, and are therefore out of scope.

**Verdict: APPROVE-WITH-NOTES.** All five prior findings are genuinely closed — none cosmetically. The `fuel_cap` plumbing is the right fix and the sweep is complete. Two MEDIUM and two LOW findings are new in the delta; all four are single-edit fixes in test/comment surface, none affects production behaviour, and none blocks merge. One out-of-delta observation is recorded at the end because it contradicts this PR's own stated rationale.

Posted as `--comment` rather than `--approve`: the PR author and this reviewer are the same git identity, and GitHub rejects self-approval.

---

## Verification method

The checked-out worktree is at `9afc3226` on `fix/nested-factory-path-derivation`, which does **not** contain `ce0ebef1`. Every claim below that the diff cannot self-evidence was therefore checked with `git grep <pattern> ce0ebef1` — i.e. pinned to the PR head rev, not to the working tree. Greps run:

| Grep (all `git grep … ce0ebef1`) | Purpose |
|---|---|
| `PluginResult::Timeout` over `*.rs` | construction/match-site sweep (below) |
| `20_000_000\|20000000` and `10_000_000\|10000000` over `*.rs` `*.toml` | MEDIUM-2 literal-elimination check |
| `fuel_cap` over `*.toml` + `plugins/vsdd-factory/hooks-registry.toml` | confirm no registry entry overrides the cap |
| `set_fuel\|get_fuel\|fuel_consumed_from_store\|fuel_async` over `invoke.rs` | confirm the plumbed value is the value set on the Store |
| `consume_fuel\|epoch_interruption` over `src/*.rs` | reachability of the `Err(_)` arm (LOW-2 below) |
| `FUEL_EXHAUSTED`, `fuel exhausted`, `fail-closed` (incl. non-`.rs`) | consumers of the changed reason string |
| `MULTI_COMMIT_CHAIN_NOT_ALLOWED` over `crates/` | validate the cited precedent |
| `non_exhaustive` over `src/`; `factory-dispatcher` over `*/Cargo.toml` | API-surface question |
| `git show ce0ebef1:…` for `invoke.rs`, `main.rs`, `registry.rs`, `executor.rs`, `resolver_classify_trap.rs`, `hooks-registry.toml`, `Cargo.toml` | read at rev |

No local build: CI at `ce0ebef1` is green on `cargo-host (ubuntu-latest)`, `cargo-host (macos-latest)`, `bats-full-suite (linux)`, `bats-darwin-leg`, and SAST. Compilation and test outcomes are taken from that, not asserted from reading.

---

## Prior findings — closure determination

### MEDIUM-1 (tautological fuel figure + unreachable test input) — **CLOSED**

Closed on the substance, not by rewording. The operator message now interpolates `fuel_cap`, and the value is traceable end-to-end:

- `invoke.rs:375` — `store.set_fuel(limits.fuel_cap)`, `?`-propagated to `InvokeError::Setup`
- `invoke.rs:423` — `classify_trap(…, limits.fuel_cap)`
- `invoke.rs:459-465` — `Trap::OutOfFuel => PluginResult::Timeout { …, fuel_cap }`
- `main.rs:854` — `"FUEL_EXHAUSTED: fuel cap of {fuel_cap} units exhausted; …"`

So the interpolated figure is *the same value handed to `set_fuel`*, which makes the new doc claim on the field ("Always reflects the actual cap set on the wasmtime Store, independent of whether `get_fuel()` succeeds") accurate rather than aspirational. `grep set_fuel\|fuel_async` over `invoke.rs` confirms there is no mid-run fuel top-up and no `fuel_async_yield_interval`, so nothing can make the recorded cap diverge from the enforced cap.

The unreachable `fuel_consumed: 10_500_000` test input is gone; the input is now `DEFAULT_FUEL_CAP`, which is the reachable state. See NEW MEDIUM-2 for a caveat on the *replacement* justification.

### MEDIUM-2 (three `20_000_000` literals) — **CLOSED, structurally**

`git grep '20_000_000\|20000000' ce0ebef1 -- '*.rs' '*.toml'` returns exactly two hits: `invoke.rs:279` (the `pub const DEFAULT_FUEL_CAP` definition) and `main.rs:1179` (prose in a comment). Both `Default` impls now reference the constant (`invoke.rs:294`, `registry.rs:190`). The literal class is eliminated, and drift is now a compile-time property rather than a test-detected one — which is what the finding asked for.

Retaining `fuel_cap_defaults_stay_in_sync` as a cross-module guard is the right call, and the rewritten rationale comment describes what the test now does (catching a refactor that bypasses the constant) rather than what it used to do. Failure messages name which `Default` impl drifted. Good.

### MEDIUM-3 (WAT test overclaimed production calibration; 10M bound unasserted) — **CLOSED**

The overclaim is genuinely retracted, not softened: `invoke.rs:1095-1102` now states outright that this is "a ceiling-enforcement test, not a production model", that the loop runs "with an empty payload and no host calls", and that it "does not reflect the regression formula measured by the perf-engineer". The formula is retained only as the thing the test is *not*. That is the honest version.

The fuel estimate is now asserted rather than implied — `invoke.rs:1161-1178` destructures the `Ok` arm and asserts `fuel_consumed > 10_000_000` and `< DEFAULT_FUEL_CAP`, with the observed value interpolated into both failure messages.

Residual, not a finding: the ~8% margin over the 10M lower bound is inherent to the workload and unchanged (iteration count is still 1,200,000). What changed is that a wasmtime accounting shift now fails with "got 9,84x,xxx" instead of an opaque `matches!` mismatch. That satisfies the diagnosability half of the finding; the margin half is a property of the chosen workload and I don't think it's worth widening.

### LOW-1 (unmeasured "exhausts even a 20M cap" claim) — **CLOSED**

`validate-index-cite-refresh/Cargo.toml` now reads "exhausted the former 10M fuel cap on first call; not re-measured at the current 20M cap." The unmeasured claim is replaced by the measured one plus an explicit statement of what is *not* known. Correct handling — it neither overclaims nor silently drops the rationale for excluding `regex`.

### LOW-2 (no assertion on the accessor production uses) — **CLOSED, and load-bearing**

`registry.rs:571-581` adds `assert_eq!(reg.hooks[0].fuel_cap(&reg.defaults), DEFAULT_FUEL_CAP)`. I checked this is not testing an override: `minimal_toml()` declares only `name`/`event`/`tool`/`plugin` for `hooks[0]`, so `fuel_cap` is `None` and the accessor at `registry.rs:312-314` exercises the `unwrap_or(defaults.fuel_cap)` fallback — the production path. And `git grep fuel_cap ce0ebef1 -- '*.toml'` returns no hits in `plugins/vsdd-factory/hooks-registry.toml`, so **every** registered plugin resolves through exactly this fallback. The assertion covers the real path.

**No prior finding was closed cosmetically.** In each case the load-bearing artifact (the interpolated field, the constant, the assertion) changed, not just the prose around it.

---

## New findings in the delta

### MEDIUM-1 (new) — two of three assertions in the non-vacuity test are vacuous

`main.rs:1226-1257`, `fuel_timeout_with_zero_consumed_still_reports_cap`. The test's stated purpose is to prove `ce0ebef1` is load-bearing. Taking the pre-`ce0ebef1` format string (`main.rs:854` at `4bdd1533`: `"FUEL_EXHAUSTED: fuel cap of {fuel_consumed} units exhausted; raise fuel_cap or reduce payload size"`) and the same inputs, the message would be `FUEL_EXHAUSTED: fuel cap of 0 units exhausted; raise fuel_cap or reduce payload size`. Each assertion against that string:

| Assertion | vs. pre-`ce0ebef1` message | Load-bearing? |
|---|---|---|
| `text.starts_with("FUEL_EXHAUSTED:")` | passes | **No** — the prefix was introduced in `4bdd1533`, so this cannot discriminate `ce0ebef1` |
| `text.contains(DEFAULT_FUEL_CAP.to_string())` → `"20000000"` | **fails** | **Yes** — this one assertion carries the whole test |
| `!text.contains(": 0 ")` | passes | **No — vacuous** |

The third is the one worth naming, because it is the assertion whose *stated intent* is to catch the exact bug ("message must NOT interpolate fuel_consumed=0 as the cap") and it cannot catch it. The buggy message renders `of 0 units`, not `: 0 `; the only colon in the string is the one in `FUEL_EXHAUSTED:`, followed by a space and `f`. So the guard passes against the very defect it names. I verified this by evaluating both message strings against all three predicates rather than reasoning about it only.

Fix is one character-class: `!text.contains("of 0 units")` (or `!text.contains(" 0 units")`). That predicate fails against the pre-fix message and passes after, making all three assertions discriminating.

The test does earn its place — assertion 2 is genuine and would fail against the pre-`ce0ebef1` code — so this is not "the test is vacuous". It is "the assertion added specifically to be the tripwire is not wired to anything."

### MEDIUM-2 (new) — the test's justifying comment asserts an unreachable production state

`main.rs:1219-1225` justifies the test as covering "the `Err(_)` path of `fuel_consumed_from_store` (when `store.get_fuel()` returns an error)" and states that before the fix "the operator-facing message read 'fuel cap of 0 units exhausted' — a false statement."

That state cannot occur. `fuel_consumed == 0` on a fuel trap requires the `Err(_) => 0` arm at `invoke.rs:487`. `Store::get_fuel()` errors only when fuel consumption is not enabled on the `Engine` config — and `engine.rs:36` sets `config.consume_fuel(true)` unconditionally, while `invoke.rs:375-376` `?`-propagates a failed `set_fuel` into `InvokeError::Setup` before execution begins. Any store that reaches a `Trap::OutOfFuel` has therefore already proven fuel is enabled, so `get_fuel()` cannot fail there and the `Err(_)` arm is dead on this path.

This matters because it is the same defect class MEDIUM-1 flagged in the prior review — an unreachable input presented as a production scenario. The fix removed one instance (`fuel_consumed: 10_500_000`) and introduced another as the *justification* for the replacement test. I'm rating this MEDIUM rather than LOW because it is a false factual claim about production behaviour sitting in a comment whose whole job is to explain why a test is load-bearing; a future reader will believe it.

The remedy is a comment rewrite, not a test deletion. The test is legitimately valuable as a **decoupling** test: it proves the message is a function of `fuel_cap` alone and does not read `fuel_consumed`, which is a property worth pinning regardless of whether `fuel_consumed: 0` is reachable. Say that, and drop the `Err(_)`-path and "false statement" framing. Note this does **not** undermine MEDIUM-1's closure: the plumbing is correct on MEDIUM-1's actual grounds (the message previously presented a cap as a measured consumption figure), which is a real defect on the reachable path.

### LOW-1 (new, unrequested) — the fuel arm drops the `fail-closed:` family prefix

No prior finding asked for a prefix change; `4bdd1533` introduced one. The three arms of `extract_reason_from_outcome` now read:

- `main.rs:848` — `"fail-closed: plugin crashed"`
- `main.rs:854` — `"FUEL_EXHAUSTED: fuel cap of {fuel_cap} units exhausted; …"`
- `main.rs:860` — `"fail-closed: plugin timed out"`

The function's own doc comment (`main.rs:769-771`) describes it as producing reasons "for fail-closed crash/timeout outcomes", so the fuel arm is now the one fail-closed reason that doesn't say `fail-closed:`. Anything that keys on that prefix to identify a fail-closed block silently stops matching fuel exhaustion.

Two specifics about the justifying comment at `main.rs:836-841`:

1. The cited precedent doesn't support the change. `git grep MULTI_COMMIT_CHAIN_NOT_ALLOWED ce0ebef1 -- crates/` shows it is emitted by the **`validate-burst-log` hook plugin** (`crates/hook-plugins/validate-burst-log/src/lib.rs:548`) as a violation message — a different producer at a different layer — and per CLAUDE.md's TD #71 example it surfaces *after* its own prefix (`block_reason="FAIL: MULTI_COMMIT_CHAIN_NOT_ALLOWED — …"`). It is a token within a prefixed message, not a replacement for the prefix. It establishes no convention for the dispatcher's own reason vocabulary.
2. "so consumers can match a fixed token" has no consumer. `git grep FUEL_EXHAUSTED ce0ebef1` returns hits only in `main.rs` and its own tests — nothing in `plugins/`, `docs/`, or the bats suite.

Blast radius is nil, and I want to be explicit about that: `git show origin/develop:…/main.rs` has only `PluginResult::Timeout { .. } => Some("fail-closed: plugin timed out")` — the entire fuel arm is new in this PR, so no shipped runbook can be broken. This is a convention call, not a regression. But the comment's asymmetry is worth noting: it protects the epoch string's wording for runbook stability while dropping the shared family prefix in the arm immediately above it. `"fail-closed: FUEL_EXHAUSTED: fuel cap of {fuel_cap} units exhausted; …"` gets both properties.

### LOW-2 (new) — `fuel_cap` is discarded in the structured telemetry path

`executor.rs:683-709`, `emit_lifecycle`. The `PluginResult::Timeout` arm destructures `fuel_cap: _` and emits `PLUGIN_TIMEOUT` with `*fuel_consumed` as the event's fuel value plus `cause` and `stderr` extras. So the `plugin.timeout` record in `.factory/logs/dispatcher-internal-*.jsonl` still carries only `fuel_consumed`, which on every fuel trap is tautologically equal to the cap. An operator who greps by trace UUID (the CLAUDE.md §"Factory Hook Diagnostics" Step 2 fallback path) sees `20000000` and still cannot tell "the cap was 20M" from "demand was 20M" — the precise ambiguity MEDIUM-1 identified. The stderr summary is fixed; the log record is not.

To be fair to the change: `fuel_cap: _` is the *right* pattern here — explicit and exhaustive, so a future field addition breaks this site loudly rather than being absorbed. The field is deliberately dropped, not accidentally. Adding `("fuel_cap".to_string(), Value::from(*fuel_cap))` to the `Timeout` extras vec would close the telemetry half; I'd take it in this PR or a follow-up, and it's not a merge blocker either way.

---

## The public-variant field addition

**Was every site swept? Yes, and the construction half is compiler-enforced.** You cannot construct a struct-variant with a missing field, so all 11 construction sites had to be updated to compile — `invoke.rs:452`/`459` in production, plus test constructions at `executor.rs:840`/`892`/`909`, `invoke.rs:1138`/`1230`/`1268`, `main.rs:1187`/`1234`/`1265`. Green `cargo-host` on both platforms confirms none were missed. TD-VSDD-060 is structurally satisfied on that side.

**Was anything silenced with `..`?** No — not by this delta. The `..` match sites are `executor.rs:644`, `main.rs:619`, `main.rs:790`, and six in the integration tests (`tests/executor_integration.rs:214`, `tests/full_stack_plugin_invocation.rs:383`/`489`/`581`/`1501`/`1612`). All are pre-existing — the delta touches no file under `tests/` — and each is an "is this a timeout at all" discriminator that has no business reading the cap, so `..` is correct there. The only pattern newly written in this delta is `executor.rs:688` `fuel_cap: _`, which is exhaustive-explicit. Nothing was silenced.

**Is adding a field to a public variant the right call?** Yes, for this crate. `PluginResult` has no external consumers: `crates/factory-dispatcher/Cargo.toml` sets `publish = false`, and `git grep factory-dispatcher ce0ebef1 -- '*/Cargo.toml'` shows no workspace crate depends on it (the `hook-sdk` hit is a comment stating the opposite — that `ResolverError` is host-side only). The consumers are the crate's own `src/` plus its integration tests. So the "breaking change to a `pub enum`" is bounded to in-repo, all of it swept.

On the alternatives: recomputing the cap at the message site is not available (`extract_reason_from_outcome` receives only `&PluginResult`); a side-channel wrapper struct would be more churn for a value that is intrinsic to the outcome; replacing `fuel_consumed` with something like `FuelOutcome { cap, consumed }` would type the relationship better but is a strictly larger break for the same information. The chosen fix is the minimal correct one. Carrying `fuel_cap` on the `Epoch` arm too, where nothing reads it, is fine — the value is truthful there.

**Does it warrant a note?** Two, neither blocking:

1. **`#[non_exhaustive]` on `PluginResult`.** This crate already has the precedent with the exact rationale — `resolver.rs:70-78`: "`#[non_exhaustive]` allows adding fields to existing variants and new variants in S-12.04+ without a breaking ABI change (F-P2-006)", and again at `resolver_loader.rs:93-95`. `PluginResult` carries no such attribute, so every future field addition repeats this sweep. Applying the crate's own convention here would make the next one a non-event. Follow-up, not this PR.
2. **The operator-visible change is the one that deserves documentation, not the API change.** The PR touches no `CHANGELOG.md` and no docs. Given `publish = false`, I don't think the variant shape needs a CHANGELOG entry. But `FUEL_EXHAUSTED:` is a **new operator-facing `block_reason` token**, and CLAUDE.md §"Factory Hook Diagnostics" documents `block_reason` and maintains a "Common blockers and recovery procedures" table that already has a WASM-fuel-exhaustion row keyed to the old behaviour. That table is where an operator will look. Worth a row or an amendment naming the new token and the "raise `fuel_cap`" remedy.

## Is `fuel_consumed` now dead weight?

No — and its doc comment is accurate. It is genuine telemetry on the `Epoch` arm, where the interrupt lands mid-execution and the value is a real partial-consumption figure, and it feeds `emit_lifecycle`'s event fuel field for both arms (`executor.rs:697`). It is only degenerate on the `Fuel` arm, and the doc comment at `invoke.rs:227-232` states that degeneracy explicitly, names the mechanism (`cap.saturating_sub(remaining)` with `remaining == 0`), and redirects readers to `fuel_cap` for operator messages. I checked the described mechanism against `fuel_consumed_from_store` at `invoke.rs:484-489` — it matches. The asymmetry with `PluginResult::Crashed` (which keeps `fuel_consumed` and gains no `fuel_cap`) is also justified: on a non-fuel trap `fuel_consumed` is a real metered figure, so there is nothing to disambiguate.

Minor wording nit: the field comment opens with "Instructions executed before the interrupt", which is true for `Epoch` and misleading for `Fuel`; the following three sentences correct it, so the comment is net accurate and I'd leave it.

## `classify_trap` signature change

Sound. One caller — `invoke.rs:417-424` — passing `limits.fuel_cap`, the same value given to `set_fuel` 42 lines earlier in the same function, so there is no path where the recorded cap and the enforced cap can diverge. The private `fn` needed no other sweep (`git grep classify_trap` shows the resolver's `classify_resolver_trap` is a separate function in a separate module, not a caller).

---

## Out-of-delta observation (no severity; recommend a follow-up issue)

Not a finding against this delta — `resolver_classify_trap.rs` is untouched by it — but it directly contradicts this PR's stated rationale, so it should not go unrecorded.

This PR exists because, per `main.rs:830-835`, conflating fuel exhaustion with wall-clock timeout "forces operators to grep the internal log for the trace UUID to determine cause — TD #71 pattern," and the two need distinct remedies (reduce payload vs. investigate slow script). The dispatcher has a **second** WASM invocation path where they are deliberately merged: `classify_resolver_trap` maps both `Trap::Interrupt` and `Trap::OutOfFuel` to `ResolverError::Timeout { name }`, whose `Display` is `"resolver '{name}' timed out"`. The comment at the `OutOfFuel` arm argues the opposite of this PR — "Both epoch and fuel limits represent 'took too long', so both map to Timeout" — and describes itself as "the definitive classification path". A resolver that exhausts its fuel therefore produces exactly the undiagnosable message this PR removes for hook plugins, and there is no cap in the error to tell the operator what to raise.

Worth a follow-up to decide deliberately whether the resolver path adopts the same fuel/epoch distinction (and whether `ResolverError` — which *is* `#[non_exhaustive]`, so this is cheap — should carry the cap), or whether the merge is intentional and the two comments should stop contradicting each other. I'm not asking for it in this PR.

---

## Summary

| | |
|---|---|
| Prior MEDIUM-1 | Closed |
| Prior MEDIUM-2 | Closed (structurally) |
| Prior MEDIUM-3 | Closed |
| Prior LOW-1 | Closed |
| Prior LOW-2 | Closed, load-bearing verified |
| New MEDIUM-1 | Vacuous tripwire assertion in `fuel_timeout_with_zero_consumed_still_reports_cap` — one-predicate fix |
| New MEDIUM-2 | Test comment asserts an unreachable production state — comment rewrite |
| New LOW-1 | Fuel arm drops the `fail-closed:` family prefix; cited precedent is cross-layer and has no consumer |
| New LOW-2 | `fuel_cap` discarded in `emit_lifecycle`, so `plugin.timeout` telemetry keeps the tautology |
| Sweep | Complete; construction sites compiler-enforced; no `..` introduced by this delta |
| Public variant change | Right call, bounded to in-repo consumers; recommend `#[non_exhaustive]` follow-up per the crate's own `resolver.rs` precedent, and a CLAUDE.md blockers-table note for the new `FUEL_EXHAUSTED:` token |

None of the four new findings blocks merge. All are confined to test assertions, comments, and one telemetry field.

Zious11 (commented):
# Final scoped review — net delta `ce0ebef1..783b88e6`

**Verdict: APPROVE — no BLOCKER. Two NITs, one observation. Nothing that should hold the merge.**

## Scope and method

Reviewed strictly the net delta `git diff ce0ebef1..783b88e6` — 2 files, **98 insertions / 23 deletions**:

- `crates/factory-dispatcher/src/executor.rs`
- `crates/factory-dispatcher/src/main.rs`

The five commits already covered at `ce0ebef1` are out of scope and were not re-reviewed. `06d43140` (four fixes + `#[non_exhaustive]`) and `783b88e6` (revert of `#[non_exhaustive]`) are the two commits in range; the `#[non_exhaustive]` change cancels out in the net delta (`git diff ce0ebef1..783b88e6 | grep -c non_exhaustive` → `0`), so the substantive delta is the four fixes.

The working tree is on another branch, so all corroboration was done against blobs and greps **pinned at `783b88e6`**:

| Check | Command (all at rev `783b88e6`) |
|---|---|
| `extract_reason_from_outcome` full body | `git show 783b88e6:crates/factory-dispatcher/src/main.rs` |
| `extract_block_info` full body | same |
| `PluginResult` / `TimeoutCause` definitions | `git show 783b88e6:crates/factory-dispatcher/src/invoke.rs` |
| `#[non_exhaustive]` sweep | `git grep -n non_exhaustive 783b88e6 -- crates/` |
| `FUEL_EXHAUSTED` consumer sweep | `git grep -n FUEL_EXHAUSTED 783b88e6` |
| `fail-closed:` literal sweep in bats + integration tests | `git grep -n "fail-closed:" 783b88e6 -- plugins/vsdd-factory/tests/ crates/factory-dispatcher/tests/ docs/` |
| `emit_lifecycle` Timeout arm + extras assembly | `git show 783b88e6:crates/factory-dispatcher/src/executor.rs` |
| `InternalLog::write_inner` / mount gate | `git show 783b88e6:crates/factory-dispatcher/src/internal_log.rs`, `.../log_dir.rs` |
| dead-`Err(_)`-arm claim | `git grep -n "consume_fuel\|set_fuel" 783b88e6 -- crates/factory-dispatcher/src/` + `engine.rs`, `invoke.rs` blobs |
| `RegistryEntry` field list vs. new test literal | `git show 783b88e6:crates/factory-dispatcher/src/registry.rs` |
| `fuel_cap` provenance | `git grep -n fuel_cap 783b88e6 -- .../registry.rs .../executor.rs .../invoke.rs` |

Compilation and test outcomes are taken from CI, not asserted here — see **CI status** at the end.

---

## Fix 1 — vacuous tripwire assertion → **LANDED**

`fuel_timeout_with_zero_consumed_still_reports_cap` (`main.rs`) replaced `!text.contains(": 0 ")` with `!text.contains("of 0 units")`.

Predicate evaluated by hand against both message forms. `DEFAULT_FUEL_CAP = 20_000_000` (`invoke.rs:279`), so `.to_string()` is `"20000000"`.

| Message form | `contains("of 0 units")` | `!contains(...)` assertion |
|---|---|---|
| **post-fix** `fail-closed: FUEL_EXHAUSTED: fuel cap of 20000000 units exhausted; raise fuel_cap or reduce payload size` | no — the only `of ` is followed by `2` | **passes** |
| **pre-fix (bug)** `fail-closed: FUEL_EXHAUSTED: fuel cap of 0 units exhausted; …` | yes | **fails** |

The predicate now discriminates. The old one did not: in the buggy form the only colons are in `fail-closed:` (followed by `` `space` F ``) and `FUEL_EXHAUSTED:` (followed by `` `space` f ``), so `": 0 "` never occurred and the assertion passed against the buggy message. Confirmed vacuous both with and without the `fail-closed: ` prefix.

The other two assertions in the same test:

- `text.starts_with("fail-closed: FUEL_EXHAUSTED:")` — a prefix regression guard. It is **not** sensitive to the cap-vs-consumed bug (passes under both forms). Correctly scoped as a prefix guard, not the tripwire.
- `text.contains(&DEFAULT_FUEL_CAP.to_string())` — discriminating; the buggy `of 0 units` form lacks `20000000`. This was already load-bearing before this round; the repaired third assertion is now a redundant but independent second signal, which is fine.

NIT-1: `!contains("of 0 units")` is technically coupled to `DEFAULT_FUEL_CAP != 0`. Since the test hardcodes `fuel_cap: DEFAULT_FUEL_CAP` and a zero default fuel budget is nonsensical, this is not a real fragility. Not worth changing.

## Fix 2 — false comment → **LANDED**

The comment no longer claims the test covers the `Err(_)` path. It reframes the test as a decoupling test ("the message is a function of `fuel_cap` alone and must not read `fuel_consumed`"), which is directly verifiable from the `format!` — it interpolates `{fuel_cap}` and nothing else.

The added `Note:` makes a **negative** reachability claim, which I verified rather than took on faith:

- `build_engine` sets `config.consume_fuel(true)` unconditionally — `engine.rs:36`. ✓
- `store.set_fuel(limits.fuel_cap).map_err(|e| InvokeError::Setup(e.to_string()))?` aborts before instantiation — `invoke.rs:375-376`. ✓
- `fuel_consumed_from_store`'s `Err(_) => 0` arm is therefore unreachable on the production path. ✓

The comment does **not** assert reachability, and the `fuel_consumed: 0` input is now correctly labelled a synthetic boundary value. Claim verified, correctly hedged.

NIT-2 (informational): the `Note` is scoped to `build_engine`; `invoke_plugin` takes `engine: &Engine` from its caller, so the dead-arm property holds because every in-tree caller uses `build_engine`'s configuration (`resolver_loader.rs:591` documents matching config). The comment doesn't overclaim — it just doesn't state that qualifier. No change needed.

## Fix 3 — `fail-closed:` family prefix restored → **LANDED**

Fuel arm at `main.rs:854` is now:

```
"fail-closed: FUEL_EXHAUSTED: fuel cap of {fuel_cap} units exhausted; raise fuel_cap or reduce payload size"
```

- **Epoch arm byte-identical**: `Some("fail-closed: plugin timed out".to_owned())` — the diff contains no hunk touching that line, and `epoch_timeout_reason_is_unaffected` is unchanged, still an `assert_eq!` on the exact literal. Confirmed. `crash_reason_is_unaffected` likewise untouched.
- **All fail-closed arms now share the family prefix**: `Crashed` → `fail-closed: plugin crashed`; `Timeout/Fuel` → `fail-closed: FUEL_EXHAUSTED: …`; `Timeout/Epoch` → `fail-closed: plugin timed out`. The comment's "all three arms" claim is accurate for the fail-closed family.
- **Bad claims removed, not reworded**: the `MULTI_COMMIT_CHAIN_NOT_ALLOWED` precedent citation is gone from the source (`git grep MULTI_COMMIT_CHAIN 783b88e6 -- crates/` → no hit in this file), and "so consumers can match a fixed token" became "so **a future consumer can** distinguish fuel from epoch" — a capability statement, not an existence claim. `git grep -n FUEL_EXHAUSTED 783b88e6` returns hits **only** in `main.rs` (the format string, the comment, and two test assertions), confirming there is in fact no existing consumer, and that the string change breaks nothing.
- **No test outside the two updated assertions matches these strings**: `git grep -n "fail-closed:" 783b88e6 -- plugins/vsdd-factory/tests/ crates/factory-dispatcher/tests/ docs/` returns nothing. The TD-71 bats suite asserts only on the presence and single-line-ness of `block_reason=`, never its content, and `extract_block_info` applies no length cap to the reason, so the extra 13 characters are inert there.

NIT-3 (wording only): "All three arms" sits in a comment block above a four-arm `match` (the `Ok` advisory arm is the fourth, and correctly carries no prefix since it forwards the plugin's own reason). A reader counting arms could stumble. Cosmetic.

## Fix 4 — `fuel_cap` in telemetry → **LANDED, and the test is genuinely load-bearing**

`emit_lifecycle` binds `fuel_cap` (was `_`) and pushes `("fuel_cap", Value::from(*fuel_cap))` into the `Timeout` extras (`executor.rs:707`).

**The new test does prove the field reaches the log, not merely that it was passed to a constructor.** It is a real filesystem round-trip:

1. `emit_lifecycle(&log, …)` → `InternalEvent::now(…).with_field("fuel_cap", …)` → `log.write(&ev)`.
2. `InternalLog::write_inner` does `serde_json::to_string(event)` then `OpenOptions::new().append(true).create(true).open(path)` + `write_all` — a direct, unbuffered write with no deferred flush, so the bytes are on disk when `write` returns.
3. The test then `read_dir`s the log directory, asserts exactly one file, `read_to_string`s it, and `serde_json::from_str`s the line, asserting `event["type"] == plugin.timeout`, `event["fuel_consumed"] == 12_345_678`, and `event["fuel_cap"] == DEFAULT_FUEL_CAP`.
4. Top-level key access is correct because `InternalEvent.fields` is `#[serde(flatten)]`.

I specifically checked the failure mode that would have made this test silently green-but-hollow or spuriously red: `InternalLog::new` sets `mount_gated: true`, and `write_inner` **skips the write entirely** when the gate is closed. It is not closed here — `factory_mount_ready` returns `true` immediately unless the log dir's parent basename is `.factory`, and the parent here is a `tempfile::tempdir()`. So the write proceeds, and a regression that dropped the field would fail assertion 3 rather than being masked. Good test.

Correctness of the field itself: `Value::from(u64)` serializes as a JSON integer; `fuel_cap` in the `Timeout` variant is documented as "always reflects the actual cap set on the wasmtime Store, independent of whether `get_fuel()` succeeds", and at runtime resolves to `entry.fuel_cap(&registry.defaults)` (`executor.rs:134`, `:298`) — the same value passed to `set_fuel`. So the logged value is the cap that was actually in force.

---

## Is the revert inert? — **Yes.**

- `git grep -n non_exhaustive 783b88e6 -- crates/` shows no hit on `PluginResult` (`invoke.rs:212`). Remaining hits are pre-existing and unrelated: `resolver.rs:78`, `resolver_loader.rs:95`, a comment in `resolver_classify_trap.rs`, a deliberate "do NOT add" note in `hook-sdk/src/host.rs`, and `finish_non_exhaustive()` in `sink-core`.
- **No wildcard arm survives in either function.** `extract_reason_from_outcome` matches exactly `Ok { .. }`, `Crashed { .. }`, `Timeout { cause: Fuel, .. }`, `Timeout { cause: Epoch, .. }` — no `_` arm, and exhaustiveness over `TimeoutCause` (`Epoch | Fuel`) is compiler-enforced. `extract_block_info`'s `is_blocking` match is `Ok { stdout, exit_code, .. }` and `Crashed { .. } | Timeout { .. }` — also no `_` arm.
- The compiler is therefore the exhaustiveness assertion in both places, which is the whole point of the revert. As a bonus, `emit_lifecycle`'s `Timeout` arm destructures all five fields with no `..`, so a future field addition to the variant is a compile error at the telemetry site too — exactly where you'd want to be forced to decide whether to log it.

## Security of writing `fuel_cap` to `.factory/logs/dispatcher-internal-*.jsonl` — **no concern; I agree with the conclusion, with one correction to the premise**

`fuel_cap` is **not** a compile-time constant at the point it is logged. It resolves as `entry.fuel_cap.unwrap_or(defaults.fuel_cap)` (`registry.rs:312-313`), i.e. a per-entry value read from `hooks-registry.toml`, falling back to the `DEFAULT_FUEL_CAP` constant. The conclusion is unchanged: it is a checked-in resource budget, world-readable in the plugin cache tarball, carrying no secret, no path, no user or payload data. The same `plugin.timeout` record already carries `stderr` (truncated plugin output) and `fuel_consumed`, both strictly higher-sensitivity than an integer instruction budget. No new disclosure surface, no new sink, no new file. Not a finding.

## Anything the four fixes introduced that wasn't there before

Three things, all sub-MEDIUM:

- **OBS-1 (observation, no action):** on a fuel trap, `fuel_consumed == fuel_cap` by construction (`cap.saturating_sub(0)`), so for `cause=fuel` the JSONL record now carries two fields with identical values. This is harmless and arguably desirable — the record becomes self-describing — and for `cause=epoch` the two fields differ and `fuel_cap` adds genuine information (how much headroom was left when wall-clock fired). Also worth noting `INTERNAL_EVENT_SCHEMA_VERSION` (currently `1`) was not bumped; correct in my view, since the change is purely additive and per-event-type extras already vary (`Ok` carries `exit_code`, `Crashed` carries `trap`), so nothing enumerated is invalidated. `HOST_ABI.md` documents a *mandatory-field* list for the **async-path** `plugin.timeout` produced by `emit_plugin_timeout_async` — a different producer — and does not enumerate the sync-path payload, so there is no doc drift to fix.
- **NIT-4:** the new test constructs `RegistryEntry` as a 13-field literal with no `..Default::default()`. I verified the literal matches the struct's field set at rev exactly. Adding a field to `RegistryEntry` will break this test's compilation — consistent with the existing test style in this crate, and a small maintenance cost paid for compiler-enforced completeness. Fine as-is.
- **NIT-5:** the test asserts `files.len() == 1` on the log directory. Correct for a single event and a single date stamp; only fragile if `InternalLog` ever writes a sidecar. Acceptable.

None of NIT-1..5 or OBS-1 are worth another fix round. Per the termination rule they should be dropped, not tracked — with the possible exception of NIT-3's one-word wording, which can ride along in any future touch of that comment.

---

## CI status

Taken from `gh pr checks 774` at review time — **partial**:

- **pass:** `SAST (Semgrep)`, `validate`, `platforms-drift`, `bats-darwin-leg (macos)`, `bats-wave-handoff (macos)`, `build-dispatcher (linux-arm64)`.
- **still queued:** `cargo-host (ubuntu-latest)`, `cargo-host (macos-latest)`, `bats-full-suite (linux)`, and four remaining `build-dispatcher` matrix legs. These had not started after ~8 minutes of polling — runner queue, not failure.

`build-dispatcher (linux-arm64)` passing confirms the lib and binary compile in release, but it does **not** build test targets, so the new test module's compilation and the two updated assertions' outcomes are not yet CI-confirmed. That is a merge precondition (green CI), not a review finding — I am not asserting compilation or test results here, and nothing in this review depends on them: every claim above is derived from reading the code at `783b88e6` and evaluating the predicates by hand.

**Verdict: APPROVE. All four fixes landed as intended, not merely "addressed". The revert is inert and restores compiler-enforced exhaustiveness. No BLOCKER. Merge once CI is green.**

PR #777, opened 2026-08-15
fix(policy15-gate): implement POLICY 15 attestation-location gate as Rust crate

Zious11:
# fix(policy15-gate): implement POLICY 15 ATTESTATION-LOCATION GATE as a Rust crate

**Mode:** maintenance / CI-tooling (fix-PR delivery track — no BCs, no stubs/Red Gate, no wave-integration gate, no per-AC demo; this crate implements a governance gate, it is not a story deliverable)
**Type:** Rust workspace crate addition + one incidental doc fix
**Target:** `develop`

This PR delivers `crates/policy15-attestation-gate/` — the production, POLICY
21-compliant Rust implementation of the **POLICY 15 ATTESTATION-LOCATION
GATE** (ADR-040 §Decisions 7–10; `policies.yaml` v1.4.24). It closes the
**implementation** half of the standing `[P0]` blocker
**F-S2107-P10-001** ("POLICY 15 ATTESTATION-LOCATION GATE vacuous"). The
**CI-job wiring** half (adding the `policy-15-attestation-location` required
check to `.github/workflows/`) is explicitly **not** in scope here — it is
tracked as Drift Item **D-969** and will land as a separate follow-up PR. Do
not read this merge as closing F-S2107-P10-001 in full.

---

## Why this exists (root cause + traceability)

ADR-040 §Decision 7 diagnosed a category error: the original gate logic
(§Decisions 1–6) was formally correct but was being *evaluated in the
factory-artifacts worktree*, which tracks zero `*.rs`/`*.bats` files out of
3,743 total — making the gate's pre-check permanently empty and the gate
itself permanently vacuous (confirmed by D-965's ratification of the
then-vacuous v1.4.22 policy text). §Decision 7 Ruling 7(a) requires the gate
to run as a required-check CI job in the **code repo**, not factory-artifacts.

§Decision 9 (F-S2107-P10-002 regression finding) further established that
the obligation is **per-commit**, not push-tip-only — a push-tip-only check
at a squash/rebase point can pass while an intermediate commit in the range
violated the obligation. §Decision 9 Ruling 9(a) names this crate,
`crates/policy15-attestation-gate/`, as the production mechanism satisfying
that per-commit iteration requirement, exercised automatically by
`cargo test --workspace --all-targets` on every CI push.

```mermaid
flowchart LR
    ADR[ADR-040 §Decisions 7-10] --> POL[policies.yaml POLICY 15<br/>v1.4.24 ATTESTATION-LOCATION GATE]
    POL --> CRATE[crates/policy15-attestation-gate/<br/>run_gate + GateOutcome]
    CRATE --> BIN[bin: policy15-attestation-gate<br/>src/main.rs]
    BIN -.->|D-969 follow-up, NOT this PR| CI[".github/workflows/ci.yml<br/>required check: policy-15-attestation-location"]
    style CRATE fill:#90EE90
    style CI fill:#FFD700,stroke-dasharray: 5 5
```

## What changed

```mermaid
graph TD
    Root["Cargo.toml workspace"] -->|new member| Gate["crates/policy15-attestation-gate"]
    Gate --> Lib["src/lib.rs<br/>run_gate() / GateOutcome enum"]
    Gate --> Main["src/main.rs<br/>CLI: policy15-attestation-gate [BASE_BRANCH]"]
    Gate --> Tests["tests/binary_integration_test.rs"]
    style Gate fill:#90EE90
```

- **`src/lib.rs`** — `run_gate(repo, base_branch) -> Result<GateOutcome, GateError>`.
  `GateOutcome` is a plain (non-`#[non_exhaustive]`) enum with four variants —
  `Fail(Vec<FailedCommit>)`, `PassWithActivations(usize)`,
  `PassZeroActivations`, `EmptyOrUnreachable(UnreachableCause)` — deliberately
  typed (not an integer) so that the v1.9 bash-era defect class ("four
  outcomes sharing two exit codes; a control asserting only `$? -ne 0` can
  pass while exercising the wrong variant") is structurally impossible:
  `matches!()` on a closed enum cannot conflate `Fail` and
  `EmptyOrUnreachable`.
  - **Guard 1 (stale-pin, runs *before* `git merge-base`):** if the pinned
    crate path (`PLUGIN_CRATE =
    "crates/hook-plugins/validate-cross-site-correspondence"`) is absent from
    the `HEAD` git tree, the gate returns `EmptyOrUnreachable(StalePin)`
    unconditionally — even if the base branch is also unresolvable. Ordering
    is pinned by `test_run_gate_guard1_stale_pin_beats_unresolvable_base`;
    without it a crate-absent + remote-removed repo would silently report the
    less-actionable `EmptyRange` instead.
  - **Per-commit iteration:** walks `merge-base..HEAD`, and for each commit
    checks whether it touches `*.rs`/`*.bats` under the pinned crate path; if
    so, the pinned `red-gate-log.md` **must** exist at that commit and contain
    a `### <Pass-N> assertion-site attestation (<PARENT-SHA>)` heading, or the
    commit is reported `Fail`.
  - **Parentless-commit handling:** root/shallow-clone-boundary commits are
    skipped with an explicit `WARNING` (this PR's F-1 fix — see below), per
    ADR-040 §Decision 9 Ruling 9(c) — previously a silent skip.
- **`src/main.rs`** — CLI binary. `BASE_BRANCH` precedence: explicit CLI
  argument > `BASE_BRANCH` env var > `develop` default, via clap's native
  `env`-backed argument derive (this PR's F-3 fix; a regression test pins the
  CLI-arg-wins case). Prints a greppable outcome identifier to stdout and
  exits 0 (either PASS variant), 2 (`Fail` / `EmptyOrUnreachable`), or 1 (hard
  error — git missing, I/O failure).
- **`tests/binary_integration_test.rs`** — black-box integration tests
  exercising the compiled binary against fixture repos.
- **`Cargo.toml` (root)** — adds `crates/policy15-attestation-gate` to
  `[workspace] members`.
- **`CLAUDE.md`** (commit `f81ab5a6`, incidental, already on this branch) —
  updates the fuel-exhaustion diagnostics row in the Factory Hook
  Diagnostics table for the `FUEL_EXHAUSTED` sentinel (unrelated to the gate
  crate; carried on this branch from prior work, included for completeness).

### Mutation testing (`cargo-mutants`)

The branch's most recent commits closed every mutant surviving an earlier
run:

- `d2a3176a` — killed the guard-1 stale-pin mutant (guard-1 return
  neutralized) and a coarse-assertion mutant (`if count != 1` → `if false`)
  by replacing a boolean `!is_pass()` check with a specific-variant
  `matches!()` assertion. Both reverted-and-reproduced before the fix landed.
- `38faf75f` — added the F-1/F-2/F-5/F-6 mutant-proving tests from
  test-writer review and the corresponding code fixes (parentless-skip
  `WARNING` emission, `BASE_BRANCH` precedence correction, stale ADR-040
  `v1.8` doc-cite refreshed to `v1.17`).

Current state: **`cargo-mutants` 0 missed / 46 caught / 4 unviable** against
`crates/policy15-attestation-gate/`.

### Test suite

18 unit tests + 5 integration tests (`tests/binary_integration_test.rs`) + 1
doctest — all green. `cargo fmt --check --all` and
`cargo clippy --workspace --all-targets -- -D warnings` both clean.

### CI (full workspace, not crate-scoped)

The crate is a new workspace member, so `.github/workflows/ci.yml`'s
`cargo-host` job (fmt + clippy + `cargo test --workspace --all-targets`) and
`validate` job exercise it end-to-end on this PR — not `-p
policy15-attestation-gate` in isolation. See CI status checks on this PR for
the live run; this PR does not merge until that run is fully green.

---

## Risk Assessment

- **Blast radius:** additive-only. New workspace crate; no existing crate's
  source changed except the root `Cargo.toml` members list and one
  documentation table row in `CLAUDE.md`. The gate binary is not yet invoked
  by any CI workflow (that wiring is D-969, out of scope here) — so this PR
  has **zero runtime effect** on any existing pipeline, hook, or CI job.
- **User impact if this PR is wrong:** none at merge time — the binary is
  dead code from CI's perspective until D-969 wires it in. Risk surfaces only
  when D-969 lands and the required check goes live; that PR will re-validate
  the gate's behavior against real commit ranges before making it a required
  check.
- **Risk level:** LOW.

## Pre-Merge Checklist

- [ ] All CI status checks passing (full workspace `cargo-host` + `validate`
      + bats — not crate-scoped)
- [ ] `cargo fmt --check --all` clean
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [ ] No critical/high security findings unresolved
- [ ] pr-reviewer fresh-eyes review: APPROVE
- [ ] code-reviewer (cognitive-diversity) review: no unresolved blocking findings
- [ ] Branch synced with latest `origin/develop` prior to merge

## Traceability

| Artifact | Reference |
|---|---|
| Blocker closed (implementation half) | `[P0] F-S2107-P10-001` |
| Follow-up (CI-job wiring, separate PR) | Drift Item `D-969` |
| Governing ADR | `ADR-040` §Decisions 7–10 |
| Governing policy | `policies.yaml` POLICY 15, v1.4.24 (ATTESTATION-LOCATION GATE clause, unchanged text since v1.4.23/D-970) |
| Ratification | ADR-040 v1.12/v1.13/v1.15/v1.16/v1.17 |


Zious11 (commented):
# PR #777 — Final Fresh-Eyes Review (pr-reviewer)

- **PR:** #777 — fix(policy15-gate): implement POLICY 15 attestation-location gate as Rust crate
- **Branch:** feature/policy15-gate-rust → develop
- **Reviewed HEAD:** 010e6140352d12c3101c983b6e848ae7c7d8a5dd
- **Reviewer:** vsdd-factory:pr-reviewer (fresh-eyes)
- **Date:** 2026-08-15

## VERDICT: APPROVE (ready to merge)

No blocking findings. The crate implements the ADR-040 §Decisions 7-10 four-outcome
POLICY 15 ATTESTATION-LOCATION GATE correctly. This is a fix-PR / CI-tooling delivery
(no BCs, no holdout/demo evidence applies). CI-job wiring (D-969) is intentionally out
of scope and disclosed — NOT flagged.

## Scope note

`gh pr diff 777` (authoritative merge-base diff, 45.7KB) is exactly: the
`crates/policy15-attestation-gate/` crate (Cargo.toml, src/lib.rs, src/main.rs,
tests/binary_integration_test.rs) + root `Cargo.toml` member addition + one incidental
`CLAUDE.md` doc-row edit (FUEL_EXHAUSTED sentinel). Local `develop` is stale, so a
`develop...HEAD` three-dot stat shows unrelated already-merged noise — only the real PR
surface was reviewed.

## Independent verification (all re-run, not trusted from description)

- `cargo test -p policy15-attestation-gate --all-targets` → 18 unit + 5 integration, all green
- `cargo test -p policy15-attestation-gate --doc` → 1 doctest green
- `cargo clippy -p policy15-attestation-gate --all-targets -- -D warnings` → exit 0
- `cargo fmt --check --all` → exit 0
- cargo-mutants (0 missed / 46 caught / 4 unviable) NOT re-run (long) — claim trusted

## 1. Four-outcome semantics — CORRECT

Fail / PassWithActivations / PassZeroActivations / EmptyOrUnreachable are a typed closed
enum (deliberate absence of `#[non_exhaustive]`). Exit-code mapping is isolated to
`is_pass()`/`exit_code()`; each variant has a greppable `identifier()`. The bash-era
"four-outcomes / two-exit-codes" defect class is genuinely structurally closed —
`matches!()` on a closed enum cannot conflate `Fail` and `EmptyOrUnreachable`.

Priority order in `run_gate_inner` is FAIL > EMPTY-or-UNREACHABLE > PASS. A real
violation is never masked by an empty-diff commit (fail_list checked first). Worst case
is a conservative false-positive-FAIL, never a false-negative-PASS — correct direction
for a gate.

## 2. Guard-1 stale-pin-before-merge-base ordering — CORRECT IN CODE

`run_gate` (lib.rs:194-209) runs `tree_path_exists(HEAD, PLUGIN_CRATE)` and returns
`StalePin` BEFORE computing merge-base. `test_run_gate_guard1_stale_pin_beats_unresolvable_base`
exercises it through `run_gate` (guard 1), distinct from the `run_gate_from_merge_base`
guard-2 path. Uses `git cat-file -e` (tree query) not filesystem `is_dir()` — the
defect-3 regression (`test_disk_present_tree_absent_is_stale_pin`) confirms this. The
ordering is real in the code, not merely asserted by a test name.

## 3. Dead code / unused imports / logic bugs in run_gate/run_gate_from_merge_base/run_gate_inner — NONE

- All imports used; clippy clean.
- `run_gate_from_merge_base` is pub-but-test-only by design (the "testable core" that
  bypasses the origin remote lookup) — legitimate API, not dead code.
- clap `#[arg(env = "BASE_BRANCH", default_value = "develop")]` yields correct
  CLI-arg > env > default precedence (F-3 regression pins it).
- Merge commits diff against `^1` (standard); individual second-parent commits are also
  evaluated in-range, so activation coverage is complete.
- `count != 1` correctly enforces exactly-once (F-2 covers count==2).

## Findings

### [LOW / non-blocking] Comment rot in tests/binary_integration_test.rs

The `test_f1_..._expected_red_pending_implementer` docstring (lines 444-457, 487-491)
still declares assertion (b) "EXPECTED RED, PENDING IMPLEMENTER" and "this assertion
currently fails today by design / DO NOT weaken." But the implementer already wired the
WARNING (`eprintln!` in lib.rs:244-246, per commit 38faf75f), so the test is GREEN. The
test name and docstring are now factually stale and will confuse a future reader.
Recommend renaming the test (drop `_expected_red_pending_implementer`) and updating the
docstring to describe it as an enforced regression. No functional impact — the assertion
still guards the right behavior.

### [INFORMATIONAL / not a defect] Fail-closed empty-diff prioritization

`EmptyOrUnreachable(UnmeasurableDiff)` outranks `PassWithActivations`, so a range with
one compliant activating commit plus any `--allow-empty` commit reports
EMPTY-or-UNREACHABLE (exit 2) rather than PASS. This is disclosed, ADR-040-ratified
fail-closed design and is conservative-safe, but could surface as a CI false-positive
once D-969 wires the gate live. The PR already commits D-969 to re-validate against real
commit ranges — appropriate place to confirm. Flagged only for D-969's awareness.

## READY

```
covered_sha: 010e6140352d12c3101c983b6e848ae7c7d8a5dd
verdict: APPROVE
blocking_findings: 0
low_findings: 1 (comment rot in F-1 integration test docstring — recommend fix, does not block)
informational: 1 (fail-closed empty-diff prioritization — D-969 follow-up awareness)
validation: 18 unit + 5 integration + 1 doctest green; clippy -D warnings exit 0; fmt clean — all independently re-run
```

Zious11 (commented):
# Fresh-eyes PR review — PR #777 @ `010e6140`

**Verdict: REQUEST_CHANGES** — 2 BLOCKING (HIGH), 3 advisory-strong (MEDIUM), 7 advisory (LOW/NIT).

Reviewed with fresh context against the diff, the PR description, and ADR-040 (v1.17). All findings below
were derived by **executing** the crate, not by reading it — which matters here, because ADR-040's own
§Consequences records that every defect in the six bash-era iterations was found by extraction-and-execution
and *none* by code review. Two of the findings below reproduce that pattern exactly.

Evidence I re-ran independently (not merely cited):

| Check | Result |
|---|---|
| `cargo test -p policy15-attestation-gate` | 18 unit + 5 integration + 1 doctest, 0 failed |
| `cargo clippy -p policy15-attestation-gate --all-targets -- -D warnings` | clean |
| `cargo fmt --check -p policy15-attestation-gate` | clean |
| `cargo mutants -p policy15-attestation-gate --file src/main.rs` | 5 mutants, 5 caught, 0 missed |
| **Gate run against this PR's own HEAD** (`./policy15-attestation-gate develop`) | **`FAIL: obligation violated`, exit 2** |

---

## BLOCKING

### [BLOCKING][HIGH][NEW] H-1 — A routine `merge develop into feature-branch` commit activates the gate and produces a spurious `Fail`. Demonstrated on this PR's own HEAD.

`run_gate_inner` iterates every commit in `merge_base..HEAD` and, for each, computes
`git diff --name-only <C>^1 <C>`. For a **merge commit**, that first-parent diff is the *entire* set of
changes the merged branch brought in — so a "sync with develop" merge is attributed wholesale to the merge
commit. If any of those files is a `*.rs`/`*.bats` under `PLUGIN_CRATE`, the merge commit becomes an
activating commit and demands a `### … assertion-site attestation (<its first parent>)` heading that no
author ever wrote, or could have written.

This is not hypothetical. Run against this very branch, exactly as ADR-040 §Decision 9 Ruling 9(c) item 5
prescribes CI should invoke it (`HEAD` = `pull_request.head.sha` = `010e6140`, base `develop`):

```
$ ./target/debug/policy15-attestation-gate develop
  FAIL 010e6140352d: attestation heading missing in log
FAIL: obligation violated
EXIT=2
```

```
$ git diff --name-only 010e6140^1 010e6140 | grep 'validate-cross-site-correspondence/.*\.rs'
crates/hook-plugins/validate-cross-site-correspondence/src/arm_a1.rs
crates/hook-plugins/validate-cross-site-correspondence/src/arm_a2.rs
crates/hook-plugins/validate-cross-site-correspondence/src/arm_b.rs
crates/hook-plugins/validate-cross-site-correspondence/src/arm_e.rs
crates/hook-plugins/validate-cross-site-correspondence/src/dispatch.rs
crates/hook-plugins/validate-cross-site-correspondence/src/frontmatter.rs
crates/hook-plugins/validate-cross-site-correspondence/src/lib.rs
crates/hook-plugins/validate-cross-site-correspondence/src/main.rs
```

`010e6140` is the PR's own "Merge remote-tracking branch 'origin/develop'" commit — created to satisfy this
PR's own pre-merge checklist item *"Branch synced with latest `origin/develop` prior to merge."* The repo's
documented workflow therefore **manufactures the failing shape on essentially every PR that syncs with
develop after the pinned crate has moved.** The moment D-969 makes `policy-15-attestation-location` a
required check, the gate goes red on well-formed PRs and the first remediation anyone reaches for will be to
weaken it — which is precisely the vacuity spiral ADR-040 exists to end.

**Why I do not think ADR-040 already settles this.** Ruling 9(d) (v1.17) ratifies *first-parent endpoint-diff
scope* and rules second-parent **history** out of scope. It reasons about whether content can *hide* from the
gate — a false-negative argument. It never addresses the inverse: that first-parent endpoint-diff makes the
merge commit itself an **activating** commit for work it did not author. Ruling 9(c) item 5 anticipates a
merge commit only in the GitHub-*synthetic* `refs/pull/N/merge` form and fixes it by pinning `ref:` to
`head.sha`; that fix does nothing for a *real* merge commit inside the PR branch's own history. The design
center Ruling 9(a) describes — the `67ffbdcc`/`38c70f9e`/`5370db80` case study — is a linear commit sequence.
So this case is a genuine gap, not a code-vs-spec conflict where the spec wins.

**Suggested resolution** (routing: `architect` for the ADR ruling, then `implementer`) — pick one and pin it
with a test:

1. Skip commits with more than one parent (`git rev-parse --verify <C>^2` succeeds ⇒ merge ⇒ skip), on the
   same reasoning as Ruling 9(d) fact 1: a real merge's files are already visible in the diff of whichever
   *non-merge* commit introduced them, so skipping merges loses no coverage against the smuggling threat
   Ruling 9(d) analyses. Concretely: `git log --no-merges <range> --format=%H`.
2. Or evaluate the range with `--first-parent` so second-parent commits are not walked and the merge
   commit's cumulative diff is the only accounting — but then the merge commit still needs an attestation
   rule, so (1) is cleaner.
3. Or ratify merge-commit activation as intended in a new ADR-040 ruling and state the author obligation
   explicitly (write an attestation heading against the merge's first parent when syncing) — I'd argue
   against this, but it is a legitimate architect call.

Whichever is chosen, add a fixture with an activating change reaching the tip **through a merge commit**.
The existing `create_range_containing_parentless_commit` helper already builds a merge, but the only file it
introduces is non-activating (`unrelated/extra.md`), so merge-commit activation is untested today.

---

### [BLOCKING][HIGH][NEW] H-2 — `git diff --name-only` output is C-quoted under the default `core.quotePath=true`, so an activating file with a non-ASCII name is silently invisible: the gate returns `PASS-zero-activations` (exit 0).

`git_diff_name_only` parses raw `git diff --name-only` stdout. Git's default `core.quotePath=true` renders
any path containing non-ASCII bytes, control characters, `"` or `\` as a C-quoted string with surrounding
double quotes and octal escapes. Both predicates in `run_gate_inner` then fail: `f.starts_with(&crate_prefix)`
fails on the leading `"`, and `f.ends_with(".rs")` fails on the trailing `"`.

Reproduced end-to-end against the compiled binary:

```
$ git diff --name-only HEAD^1 HEAD
"crates/hook-plugins/validate-cross-site-correspondence/src/caf\303\251.rs"

$ git -c core.quotePath=false diff --name-only HEAD^1 HEAD
crates/hook-plugins/validate-cross-site-correspondence/src/café.rs

$ policy15-attestation-gate develop
PASS-zero-activations
EXIT=0
```

A `.rs` file added under the pinned crate, with no `red-gate-log.md` attestation anywhere, yields a **green
required check.** That is a silent vacuous pass — the exact outcome class this crate was built to make
structurally impossible. The same parsing shape also means a path containing a newline would split into two
bogus entries.

I want to be calibrated about the *likelihood*: non-ASCII filenames in a Rust hook-plugin crate are unusual,
and I have no reason to think this is being exploited. I am marking it blocking not on exploit risk but on
CLAUDE.md Canonical Principle Rule 4 — this is an AI-authored defect with a **one-line, zero-risk fix**, and
deferring a one-line fix to a follow-up is the defer-pattern smell the principle names explicitly.

**Suggested fix** — use NUL-delimited output, which is quoting-immune, in `git_diff_name_only`:

```rust
let output = run_git(repo, &["diff", "--name-only", "-z", from, to])?;
// …
let files = String::from_utf8_lossy(&output.stdout)
    .split('\0')
    .filter(|s| !s.is_empty())
    .map(|s| s.to_string())
    .collect();
```

A `-c core.quotePath=false` prefix on `run_git` is the cheaper alternative, but `-z` is strictly stronger
(it also fixes embedded newlines) and is the conventional choice for machine-parsed git output. Please add a
regression fixture committing a non-ASCII-named `.rs` under `PC` and asserting `Fail`/activation rather than
`PassZeroActivations` — otherwise this silently regresses.

---

## MEDIUM (advisory, but I'd fix all three before merge)

### [SUGGESTION][MEDIUM][NEW] M-1 — `main.rs`'s `_ => {}` wildcard defeats the exhaustiveness invariant the crate's own doc comment claims as its central safety property.

`lib.rs` states the design intent unambiguously:

```rust
// NOTE: #[non_exhaustive] is deliberately absent so that exhaustive match at
// every call site is a compile error when a new outcome is added — the same
// property that guards PluginResult in the dispatcher crate.
```

But the crate's only real call site does this:

```rust
match &outcome {
    GateOutcome::Fail(failed) => { … }
    GateOutcome::EmptyOrUnreachable(cause) => match cause { … }
    _ => {}                       // ← swallows every future variant
}
```

Adding a fifth `GateOutcome` variant compiles clean here and silently prints no detail line. `is_pass()` has
the same leak in softer form — `matches!(self, PassWithActivations(_) | PassZeroActivations)` means a new
variant defaults to non-pass with no compile error (fail-closed, so lower impact, but still not the stated
property).

The whole rationale for omitting `#[non_exhaustive]` is forfeited by one wildcard. Fix is one line:

```rust
GateOutcome::PassWithActivations(_) | GateOutcome::PassZeroActivations => {}
```

Worth a short comment on `is_pass()` noting the deliberate fail-closed default, so the next reader doesn't
have to re-derive it. The inner `match cause` on `UnreachableCause` is already correctly exhaustive — good.

### [SUGGESTION][MEDIUM][NEW] M-2 — The F-1 test ships a large, emphatic, and now-false "EXPECTED RED, PENDING IMPLEMENTER" doc block, and its function name encodes the same stale claim.

`test_f1_parentless_commit_in_range_is_skipped_with_warning_expected_red_pending_implementer` carries ~15
lines of doc asserting the WARNING assertion "currently fails today by design", "is the implementer's target
for wiring the ADR-040 §Decision 9 Ruling 9(c) WARNING", and "DO NOT weaken this assertion to make it pass."

The WARNING was implemented in **the same commit that added this test** (`38faf75f` contains both the
`eprintln!("WARNING: commit {commit_sha} has no first parent …")` in `lib.rs` and the test). The test is
GREEN as merged — I ran it. So the file now instructs future readers that a passing test is expected to fail,
and names an implementer task that is already done.

This matters beyond tidiness: a stale Red-Gate marker is exactly what causes a later agent to "fix" a
green test by deleting it, or to re-do completed work. Please rename to something like
`test_f1_parentless_commit_in_range_is_skipped_with_warning` and rewrite the doc block to describe the
implemented behavior and cite ADR-040 §Decision 9 Ruling 9(c) item 4 as the source of the requirement.
The same stale framing appears in the file-level module doc ("F-1's WARNING assertion is EXPECTED RED
pending implementer") and in the `lib.rs` test-section comment at the "Coverage-gap fix-pass tests" banner —
sweep all three (TD-VSDD-060 sibling-site sweep).

### [SUGGESTION][MEDIUM][NEW] M-3 — `FailReason::AttestationMissing` misreports the duplicate-heading case; the operator sees "attestation heading missing in log" when there are two.

The obligation is *exactly one* heading (`if count != 1`), and `test_f2_two_attestation_headings_same_parent_fails`
deliberately exercises `count == 2`. That path is correctly a `Fail` — but it is reported through
`FailReason::AttestationMissing`, and `main.rs` prints `"attestation heading missing in log"`. An author who
wrote *two* attestations is told the attestation is missing, and will most likely respond by adding a third.

Given this crate's stated purpose is diagnostic precision within a shared exit code, the diagnosis should
distinguish the arms. Suggest either a third variant (`FailReason::AttestationNotExactlyOnce { count: usize }`)
or, minimally, carrying the count and printing `"expected exactly 1 attestation heading, found {count}"`.
Cheap, and it makes the F-2 control assert on a distinguishable reason rather than sharing one with the
`count == 0` case.

---

## LOW / NIT (advisory)

- **[NIT][LOW][NEW] L-1 — `Cargo.lock` carries transitive bumps for *existing* crates, mildly contradicting the PR body's "additive-only / zero runtime effect."** Commit `d5a90e74` re-resolves `windows-sys 0.52.0 → 0.61.2` (4 sites: `errno`, `rustix`, `tempfile`, `winapi-util`), `socket2 0.5.10 → 0.6.3` (`hyper-util`), and `itertools 0.11.0 → 0.14.0` (`prost-derive`). None are required by adding `clap`. CI is green so this is fine functionally, and the *source*-level claim in the body is accurate — but "no existing crate's source changed" understates it: existing crates now build against different transitive dependency versions. Either revert the incidental bumps or add one sentence to the Risk Assessment acknowledging them.

- **[NIT][LOW][NEW] L-2 — `main.rs`'s `FailReason::AttestationMissing` stderr arm is unexercised by any test.** All five integration tests produce `LogAbsent`, `UnmeasurableDiff`, `PASS-N`, or `PASS-zero`; none drive `AttestationMissing` through the binary. I checked whether this contradicts the "0 missed" claim and it does **not** — I re-ran `cargo mutants --file src/main.rs` and got 5/5 caught, because cargo-mutants generates no mutant for that match-arm literal. Worth stating plainly for the record: **mutation score is not coverage.** A one-line variation on the F-6 branch-1 fixture (seed the log with a non-matching heading instead of omitting it) closes it.

- **[NIT][LOW][NEW] L-3 — `UnmeasurableDiff` retains only the *last* empty-diff commit.** `empty_diff_commit = Some(commit_sha.clone())` overwrites on each hit, so with several `--allow-empty` commits in range the operator is pointed at one arbitrary commit. `Vec<String>` mirrors the `fail_list` treatment and costs nothing.

- **[NIT][LOW][NEW] L-4 — The `EmptyRange` identifier is factually wrong in the unresolvable-base case.** `git merge-base` failing (no `origin` remote, unknown branch) returns `EmptyRange`, whose identifier reads `"EMPTY-or-UNREACHABLE: git range returned no commits"` — but no range was ever computed. `main.rs`'s stderr detail line gets this right ("…or base branch unresolvable") and the variant's doc comment lists it as a cause, so this is only the greppable stdout string. Since controls grep that string (§Decision 10 Req 4), consider splitting `EmptyRange` from an `UnresolvableBase` cause, or neutralising the wording to `"EMPTY-or-UNREACHABLE: no commits in range or base unresolvable"`.

- **[NIT][LOW][NEW] L-5 — The `lib.rs` `#[cfg(test)]` `Repo` fixture does not assert git-command success, unlike its `tests/` twin.** `Repo::commit` and `Repo::git` discard the exit status. `tests/binary_integration_test.rs` got this right (`git_stdout` asserts, `commit` asserts). The asymmetry matters: on a machine with global `commit.gpgsign = true` (or `gpg.format=ssh` misconfigured), fixture commits fail silently and some controls pass vacuously — `test_stale_pin`, for instance, would still see `StalePin` from guard 1 and go green having exercised nothing. Suggest mirroring the assert-on-failure helper into `lib.rs`, and adding `-c commit.gpgsign=false` to `Repo::new`'s git config so the suite is hermetic.

- **[NIT][LOW][NEW] L-6 — The `#![allow(clippy::expect_used, clippy::unwrap_used, clippy::panic)]` attributes are no-ops.** `[workspace.lints.clippy]` in the root `Cargo.toml` is an empty table, so none of those lints are enabled. Harmless and arguably good documentation of intent — but if the intent is that production paths are lint-enforced against `unwrap`/`expect` (CLAUDE.md Conventions), that enforcement does not currently exist workspace-wide. Out of scope for this PR; flagging because the allows imply a policy that isn't switched on. (I verified by inspection that `lib.rs` and `main.rs` non-test code contain **zero** `unwrap()`/`expect()`, and `println!`/`print!` appear only in `main.rs` and the doctest — item 5 of the review brief is clean.)

- **[NIT][LOW][NEW] L-7 — The parentless skip is observable only as a stderr string, not as an outcome variant.** If *every* commit in range were parentless-skipped, the gate returns `PassZeroActivations`, exit 0 — an unqualified green whose only trace is a `WARNING` line. The crate's own thesis is that outcomes must be named and asserted as variants rather than inferred from strings (`PassZeroActivations` is documented as "named (not anonymous exit-0) so it is observable across CI runs"). A skip is the same kind of event. ADR-040 §Decision 9 Ruling 9(c) item 4 mandates only WARNING-and-skip, so the current behavior is spec-conformant and I am **not** asking for a change here — but if a future ruling revisits it, threading a `skipped: Vec<String>` onto the pass variants would let controls assert on it the way they assert on everything else.

---

## What I verified and found correct (no rubber-stamp)

Each item from the review brief, with what I actually did:

1. **Four-outcome design / outcome conflation — CORRECT.** `GateOutcome` is a closed enum; `#[non_exhaustive]` is deliberately absent and that intent is documented at the type. Every one of the four variants has a dedicated control asserting the *variant* via `matches!` plus its exact `identifier()` string and `exit_code()` — not `!is_pass()`. The specific historical conflation (stale-pin vs empty-range) is pinned from **both** directions: `test_guard_ordering_stale_pin_beats_empty_range` (through `run_gate_from_merge_base`, guard 2) and `test_run_gate_guard1_stale_pin_beats_unresolvable_base` (through `run_gate`, guard 1, before the merge-base call). The duplicated guard is justified in a comment naming the pinning test. The `FAIL > EMPTY-or-UNREACHABLE > PASS` priority ordering in `run_gate_inner` is correct and fail-closed. Defect-3 (disk-present/tree-absent) is covered by `test_disk_present_tree_absent_is_stale_pin` and the implementation correctly uses `git cat-file -e` rather than `is_dir()`. The `^### ` line anchor is real (`starts_with("### ")` rejects `#### `) and `test_prose_sha_does_not_cause_false_count` proves prose quoting the same SHA does not inflate the count. **The only conflation bug I found is H-1, which is a scope question, not an outcome-mapping question.**
2. **Parentless-commit handling — CORRECT.** `commit_has_parent` uses `git rev-parse --verify <C>^1` and, on failure, emits an explicit `WARNING: commit <sha> has no first parent (root commit or shallow-clone boundary) — skipping POLICY 15 evaluation for this commit` on stderr and `continue`s. That is WARNING-and-skip, no crash, no silent skip — matching ADR-040 §Decision 9 Ruling 9(c) item 4 verbatim. The `create_range_containing_parentless_commit` fixture is genuinely clever and genuinely load-bearing: it uses `git commit-tree` with no `-p` to synthesize a real parentless commit, asserts the fixture invariant (`rev-parse <root>^1` must fail), then merges with `--allow-unrelated-histories` so `git merge-base` still resolves through the first parent while `git log base..HEAD` really contains the parentless commit. That is the hard part of testing this, and it was done properly. My only complaint is the stale doc block (M-2).
3. **`BASE_BRANCH` CLI-vs-env precedence — CORRECT.** `#[arg(env = "BASE_BRANCH", default_value = "develop")]` gives clap's native CLI > env > default. I did not take this on the test's word — I built the binary and drove all four paths against a scratch repo where only `origin/cli-branch` resolves:

   | Invocation | Result |
   |---|---|
   | `gate cli-branch` | `PASS-zero-activations` (exit 0) |
   | `BASE_BRANCH=env-branch gate` | `PASS-zero-activations` (exit 0) |
   | `BASE_BRANCH=nope-does-not-exist gate cli-branch` | `PASS-zero-activations` (exit 0) — **CLI wins** |
   | `gate` (no arg, no env) | `EMPTY-or-UNREACHABLE` — falls back to `develop` |

   `test_f3_cli_arg_wins_over_base_branch_env_var` pins it correctly, and its fixture is well-constructed: it deliberately adds a non-crate commit past `base` so the range is non-empty, closing the loophole where `EmptyRange` would arise regardless of which name won.
4. **Error taxonomy — good.** `GateError` is a proper `thiserror` enum; semantic outcomes (including stale pin and unresolvable base) correctly return `Ok(GateOutcome)` rather than `Err`, and that boundary is documented on the type. Hard errors exit 1, distinct from the gate's 0/2. One soft spot worth knowing about: `git_show_file` maps *any* non-zero `git show` exit to `FileAbsent`, so a genuinely corrupt object would be reported as `LogAbsent` → `Fail`. That is fail-closed and therefore acceptable; not raising it as a finding.
5. **Code quality — clean.** Zero `unwrap()`/`expect()` outside `#[cfg(test)]` code and the `no_run` doctest. `println!` appears only in `main.rs` (the greppable identifier to stdout) and the doctest; detail lines correctly go to stderr, keeping stdout single-purpose for CI grepping. SHA truncation is byte-slicing but guarded by `len() >= 12` on hex-ASCII, so it cannot panic. `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` both clean on my re-run.
6. **Mutation-testing claim — CREDIBLE, and the assertions are load-bearing.** I traced the plausible mutants by hand against the fixtures and they are genuinely killed, not coarsely covered:
   - `starts_with(&crate_prefix) → true` is killed by `test_rs_outside_crate_does_not_activate` (a `.rs` in a different crate would activate → `Fail` vs expected `PassZeroActivations`).
   - `ends_with(".rs") || ends_with(".bats") → true` is killed by `test_pass_zero_activations`, whose range includes a commit touching `{PC}/docs/placeholder.md` — inside the crate prefix, wrong extension. That fixture's ordering (base captured *before* the seed commit) is subtle and correct; it looks incidental but is what makes the mutant die.
   - `activations += 1` neutralisation is killed by `test_multiple_activations_all_compliant` asserting `PassWithActivations(2)` specifically.
   - `count != 1` comparison flips are killed from both sides — `count == 0` by `test_positive_2`, `count == 2` by `test_f2`.
   - The `>= 12` truncation guard is killed by F-6's paired assertions: the truncated SHA must be present **and** the full 40-char SHA must be absent. That negative assertion is what makes it load-bearing rather than decorative.
   - Guard-1 return neutralisation is killed by `test_run_gate_guard1_stale_pin_beats_unresolvable_base`, which was clearly written for exactly that mutant.

   I found **no** coarse or non-load-bearing assertion in the suite. `test_identifier_strings_are_greppable` is the weakest (pure string spot-check on hand-constructed values) but it is supplementary, not the primary control for any variant. The `d2a3176a` commit message's claim that a `!is_pass()` check was replaced by a specific-variant `matches!()` is visible in the diff at `test_unresolvable_base_fails_closed`, with a comment explaining why — that is the right instinct and I'd like to see it applied to `is_pass()` itself (M-1).

**Diff coherence, size, and description:** 7 files, +1865/−7. All changes relate to the stated purpose except the `CLAUDE.md` fuel-diagnostics row, which the description discloses as incidental and carried from prior work — acceptable disclosure. The description is unusually accurate and the D-969 scope boundary is stated clearly and repeatedly; my only correction is L-1. Commit messages are conventional-format with clear scopes. Size is above the 500-line flag threshold but is almost entirely new-crate source and tests with no existing-file modification, so I do not consider it a review-quality risk. No demo-evidence requirement applies — this is a fix/CI-tooling PR with no ACs, correctly delivered on the fix-PR track.

---

## Summary

| ID | Severity | Blocking | Category | Finding |
|---|---|---|---|---|
| H-1 | HIGH | **yes** | correctness | Merge commits activate the gate → spurious `Fail`; reproduced on this PR's own HEAD |
| H-2 | HIGH | **yes** | correctness / vacuity | `core.quotePath` C-quoting hides non-ASCII activating paths → silent `PASS-zero-activations` |
| M-1 | MEDIUM | no | design integrity | `_ => {}` in `main.rs` defeats the documented non-`non_exhaustive` invariant |
| M-2 | MEDIUM | no | doc drift | Stale "EXPECTED RED / PENDING IMPLEMENTER" block + test name; fix landed in the same commit |
| M-3 | MEDIUM | no | diagnostics | Duplicate-heading case reported as "attestation missing" |
| L-1 | LOW | no | coherence | Incidental transitive `Cargo.lock` bumps vs "additive-only" claim |
| L-2 | LOW | no | coverage | `main.rs` `AttestationMissing` arm untested (no mutant targets it) |
| L-3 | LOW | no | diagnostics | `UnmeasurableDiff` keeps only the last empty-diff commit |
| L-4 | LOW | no | diagnostics | `EmptyRange` identifier wrong for unresolvable-base |
| L-5 | LOW | no | test hygiene | `lib.rs` fixture does not assert git success; no `commit.gpgsign=false` |
| L-6 | NIT | no | lint config | `[workspace.lints.clippy]` empty → the `#![allow(...)]` attrs are no-ops |
| L-7 | NIT | no | observability | Parentless skip observable only as a stderr string, not a variant (spec-conformant) |

All findings are **NEW** — none is a re-confirmation of a previously-reported finding, and I had no access to prior review history.

**H-1 is the one I would not merge without.** H-2 is blocking only under CLAUDE.md Canonical Principle Rule 4 (one-line fix, in-scope), and I'd accept a reasoned architect override on it far more readily than on H-1. The rest can ride along in the same fix burst.

This is careful, well-motivated work — the guard-ordering tests and the parentless-commit fixture in particular are better than what I usually see for a CI-tooling crate. H-1 is not a coding mistake; it is a scope assumption (linear history) that the repo's own sync-with-develop workflow violates, and it took executing the binary against this branch to surface it.

Zious11:
## Fix verification — H-1/H-2/M-1/M-3 independently re-confirmed at HEAD `6854b25f`

Following code-reviewer's REQUEST_CHANGES (2 BLOCKING + 3 MEDIUM at HEAD `010e6140`), commits `6c1ad212` and `6854b25f` applied ADR-040 v1.18 fixes. I independently re-verified each finding by construction and execution against the compiled binary (not by trusting commit messages), per TD-VSDD-059:

**H-1 (merge-commit spurious FAIL) — FIXED, verified both directions:**
- Scenario A (clean merge pulling in already-attested content from a second parent): old two-dot diff would have flagged the file as newly-changed (reproduced this precondition); the fixed binary correctly returns `PASS-zero-activations` — no spurious FAIL.
- Scenario B (merge commit resolving a genuine conflict, introducing content unattested on either parent): combined diff correctly shows the file as changed; gate correctly returns `FAIL` on all 3 commits (both linear commits + the merge). Bypass has NOT been opened.

**H-2 (quotePath vacuous PASS) — FIXED:** constructed a non-ASCII-named `.rs` file under the pinned crate with no attestation. Default `git diff` C-quotes the path (reproduced the vulnerability precondition); the fixed binary correctly detects it and returns `FAIL` (previously would have silently returned `PASS-zero-activations`).

**M-1 (exhaustiveness) — FIXED:** confirmed via code inspection — `main.rs`'s match on `GateOutcome` now has explicit `PassWithActivations(_)`/`PassZeroActivations` arms, no `_ => {}` wildcard.

**M-3 (ambiguous-attestation diagnosis) — FIXED:** confirmed `FailReason::AttestationAmbiguous { count }` is a genuine distinct variant from `AttestationMissing`, with a distinguishable message, tested explicitly for `count: 2`.

**Additional re-verification at this HEAD:** `cargo fmt --check --all` clean, `cargo clippy --workspace --all-targets -D warnings` clean, `cargo test --workspace --all-targets` 194 test-result blocks / 0 failures (crate: 22 unit + 5 integration + 1 doctest), `cargo mutants -p policy15-attestation-gate`: 57 tested / 53 caught / 4 unviable / 0 missed. CI: 13/13 checks green.

No new issues found during this verification pass. Proceeding to merge.

PR #782, opened 2026-08-25
fix(ci): rc.24 CI-green prep — test-harness fixes + h2 0.4.19 RUSTSEC-2026-0258 bump

Zious11:
# fix: rc.24 CI-green prep (test/CI/dependency-only)

**Type:** Fix PR (release-prep, no product logic changed)
**Target release:** v1.0.0-rc.24
**Branch:** `fix/rc24-ci-green` → `develop`

Makes `develop` CI-green ahead of cutting v1.0.0-rc.24. All three commits are
test-harness, CI-gate, or dependency-lockfile fixes — no application/dispatcher
logic changed.

---

### 1. `test(ci): stop merging cargo metadata stderr into jq in wasmtime version gate`
Fixes an AC-003 false-fail in
`plugins/vsdd-factory/tests/s21-12-version-and-deny-gate.bats`: the wasmtime
version-gate test was merging `cargo metadata`'s stderr into the stdout stream
consumed by `jq`, so any stderr noise (warnings, progress output) corrupted the
JSON payload and caused spurious gate failures unrelated to the actual
wasmtime version check.

**Behavior change disclosed (per review cycle 1 feedback):** beyond the
stderr/stdout separation, this commit also converts a hard-fail into a
`skip` when `cargo metadata --locked` itself fails or returns no output (an
environment-gap case, e.g. a missing/broken toolchain in the CI runner) —
this is a deliberate loosening of the gate's failure mode for that specific
environment-gap condition, not just a stream-corruption fix. Review cycle 1
found the `exit != 0` arm of that skip guard was unreachable under bats'
`errexit` semantics; fixed in cycle 2 (see Review Convergence below) by
wrapping the capture in `set +e`/`set -e`, matching the existing AC-004
pattern in the same file.

### 2. `test(ci): treat superseded stories as terminal in sprint-state PC4 gate`
Extends the PC4 completeness/fidelity exclusion in
`plugins/vsdd-factory/tests/sprint-state-format.bats` to also treat
`**SUPERSEDED**` stories as terminal — the same class of exclusion already
applied to `retired` stories. Without this, superseded stories were incorrectly
counted against PC4 completeness/fidelity, producing false-fail gate blocks.

### 3. `fix(security): bump h2 0.4.14 → 0.4.19 — clear RUSTSEC-2026-0258 DoS`
Cargo.lock-only bump of `h2` to clear **RUSTSEC-2026-0258** (unbounded empty
DATA frames DoS), reachable in the production `factory-dispatcher` binary via
[...]

PR #783, opened 2026-08-25
fix: eliminate flaky async e2e timing failure blocking rc.24 (test-only)

Zious11:
# fix: eliminate flaky async e2e timing failure blocking rc.24 (test-only)

**Type:** Fix PR (test-harness timing tolerance, no product logic changed)
**Target release:** v1.0.0-rc.24 (final blocker)
**Branch:** `fix/flaky-async-e2e-timing` → `develop`

Eliminates a confirmed-flaky async e2e test that failed the post-merge
`develop` push-run (`bats-full-suite` linux job) after PR #782 merged
(579d75f3), even though the release code itself is proven green — the same
CI run's `cargo-host` jobs (ubuntu + macos), running
`cargo test --workspace --all-targets` (which includes these exact async e2e
tests), **passed**. Only the bats path failed, on fixed 15s deadlines during
debug-WASM cold start under load, roughly 2200 lines into `run-all.sh`. This
is a proven timing flake, not a functional regression — the fix widens
timing tolerances and narrows one bats test's scope to what it actually
intends to cover. No `crates/` production/dispatcher logic is touched.

---

### 1. `crates/factory-dispatcher/tests/full_stack_plugin_invocation.rs`
Widened async e2e timing ceilings:
- TC-5 `plugin.completed` wait: 15s → 60s
- TC-5 crash fallback wait: 2s → 10s
- TC-6 JoinHandle `tokio::time::timeout`: 15s → 60s

`wait_for_log_event` was already event-driven (100ms poll interval) — only
the *tolerance* for slow-but-correct completion changed. Assertion semantics
are unchanged: the tests still assert the same events occur, only the
deadline before treating a still-completing event as a failure is relaxed to
account for debug-WASM cold-start latency under concurrent CI load.

### 2. `plugins/vsdd-factory/tests/host-abi-hygiene.bats`
Narrowed T-012 from `cargo test -p factory-dispatcher --all-targets` (which
re-ran the same flaky async e2e suite that the `cargo-host` CI job already
covers independently in `ci.yml`) to `cargo test -p factory-dispatcher --lib
-- s19_09` — the 6 unit tests matching T-012's documented intent (D19
`read_prefix` + D22 timestamp behavior). This removes a redundant
debug-WASM cold-start path from the bats suite in `ci.yml` without losing
coverage there. Added a sanity gate asserting exactly 6 tests are selected
(strengthened in review cycle 1 to also assert each test's *identity*, not
just the count — see Review Convergence below), so future filter drift
fails loudly instead of silently changing scope.

### 3. `.github/workflows/release.yml` (added in review cycle 1, fixing F-001)
`release.yml`'s `validate` job has no independent `cargo test --workspace
--all-targets` execution step of its own — its "Pre-build factory-dispatcher
test binaries" step ran `--no-run` (compile only), and the *only* step that
actually executed test assertions was `./run-all.sh`, which drives T-012.
Narrowing T-012 (commit 2 above) therefore silently dropped the release-tag
validation gate's coverage from 241 lib tests + all integration binaries
down to 6 tests, with no compensating execution path the way `ci.yml`'s
separate `cargo-host` job provides. Fixed by converting the "Pre-build"
step into a real execution step (`cargo test --workspace --all-targets`,
renamed to reflect what it now does) — this is a strict superset of
`-- s19_09`, so it still warms `target/` for T-012 as a side effect,
avoiding a second, duplicate compile. The step's comment was rewritten to
describe current T-012 behavior and explain why `release.yml` now needs
this independent step. This fix is flake-safe: it only became possible
because commit 1 already widened the async e2e ceilings, so restoring full
execution in `release.yml` does not reintroduce the original flake — if
anything it makes `release.yml` strictly more robust than before this PR.

---

## Architecture Changes

N/A — no architecture/subsystem changes. Diff (after review cycle 1): one
Rust test file (timing constants), one `.bats` test file (test selector +
count/identity sanity gates), and one CI workflow file (`release.yml`
test-execution step). No `ARCH-INDEX.md` subsystem, no ADR is affected.

## Story Dependencies

N/A — not a story PR. Release-prep CI-stability fix with no `depends_on`
entry in `STORY-INDEX.md`. Depends only on PR #782 (579d75f3, already merged
to `develop`) as its base.

## Spec Traceability

N/A — no BC/AC/VP implemented or amended. Traceability for the underlying
defect is the CI gate itself:
- `bats-full-suite` (linux) push-run on `develop`, post-#782 merge —
  T-012 in `host-abi-hygiene.bats` failed on fixed 15s deadlines in
  `full_stack_plugin_invocation.rs` TC-5/TC-6 during debug-WASM cold start
  under load.
- Same CI run's `cargo-host` (ubuntu + macos) jobs, running the identical
  tests via `cargo test --workspace --all-targets`, **passed** — confirming
  the failure is a bats-path timing artifact, not a functional defect.
- Review cycle 1 additionally traced `release.yml`'s `validate` job as a
  second consumer of T-012 that needed its own independent execution path
  restored (see commit 3 above and Review Convergence below).

## Test Evidence

| Verification | Result |
|---------------|--------|
| `full_stack_plugin_invocation.rs` full-file e2e run, CPU-saturating load (16 cores pinned) | 3/3 PASS |
| `full_stack_plugin_invocation.rs` individual test runs | 5/5 PASS |
| `host-abi-hygiene.bats` T-012 (narrowed selector + count/identity gates) | 9/9 PASS, <1s (was multi-second debug-WASM cold start) |
| `release.yml` YAML validity (`yaml.safe_load`) | valid |
| `cargo fmt --check --all` | clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |

No new tests were added — this is a timing-tolerance widening plus a
scope-narrowing of an existing bats test to its documented intent (with
count + per-test-identity sanity gates against selector drift), plus a
`release.yml` fix restoring full-suite execution on the release-tag path.
Full `cargo test --workspace --all-targets` (now via `ci.yml`'s `cargo-host`
job AND `release.yml`'s own execution step) + `run-all.sh` are run by CI on
this PR and, most importantly, by the authoritative post-merge `develop`
push-run (the same job that surfaced the original flake).

## Demo Evidence

N/A — no user-observable behavior changed (test-timing tolerance +
test-selector narrowing only, per the `fix-pr-delivery` skill's
behavior-changing-fix criteria: no output, error message, CLI flag, API
response, or security-restriction change).

---

## Risk Assessment

- **Blast radius:** Test harness + release CI workflow only — one Rust
  `#[tokio::test]` file (timing constants), one `.bats` file (test selector
  + sanity gates), and one workflow YAML file (`release.yml` test-execution
  step). No `crates/` production/dispatcher source changed.
- **User impact:** None. The dispatcher binary's runtime behavior is
  unchanged; this only affects how long CI is willing to wait for
  already-correct async completion before declaring a timeout, and how much
  of the test suite actually executes on the release-tag path (increased,
  not decreased, by this PR after the cycle-1 fix).
- **Risk level:** LOW.

## Security Review

**Verdict: APPROVE** (independent pass by `security-reviewer`, diffed
`579d75f3f1` → `1301ea01b` directly rather than relying on the PR
description).

### Verification performed
- Confirmed the diff stat matches exactly (2 files, 64 insertions / 12
  deletions) — no `crates/*/src/` production code, no `Cargo.toml`/
  `Cargo.lock` dependency change, no auth/authz logic, no workflow file
  touched.
- **Coverage-drop check:** confirmed `.github/workflows/ci.yml`'s
  `cargo-host` job runs `cargo test --workspace --all-targets`, which
  independently exercises the full `full_stack_plugin_invocation.rs` suite
  (including the widened-timeout tests) — narrowing bats T-012 does not
  silently drop CI coverage, it removes a duplicate execution path.
- **Sanity-gate accuracy check:** grepped `s19_09` test names in
  `crates/factory-dispatcher/src/` and confirmed exactly 6 matches, matching
  the new gate's assertion verbatim — not a hallucinated count.
- **Injection check:** the new bats selector and sanity-gate `grep` use only
  static hardcoded strings — no interpolation of untrusted/external data.
- **Timing-widening check:** confirmed the widened waits only extend
  tolerance for slow-but-correct completion; no assertion is weakened or
  disabled, and there is no production-reachable resource-exhaustion vector
  (test-harness-scoped, bounded by existing CI workflow-level timeouts).

### Findings

| ID | Title | Severity | CWE | Disposition |
|----|-------|----------|-----|-------------|
| SEC-001 | Widened CI wall-clock timeout tolerance (15s→60s / 2s→10s) | LOW / informational | N/A — no applicable CWE, no attacker-reachable path | Non-blocking. Worst case: CI job wall-clock increases up to ~55s if the flake condition recurs at the new bound. Cost/CI-throughput consideration for devops-engineer, not a security gate. |

No CRITICAL, HIGH, or MEDIUM findings.

## Holdout Evaluation

N/A — test/CI-only change; no product behavior evaluated at wave gate.

## Adversarial Review

N/A — evaluated at Phase 5 for story-level product changes; not applicable
to this CI-stability fix PR.

---

## Review Convergence

Tracked in
[`.factory/code-delivery/fix-flaky-async-e2e-timing/review-findings.md`](./review-findings.md).
Note: the PR author and authenticated `gh` identity are the same account —
GitHub API rejects self-review-*state* changes (`APPROVE`/`REQUEST_CHANGES`
as a formal review state), so pr-reviewer posts its verdict via
`gh pr review --comment` (or a plain PR comment) instead; that comment is
treated as the approval of record.

| Cycle | Verdict | Blocking | Suggestions | Nits | Disposition |
|-------|---------|----------|-------------|------|--------------|
| 1 | REQUEST_CHANGES | F-001: `release.yml` unswept sibling site (TD-VSDD-060) — no independent `cargo test --workspace --all-targets` execution step, so T-012's narrowing silently dropped release-tag validate-gate coverage from 241 tests to 6 | F-002: hoist TC-4/7/9's existing timing constants alongside TC-5/6's widened ones into a shared named constant · F-003: strengthen the T-012 sanity gate to assert test *identity*, not just count | F-004: `timeout 120` over-provisioned for the narrowed (0.02s) run | F-001 (blocking) + F-003 (cheap, strictly stronger) routed to devops-engineer, fixed in commit `d39c62e0` — converted `release.yml`'s compile-only prebuild step into a real `cargo test --workspace --all-targets` execution step (flake-safe now that commit 1 widened the async ceilings) and added per-test-identity assertions to T-012's sanity gate. F-002 and F-004 left as-is (non-blocking, out of the minimal-fix scope for this cycle) |
| 2 | **APPROVE** | none | N-001: `release.yml`'s new step omitted `CI_REQUIRE_ARTIFACTS`/`VSDD_CORPUS_ROOT` env vars present in `ci.yml`'s equivalent step (same defect class as F-001, smaller magnitude — not escalated to blocking since coverage is preserved today) | N-002: `validate` job lacks toolchain pin + rust-cache (follow-up, not this PR's scope) · N-003: release gate now executes the de-flaked e2e suite for real (observation, confirmed correct) | Independently re-verified F-001/F-003 with load-bearing evidence (re-ran tests locally incl. a mutation control on the T-012 identity gate, confirmed zero duplicate compilation empirically, confirmed byte-identical command parity with `ci.yml`). N-001 folded in — commit `6bbdb507` adds the matching env block. F-004 **withdrawn** by reviewer (the 120s ceiling does real work under `ci.yml`'s `bats-full-suite` job, which never warms the debug lib-test target the way `release.yml` now does) |

---

## Pre-Merge Checklist

- [ ] All CI status checks passing on the PR
- [ ] pr-reviewer convergence to 0 blocking findings (posted via `--comment`,
      self-approve restriction noted above)
- [ ] Post-merge `develop` push-run `ci` (mounts `factory-artifacts` fresh,
      runs the full bats suite including the previously-flaky
      `host-abi-hygiene.bats` T-012 path) confirmed GREEN — this is the
      authoritative proof the release path is clear for rc.24

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,6 +57,9 @@ jobs:
         # separate compile-only pre-build step is no longer needed — running
         # the full suite once here, before ./run-all.sh, avoids paying to
         # compile factory-dispatcher's test targets twice.
+        env:
+          CI_REQUIRE_ARTIFACTS: "1"
+          VSDD_CORPUS_ROOT: ${{ github.workspace }}/.factory
         run: cargo test --workspace --all-targets
 
       - name: Run test suite
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,15 +31,33 @@ jobs:
           sudo wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
           sudo chmod +x /usr/local/bin/yq
 
-      - name: Pre-build factory-dispatcher test binaries
-        # T-012 in host-abi-hygiene.bats runs `timeout 120 cargo test -p
-        # factory-dispatcher --all-targets`. On a cold runner without a Rust
-        # cargo cache, compilation alone can exceed the 120-second bats timeout
-        # (exit code 124, rc.23 release run 29656342082). Pre-building here
-        # warms target/ so the bats test only executes the already-compiled
-        # binary within its window. --no-run = compile only; the actual test
-        # assertions still run inside the bats 120-second timeout as intended.
-        run: cargo test -p factory-dispatcher --all-targets --no-run
+      - name: Run full workspace test suite (pre-warms target/ for bats T-012)
+        # Historically this step was "Pre-build factory-dispatcher test
+        # binaries" — a `cargo test -p factory-dispatcher --all-targets
+        # --no-run` compile-only warm-up, because T-012 in host-abi-hygiene.bats
+        # used to run the equivalent test command for real and could blow its
+        # 120-second bats timeout on a cold runner (exit code 124, rc.23
+        # release run 29656342082).
+        #
+        # PR #783 narrowed T-012 to `cargo test -p factory-dispatcher --lib --
+        # s19_09` (6 unit tests) to eliminate an async e2e wall-clock flake in
+        # full_stack_plugin_invocation.rs — see the T-012 comment in
+        # host-abi-hygiene.bats for the full rationale. In ci.yml that
+        # narrowing doesn't drop coverage: the separate `cargo-host` job runs
+        # `cargo test --workspace --all-targets` independently of the bats
+        # path. release.yml has no equivalent independent job — the bats run
+        # below was, until now, the ONLY step here that executed test
+        # assertions, and it no longer covers the full suite.
+        #
+        # This step is release.yml's replacement for that lost coverage: it
+        # actually EXECUTES `cargo test --workspace --all-targets` (not
+        # --no-run), restoring release-gate parity with ci.yml's cargo-host
+        # job. It still incidentally warms target/ ahead of bats' T-012 (a
+        # strict superset of `-p factory-dispatcher --lib -- s19_09`), so a
+        # separate compile-only pre-build step is no longer needed — running
+        # the full suite once here, before ./run-all.sh, avoids paying to
+        # compile factory-dispatcher's test targets twice.
+        run: cargo test --workspace --all-targets
 
       - name: Run test suite
         run: cd plugins/vsdd-factory/tests && ./run-all.sh
--- a/crates/factory-dispatcher/tests/full_stack_plugin_invocation.rs
+++ b/crates/factory-dispatcher/tests/full_stack_plugin_invocation.rs
@@ -709,23 +709,29 @@ async fn test_e2e_BC_1_14_001_async_hook_output_reaches_sink_when_fast() {
     // crash-equivalent terminal events (plugin.crashed) are also acceptable per TC-5's
     // environment-tolerance requirement (factory-health binary may be absent in CI).
     //
-    // 15s bound covers debug WASM cold-start + WASM compile time in parallel test runs.
+    // 60s bound (widened from 15s per rc.24 flake fix) covers debug WASM cold-start +
+    // WASM compile time when the CI runner is under heavy parallel-test-run load.
+    // wait_for_log_event polls the log file, so this only lengthens how long a slow-but-
+    // correct completion is tolerated — it does not weaken the assertion: the test still
+    // fails if the event never arrives.
     let completed = wait_for_log_event(
         &log_dir,
         "plugin.completed",
         Some("session-start-telemetry"),
-        Duration::from_secs(15),
+        Duration::from_secs(60),
     )
     .await;
 
     // Also accept plugin.crashed as a terminal event (session-start-telemetry may crash
     // if exec_subprocess / factory-health is not on PATH; crash ≠ stub, real WASM ran).
+    // Fallback bound widened from 2s to 10s (rc.24 flake fix) for the same cold-start
+    // reason as the primary wait above.
     let crashed = if !completed {
         wait_for_log_event(
             &log_dir,
             "plugin.crashed",
             Some("session-start-telemetry"),
-            Duration::from_secs(2),
+            Duration::from_secs(10),
         )
         .await
     } else {
@@ -829,7 +835,11 @@ async fn test_e2e_BC_1_14_001_async_block_verdict_discarded() {
         Arc::new(ResolverRegistry::new()),
     );
 
-    let outcome = tokio::time::timeout(Duration::from_secs(15), handle)
+    // 60s bound (widened from 15s per rc.24 flake fix) covers debug WASM cold-start when
+    // the CI runner is under heavy parallel-test-run load. The JoinHandle either resolves
+    // (and the assertions below still run) or the timeout still fails the test — this only
+    // stops a slow-but-correct completion from being flagged as a failure.
+    let outcome = tokio::time::timeout(Duration::from_secs(60), handle)
         .await
         .expect("join did not panic")
         .expect("JoinHandle ok");
--- a/plugins/vsdd-factory/tests/host-abi-hygiene.bats
+++ b/plugins/vsdd-factory/tests/host-abi-hygiene.bats
@@ -516,4 +516,31 @@ _scan_bare_literals() {
     echo "T-012: filter scope drifted — update either the filter or this count."
     false
   fi
+
+  # Identity gate: a count-only check ("running 6 tests") would still pass
+  # if a filter typo swapped in six DIFFERENT tests that happen to also
+  # match "s19_09" (e.g. a future rename collision). Assert each of the six
+  # documented test names individually appears in the output, not just the
+  # aggregate count.
+  local expected_tests=(
+    "t001_s19_09_read_prefix_instantiates_without_link_error_via_production_linker"
+    "t002_s19_09_read_prefix_round_trip_bytes_correct_and_out_ptr_nonzero_via_production_path"
+    "t002b_s19_09_read_prefix_head_c_bound_clamps_out_len_to_max_bytes"
+    "t003_s19_09_read_prefix_capability_absent_returns_capability_denied_via_production_path"
+    "t015_s19_09_read_prefix_empty_file_returns_ok_with_zero_ptr_len_no_grow"
+    "test_s19_09_t013_emit_plugin_completed_async_has_timestamp_field"
+  )
+  local missing_tests=()
+  for t in "${expected_tests[@]}"; do
+    echo "$output" | grep -qF "$t" || missing_tests+=("$t")
+  done
+  if [ "${#missing_tests[@]}" -gt 0 ]; then
+    echo "FAIL: the 's19_09' filter matched 6 tests by count, but by NAME the"
+    echo "following documented test(s) are missing from the output (a wrong-test"
+    echo "filter-typo scenario, not just a wrong-count scenario):"
+    printf '  missing: %s\n' "${missing_tests[@]}"
+    echo "--- actual test names in output ---"
+    echo "$output" | grep -E "^test .* \.\.\. (ok|FAILED)" || true
+    false
+  fi
 }
--- a/plugins/vsdd-factory/tests/host-abi-hygiene.bats
+++ b/plugins/vsdd-factory/tests/host-abi-hygiene.bats
@@ -21,7 +21,9 @@
 #            "plugin.completed"/"plugin.timeout" literals in production
 #            code (before #[cfg(test)]) of read_file.rs, read_prefix.rs,
 #            emit_event.rs (extended to four classes per story v1.1/AC-008)
-#     T-012: cargo test -p factory-dispatcher exits 0 (bidirectional regression gate)
+#     T-012: cargo test -p factory-dispatcher --lib -- s19_09 exits 0 (bidirectional
+#            regression gate, scoped to the D19/D22 read_prefix + timestamp unit tests;
+#            narrowed from --all-targets per rc.24 flake fix — see T-012 comment below)
 #
 # RED gate status (pre-D20/D21 at develop 9787c056):
 #   T-004: FAILS — "enforced in S-1.5 via epoch interruption" present in read_file.rs
@@ -434,7 +436,36 @@ _scan_bare_literals() {
 # ---------------------------------------------------------------------------
 # T-012  AC-008/AC-009 bidirectional regression gate
 #
-# Gate: `cargo test -p factory-dispatcher --all-targets` exits 0.
+# Gate: `cargo test -p factory-dispatcher --lib -- s19_09` exits 0.
+#
+# T-012's actual intent (per the original RED/GREEN narrative below) is
+# narrower than "the entire factory-dispatcher test suite": it verifies that
+# the D19 (read_prefix wired into setup_host_on_store_data, the production
+# linker) and D22 (timestamp field on plugin.completed) unit tests pass, and
+# that the D21 literal→named-constant sweep did not change the event-type
+# string *values* those tests assert on. All six tests carrying that intent
+# live as `#[cfg(test)]` unit tests inside src/ (invoke.rs, emit_event.rs) and
+# share the "s19_09" name fragment:
+#   t001_s19_09_read_prefix_instantiates_without_link_error_via_production_linker
+#   t002_s19_09_read_prefix_round_trip_bytes_correct_and_out_ptr_nonzero_via_production_path
+#   t002b_s19_09_read_prefix_head_c_bound_clamps_out_len_to_max_bytes
+#   t003_s19_09_read_prefix_capability_absent_returns_capability_denied_via_production_path
+#   t015_s19_09_read_prefix_empty_file_returns_ok_with_zero_ptr_len_no_grow
+#   test_s19_09_t013_emit_plugin_completed_async_has_timestamp_field
+#
+# Narrowed from `cargo test -p factory-dispatcher --all-targets` (rc.24 flake
+# fix): `--all-targets` re-runs every integration-test binary in
+# crates/factory-dispatcher/tests/, including the BC-1.14.001 async e2e tests
+# in full_stack_plugin_invocation.rs. Those tests spawn real debug-build WASM
+# plugins and wait on wall-clock deadlines; ~2200 lines into this bats suite
+# the CI runner is under enough load that those deadlines occasionally blow
+# out (confirmed flake — the SAME tests pass under the `cargo-host` CI job's
+# `cargo test --workspace --all-targets`, which duplicates this gate's
+# coverage anyway). T-012 does not need that coverage to prove its own
+# intent, and re-paying debug-WASM cold-start here only adds risk with no
+# unique signal. `--lib` runs unit tests only (no tests/ integration
+# binaries, no WASM-plugin-file cold start) and `-- s19_09` scopes to the
+# tests the RED/GREEN narrative below actually documents.
 #
 # RED today: T-001/T-002/T-003 (invoke.rs production-linker) fail at runtime
 # because read_prefix is absent from setup_host_on_store_data (D19 unimplemented);
@@ -442,13 +473,13 @@ _scan_bare_literals() {
 # plugin.completed (D22 unimplemented).  T-009/T-010 are now compile-safe
 # bats grep-gates and no longer cause compile errors here.
 #
-# GREEN after D19+D20+D21+D22: T-001..T-003 pass (read_prefix wired), T-013
-# passes (timestamp field present), all existing tests still pass; exits 0.
+# GREEN after D19+D20+D21+D22: T-001..T-003 (+T-002b/T-015) pass (read_prefix
+# wired), T-013 passes (timestamp field present); exits 0.
 #
 # Bidirectional: RED pre-fix (runtime failures in T-001/T-002/T-003/T-013);
-# GREEN post-fix (entire factory-dispatcher test suite passes).
+# GREEN post-fix (the D19/D22 unit test set above passes).
 # ---------------------------------------------------------------------------
-@test "T-012 AC-008 regression: cargo test -p factory-dispatcher passes after D21 sweep" {
+@test "T-012 AC-008 regression: cargo test -p factory-dispatcher --lib -- s19_09 passes after D21 sweep" {
   REPO_ROOT="$(cd "${BATS_TEST_DIRNAME}/../../.." && pwd)"
 
   # Select platform-compatible timeout command.
@@ -463,15 +494,26 @@ _scan_bare_literals() {
   local exit_code=0
   cd "$REPO_ROOT"
   # shellcheck disable=SC2086
-  output=$(${timeout_cmd} cargo test -p factory-dispatcher --all-targets 2>&1) \
+  output=$(${timeout_cmd} cargo test -p factory-dispatcher --lib -- s19_09 2>&1) \
     || exit_code=$?
 
   if [ "$exit_code" -ne 0 ]; then
-    echo "FAIL: cargo test -p factory-dispatcher exited $exit_code"
+    echo "FAIL: cargo test -p factory-dispatcher --lib -- s19_09 exited $exit_code"
     echo "--- output (last 40 lines) ---"
     echo "$output" | tail -40
     echo "---"
     echo "T-012: existing test assertions on event type strings must pass after D21 sweep."
+    false
+  fi
+
+  # Sanity gate: the name filter must actually select the six documented
+  # tests (not zero tests via a silently-broken filter, and not fewer than
+  # expected via a partial name collision). "running N tests" is libtest's
+  # own summary line.
+  if ! echo "$output" | grep -qE "running 6 tests?"; then
+    echo "FAIL: expected the 's19_09' filter to select exactly 6 unit tests; got:"
+    echo "$output" | grep -E "running [0-9]+ tests?" || echo "(no 'running N tests' line found)"
+    echo "T-012: filter scope drifted — update either the filter or this count."
     false
   fi
 }


Zious11 (commented):
## pr-reviewer verdict: REQUEST_CHANGES

*(Posted as a comment-type review — the PR author and the authenticated `gh` identity are the same account, so GitHub rejects a formal `APPROVE`/`REQUEST_CHANGES` review-state change. This comment is the review of record.)*

Fresh-eyes review of `579d75f3f1` → `1301ea01b`, read from the diff directly (`git diff 579d75f3...1301ea01`) rather than from the PR description.

**The timing fix itself is correct and I would approve it as-is.** One blocking finding is about a sibling site the diff did not sweep, plus an affirmative claim in the PR body that is factually wrong for one of the two workflows that consume the changed test.

---

### What I verified independently

| Check | Method | Result |
|---|---|---|
| 6-test count in the new sanity gate is real, not hallucinated | Ran `cargo test -p factory-dispatcher --lib -- s19_09` locally | PASS — `running 6 tests`, all 6 green in 0.02s, names match the comment block verbatim (`t001/t002/t002b/t003/t015_s19_09_*`, `test_s19_09_t013_*`), `235 filtered out` |
| Sanity-gate regex is not accidentally loose | Traced `grep -qE "running 6 tests?"` against `running 16 tests` / `running 60 tests` | PASS — neither matches; `--lib` guarantees exactly one libtest summary line |
| Zero-test escape hatch is closed | Read gate placement | PASS — gate runs after the exit-code check, so a silently-broken filter (`running 0 tests`, cargo exit 0) fails loudly. Good design. |
| Assertion semantics unchanged by the widening | Read TC-5 / TC-6 assert blocks at head | PASS — no assertion weakened, removed, or made conditional; only the deadline moved |
| `--lib` actually contains the 6 tests | `crates/factory-dispatcher/Cargo.toml` has `[lib] name = "factory_dispatcher"`; tests live in `src/invoke.rs` + `src/host/emit_event.rs` | PASS |
| Coverage-drop check for `ci.yml` | `ci.yml` `cargo-host` (ubuntu + macos) runs `cargo test --workspace --all-targets`; triggers are `push: [main, develop]` and `pull_request: [main, develop]` | PASS — the narrowing is a genuine de-duplication *here* |
| Coverage-drop check for `release.yml` | `.github/workflows/release.yml` | **FAIL — see F-001** |
| Diff coherence, size, commit quality | 2 files / 64+ / 12-, single commit, `fix(test):` conventional, no AI attribution, no unrelated hunks | PASS |

---

## Findings

### [BLOCKING] F-001 — `release.yml` is an unswept sibling site; the "no coverage drop" claim does not hold there

**Category:** missing / coherence
**File:** `.github/workflows/release.yml` (lines 33–45), consequence of `plugins/vsdd-factory/tests/host-abi-hygiene.bats:497`

`release.yml`'s `validate` job has a step that exists *solely* to serve T-012's old invocation:

```yaml
- name: Pre-build factory-dispatcher test binaries
  # T-012 in host-abi-hygiene.bats runs `timeout 120 cargo test -p
  # factory-dispatcher --all-targets`. On a cold runner without a Rust
  # cargo cache, compilation alone can exceed the 120-second bats timeout
  # (exit code 124, rc.23 release run 29656342082). Pre-building here
  # warms target/ so the bats test only executes the already-compiled
  # binary within its window.
  run: cargo test -p factory-dispatcher --all-targets --no-run
```

Two problems, one cosmetic and one substantive:

1. **The comment now describes a command that no longer exists.** This is exactly the TD-VSDD-060 sibling-site-sweep case named in `CLAUDE.md` — the T-012 invocation was changed without grepping its callsites. `grep -rn -- "--all-targets" .github/workflows/` finds this in one hop.

2. **The substantive one:** `release.yml` has **no `cargo test --workspace --all-targets` step**. Its only *executing* test step is `run-all.sh`. So on the release-tag path, T-012 was the thing actually running the `factory-dispatcher` test binaries — the prebuild is `--no-run` (compile only). After this PR, `release.yml`'s `validate` job goes from executing **241 lib unit tests + every `crates/factory-dispatcher/tests/*` integration binary** down to **6 unit tests**. `validate` gates the whole release chain (`build-binaries` → `commit-binaries` → `release`), so this is a real gate, not decoration.

The PR body states:

> confirmed `.github/workflows/ci.yml`'s `cargo-host` job runs `cargo test --workspace --all-targets` … narrowing bats T-012 does not silently drop CI coverage, it removes a duplicate execution path.

That is true for `ci.yml` and false for `release.yml`. The coverage-drop check examined one of the two workflows that run `run-all.sh`.

**Mitigating context (why this is a defense-in-depth loss, not an unguarded hole):** the tag is cut at `main`'s tip after the release PR merges, and `ci.yml` triggers on both `pull_request: [main]` and `push: [main]`, so the tagged tree has already had `cargo test --workspace --all-targets` run on it. I am still calling this blocking rather than a suggestion because (a) the unswept comment is a named forbidden pattern in this repo's own `CLAUDE.md`, (b) the PR makes an affirmative correctness claim that is wrong, and (c) the fix is a few lines of YAML — squarely inside Canonical Principle Rule 4 (fix in scope, don't defer).

**Suggestion — pick one:**

- *Preferred (restores the coverage):* replace the now-vestigial prebuild with a real execution step, and update the comment:
  ```yaml
  - name: cargo test (factory-dispatcher, all targets)
    # run-all.sh's T-012 was narrowed to `--lib -- s19_09` (PR #783), so the
    # factory-dispatcher integration binaries are no longer executed by the bats
    # suite. release.yml has no cargo-host equivalent, so run them explicitly here.
    run: cargo test -p factory-dispatcher --all-targets
  ```
- *Minimum:* if you deliberately accept `ci.yml`'s push-to-`main` run as sufficient, then **delete** the prebuild step (it now warms a cache nothing uses) and say so in the PR body, replacing the "does not silently drop CI coverage" sentence with the accurate version: *"drops a duplicate execution path in `ci.yml`; in `release.yml` it drops execution entirely, accepted because `ci.yml` runs `cargo test --workspace --all-targets` on push to `main` before the tag is cut."*

Either way the stale comment must go.

---

### [SUGGESTION] F-002 — The widening fixes the two tests that failed, not the class

**Category:** coherence
**File:** `crates/factory-dispatcher/tests/full_stack_plugin_invocation.rs`

TC-5 and TC-6 went 15s → 60s. Three sibling waits in the same file, on the same debug-WASM cold start, on the same runners, were left alone:

- L649 — TC-4 `plugin.invoked`, `Duration::from_secs(30)`
- L1085 — TC-7 `plugin.invoked`, `Duration::from_secs(30)`
- L1300 / L1311 — TC-9 `plugin.timeout` 30s, crash fallback `Duration::from_secs(5)` (TC-5's equivalent fallback just went to 10s)

If the diagnosis is "fixed deadlines blow out when the runner is loaded," these three sit at half the new tolerance with the same failure mode. The narrowing in F-002's sibling change does reduce their exposure (they no longer run inside `run-all.sh` at all), so this is not urgent — but the underlying smell is five different magic numbers for one physical phenomenon.

**Suggestion:** hoist a single named constant and use it at every cold-start wait:
```rust
/// Wall-clock tolerance for debug-build WASM cold start + compile on a loaded
/// CI runner. Widened 15s→60s in rc.24 after a confirmed flake; see PR #783.
const WASM_COLD_START_BOUND: Duration = Duration::from_secs(60);
/// Headroom for a terminal event that follows an already-observed lifecycle event.
const TERMINAL_EVENT_FALLBACK: Duration = Duration::from_secs(10);
```
One place to tune, and the next flake in TC-4/7/9 gets fixed by the same edit instead of a fourth round of this PR. If 30s is genuinely sufficient for those three for a reason I can't see from the diff, a one-line comment saying why would close it.

---

### [SUGGESTION] F-003 — The sanity gate asserts cardinality but not identity

**Category:** coverage
**File:** `plugins/vsdd-factory/tests/host-abi-hygiene.bats:508–517`

The gate is a genuinely good addition — it's the difference between this narrowing being safe and being a silent-coverage-loss trap. But `running 6 tests` only pins the *count*. Add one `s19_09` test and delete another and the count stays 6 while the 6 names documented in the comment block above go stale, which is the drift the gate is there to catch.

`cargo test` already prints `test <path>::<name> ... ok` for each. Same amount of code, strictly stronger, and it also removes the (currently theoretical) risk that a future `--lib` removal produces multiple `running N tests` lines where `grep -q` passes on any one of them:

```bash
for t in t001_s19_09_read_prefix_instantiates \
         t002_s19_09_read_prefix_round_trip \
         t002b_s19_09_read_prefix_head_c_bound \
         t003_s19_09_read_prefix_capability_absent \
         t015_s19_09_read_prefix_empty_file \
         test_s19_09_t013_emit_plugin_completed_async; do
  echo "$output" | grep -qE "^test .*${t}.* \.\.\. ok$" || {
    echo "FAIL: expected T-012 test '${t}' in the s19_09 selection; it was not run."
    echo "T-012: filter scope drifted — update either the filter or this list."
    false
  }
done
```

---

### [NIT] F-004 — `timeout 120` is now heavily over-provisioned

**Category:** coherence
**File:** `plugins/vsdd-factory/tests/host-abi-hygiene.bats:497`

The 120s bats timeout was sized for `--all-targets` on a cold runner. The narrowed command runs in 0.02s once compiled. Harmless as-is, so genuinely a nit — but if F-001 is resolved by *deleting* the `release.yml` prebuild, the 120s window becomes the only thing absorbing a cold-cache compile of the lib target on the release runner, and it's worth a comment saying that's deliberate rather than leftover.

---

## Checklist summary

| # | Item | Result |
|---|---|---|
| 1 | Diff coherence | PASS — both hunks serve the stated flake fix; no unrelated changes |
| 2 | Description accuracy | **FAIL** — the coverage-drop claim is wrong for `release.yml` (F-001) |
| 3 | Test coverage | PASS with caveat — no new product code; the new sanity gate is the right instinct, hardenable per F-003 |
| 4 | Demo evidence | N/A, correctly justified — no user-observable behavior changed |
| 5 | Commit quality | PASS — `fix(test):` conventional, detailed body, no AI attribution |
| 6 | Diff size | PASS — 64+/12- across 2 files |
| 7 | Missing changes | **FAIL** — `release.yml` sibling site unswept (F-001, TD-VSDD-060) |
| 8 | Dependency status | PASS — base #782 / `579d75f3` already merged to `develop` |

**Verdict: REQUEST_CHANGES on F-001 only.** F-002/F-003/F-004 are non-blocking. The timing widening and the T-012 narrowing are both sound engineering and I have no objection to either; the blocker is the one workflow the sweep missed. Re-review should be quick.

Zious11 (commented):
_Duplicate submission — the pr-reviewer verdict of record is the review immediately above ([#pullrequestreview-5023625990](https://github.com/drbothen/vsdd-factory/pull/783#pullrequestreview-5023625990)). A submitted review cannot be deleted via the API, so this one is collapsed to a pointer. Verdict: **REQUEST_CHANGES**, 1 blocking finding (F-001)._

Zious11 (commented):
## PR Review — Cycle 2 (fresh-eyes, independent re-review of `1301ea01b..d39c62e0`)

**Verdict: APPROVE** ✅

I re-diffed the delta myself rather than relying on the fix summary, and I executed the changed test locally (including a mutation/negative control) rather than reasoning about it. Both cycle-1 items I asked for are genuinely closed, with load-bearing evidence. Details below, followed by three non-blocking observations and my disposition on F-002/F-004.

Delta reviewed: 2 files, +54/−9 — `.github/workflows/release.yml`, `plugins/vsdd-factory/tests/host-abi-hygiene.bats`. Single commit `d39c62e0`. No unrelated changes; diff is tightly scoped to the two findings.

---

### F-001 [BLOCKING → CLOSED] — `release.yml` full-suite execution restored

The old step (`cargo test -p factory-dispatcher --all-targets --no-run`) is replaced by a real execution step, `cargo test --workspace --all-targets`, in the `validate` job, positioned before `Run test suite` (`./run-all.sh`). Verified against each sub-question you asked:

**Right command.** `cargo test --workspace --all-targets` is byte-identical to the command in `ci.yml`'s `cargo-host` job, which is the coverage this narrowing dropped. Correct target for the parity claim.

**Right placement.** It sits after `Mount factory artifacts` (so the corpus tests that parent-walk to `.factory/` can find it — this is the same ordering constraint `ci.yml` documents as the F-S2107-P7-003 fix) and before the bats run. Both orderings are correct and non-accidental.

**No duplicate/wasted compilation — verified empirically, not asserted.** My real concern here was feature unification: `--workspace` unifies dependency features across all members, so a subsequent `-p factory-dispatcher --lib` invocation can resolve a *narrower* feature set, producing a different fingerprint and forcing a full rebuild of `tokio`/`wasmtime`/etc. That would silently defeat the warm-up that was this step's original reason to exist. I tested it on a clean fingerprint state:

```
$ cargo test --workspace --all-targets --no-run     # 20.6s wall
$ cargo test -p factory-dispatcher --lib -- s19_09
running 6 tests
test result: ok. 6 passed; 0 failed; 235 filtered out; finished in 0.07s
cargo test -p factory-dispatcher --lib -- s19_09  0.637 total
```

Zero `Compiling` lines, 0.64s end-to-end. The comment's claim that the workspace run is "a strict superset" and that a separate compile-only pre-build "is no longer needed" is **accurate**, including in the fingerprint sense — which is the non-obvious part. (Incidental confirmation of the F-001 magnitude: `6 passed; 235 filtered out` = the 241 lib tests T-012 used to cover.)

**Comment accuracy.** The rewritten comment correctly describes: the historical rationale, the rc.23 exit-124 incident, what PR #783 narrowed T-012 to, why the narrowing is coverage-neutral in `ci.yml` but not here, and why the new step replaces the old one. It does not overstate what was wrong before. One small nuance under "parity" — see N-001.

**Sibling sweep (TD-VSDD-060).** `git grep "Pre-build factory-dispatcher"` across the tree returns exactly one hit: the new comment's own historical reference. No test, doc, or workflow asserts on the old step name. Renaming the step breaks nothing.

**YAML validity.** `yaml.safe_load` on the new `release.yml` parses clean.

---

### F-003 [SUGGESTION → CLOSED] — T-012 identity gate

**All 6 names correct, no typos.** Each entry in `expected_tests` matches a real `fn` verbatim:

| Expected name | Source |
|---|---|
| `t001_s19_09_read_prefix_instantiates_without_link_error_via_production_linker` | `crates/factory-dispatcher/src/invoke.rs` |
| `t002_s19_09_read_prefix_round_trip_bytes_correct_and_out_ptr_nonzero_via_production_path` | `invoke.rs` |
| `t002b_s19_09_read_prefix_head_c_bound_clamps_out_len_to_max_bytes` | `invoke.rs` |
| `t003_s19_09_read_prefix_capability_absent_returns_capability_denied_via_production_path` | `invoke.rs` |
| `t015_s19_09_read_prefix_empty_file_returns_ok_with_zero_ptr_len_no_grow` | `invoke.rs` |
| `test_s19_09_t013_emit_plugin_completed_async_has_timestamp_field` | `src/host/emit_event.rs` |

The sixth name's different prefix shape (`test_s19_09_t013_…` vs `tNNN_s19_09_…`) is the one I'd most expect to be transcribed wrong; it's right.

**Still passes.** `bats host-abi-hygiene.bats` at `d39c62e0` — 9/9 ok, including T-012. Whole-file run, not just the changed test, so no collateral damage to the other hygiene gates (notably T-011's `_scan_bare_literals`, which only scans `.rs` and is unaffected by the new string literals).

**It is load-bearing, not a paper fix (TD-VSDD-059).** I ran a mutation control — flipped one character of case in `…no_grow` → `…no_grOW` on a scratch copy — and the gate fired correctly, with the count gate still passing:

```
not ok 1 T-012 …
# FAIL: the 's19_09' filter matched 6 tests by count, but by NAME the
# following documented test(s) are missing from the output …
#   missing: t015_s19_09_read_prefix_empty_file_returns_ok_with_zero_ptr_len_no_grOW
```

That is exactly the wrong-tests-right-count scenario F-003 was about, and the diagnostic block dumps the actual test names, which is what a future maintainer needs.

**Bash mechanics** — checked the portability traps, all clean: `${#missing_tests[@]}` on an empty array is safe under `set -u` (only bare `${arr[@]}` is not, and that expansion is guarded behind the `-gt 0` check); array `+=` is bash 3.1+, so macOS bash 3.2 is fine; `grep -qF` substring-matches libtest's `test <module>::<name> ... ok` lines correctly; `false` terminates under bats' errexit and also returns non-zero as the last statement either way.

---

### New issues introduced — none blocking

**N-001 [SUGGESTION] `release.yml`'s new step omits the env block that `ci.yml`'s `cargo-host` sets on the identical command.** `cargo-host` runs `cargo test --workspace --all-targets` with:

```yaml
env:
  CI_REQUIRE_ARTIFACTS: "1"
  VSDD_CORPUS_ROOT: ${{ github.workspace }}/.factory
```

`CI_REQUIRE_ARTIFACTS=1` is what converts the ~11 `validate-cross-site-correspondence` corpus tests from *silently skipping* to *hard-failing* when `.factory/` isn't discoverable. Without it, the release gate would report green with those tests skipped — which is a smaller-magnitude instance of the exact defect class F-001 was about, and it makes the comment's "restoring release-gate parity with ci.yml's cargo-host job" true of the command but not quite of the coverage semantics.

I am **not** escalating this to blocking, for a concrete reason: `.factory/` *is* mounted at the workspace root in this job, and the 8-level parent-walk from `CARGO_MANIFEST_DIR` finds it there, so those tests will actually execute — `ci.yml`'s own comment calls `VSDD_CORPUS_ROOT` "belt-and-braces explicit override." Real coverage is preserved today. The env block is cheap insurance against a future mount refactor, and a good follow-up.

**N-002 [NIT] The `validate` job has no toolchain pin and no cargo cache.** Every Rust-executing job in `ci.yml` (`cargo-host`, `bats-full-suite`) pairs `dtolnay/rust-toolchain@1.95.0` with `Swatinem/rust-cache`. `release.yml`'s `validate` has neither, so it now compiles *and runs* the entire workspace stone-cold on every tag. Correctness is unaffected (`rust-toolchain.toml` governs the channel, as it did before this change), and release is tag-triggered so the cost is rare — but the fix meaningfully amplifies this job's wall-clock, and it's adjacent to the TD #70 cargo-cache-reuse thread.

**N-003 [observation, no action]** The release gate now *executes* `full_stack_plugin_invocation.rs` — the async e2e test this PR exists to de-flake — where it previously only compiled it (`--no-run`). Worth stating out loud since it's a consequence of the fix rather than an accident. I'm satisfied it's correct: the PR's own diagnosis is that the flake is specific to running those tests ~2200 lines into the bats suite under load, and that they pass under a dedicated `cargo test --workspace --all-targets` step — which is precisely the context this new step provides, at exposure identical to what `cargo-host` already carries on every PR. Restoring the coverage is clearly better than the alternative.

---

### Disposition on F-002 and F-004

**F-002 (hoist TC-4/7/9 timing constants alongside TC-5/6's) — agreed, non-blocking. Not re-escalated.** It's a readability/maintainability improvement in test code, with no correctness or coverage consequence, and it's orthogonal to the flake fix. Reasonable to leave out of a minimal-fix scope; worth picking up next time that file is touched.

**F-004 (`timeout 120` over-provisioned) — I am withdrawing this finding.** I raised it in cycle 1 reasoning from `release.yml`, where the warm-up now makes T-012 a sub-second run. But T-012 also runs under `ci.yml`'s `bats-full-suite` job, which pre-builds only `--release` artifacts (`cargo build --release -p factory-dispatcher` plus release-profile WASM) and never warms the **debug** lib-test target. In that job T-012 may genuinely have to compile, so the 120s headroom is doing real work. It's a ceiling, not a wait — nothing pays it when the target is warm. Leaving it at 120 is the correct call, and my original NIT was wrong. Please disregard it rather than "deferring" it.

---

### Checklist

| # | Item | Result |
|---|---|---|
| 1 | Diff coherence | ✅ 2 files, both traceable to F-001/F-003; nothing unrelated |
| 2 | Description accuracy | ✅ comment rewrite matches actual behavior; parity claim verified (see N-001 nuance) |
| 3 | Test coverage | ✅ changed test executed locally, 9/9 in-file; new gate mutation-verified |
| 4 | Demo evidence | n/a — CI-config/test-harness fix, no user-facing AC |
| 5 | Commit quality | ✅ conventional format, cites PR #783 review F-001, single commit |
| 6 | Diff size | ✅ 54 insertions |
| 7 | Missing changes | ✅ both required fixes present; sibling sweep clean |
| 8 | Dependency status | ✅ no upstream PR dependencies |

Both cycle-1 findings are closed with verified, load-bearing evidence. The three new observations are all non-blocking and none should hold the merge. Approving — CI-wait and merge may proceed.

Zious11 (commented):
# pr-reviewer — PR #783 (`fix/flaky-async-e2e-timing` → `develop`)

**Current verdict (cycle 3, HEAD `6bbdb507`): APPROVE.** This is the approval of record; CI-wait and merge may proceed.

Review cycles are recorded newest-first. Cycles 2 and 1 are retained below for the finding history.

---

# Cycle 3 — verdict: APPROVE (confirmation pass on N-001 closure)

Scope: confirmation check on the delta `d39c62e0..6bbdb507` only — not a re-run of the full 8-item checklist, which was completed at cycle 2 and whose result is unchanged. The delta was diffed directly rather than taken from the fix summary.

### N-001 [SUGGESTION → CLOSED] — `release.yml` env parity with `ci.yml`'s `cargo-host`

| Field | Value |
|-------|-------|
| Severity | suggestion → **closed** |
| Category | coverage |
| File | `.github/workflows/release.yml` |

The `validate` job's step `Run full workspace test suite (pre-warms target/ for bats T-012)` now carries:

```yaml
        env:
          CI_REQUIRE_ARTIFACTS: "1"
          VSDD_CORPUS_ROOT: ${{ github.workspace }}/.factory
        run: cargo test --workspace --all-targets
```

Verified:

- **Key names and values are byte-identical** to `ci.yml`'s `cargo-host` → `cargo test (workspace, all targets)` step. The coverage-semantics half of the parity claim, which cycle 2 found true only of the command, now holds too.
- **Placement is correct.** `env:` sits at 8-space indentation, the same level as `run:`, immediately before it and inside the intended step — the intervening lines are that step's own comment block, not a step boundary. Confirmed structurally, not by eye: `yaml.safe_load` attaches the mapping `{'CI_REQUIRE_ARTIFACTS': '1', 'VSDD_CORPUS_ROOT': '${{ github.workspace }}/.factory'}` to the step whose `run` is exactly `cargo test --workspace --all-targets`.
- **YAML still valid** — parses clean; `validate` still resolves to 7 steps.
- **Nothing else changed.** `d39c62e0..6bbdb507` is a single commit; `git diff --stat` is `.github/workflows/release.yml | 3 +++` — 3 insertions, 0 deletions, no other files.
- **Commit quality** — `fix(ci): add CI_REQUIRE_ARTIFACTS/VSDD_CORPUS_ROOT env parity to release.yml test step (PR #783 cycle-2)`; conventional, cites the review finding, no AI attribution.

### Ordering check performed beyond the three confirmation items

`CI_REQUIRE_ARTIFACTS=1` converts the `validate-cross-site-correspondence` corpus tests from silently skipping to hard-failing when `.factory/` is undiscoverable, so adding it makes the step's success newly dependent on mount ordering — the same dependency `ci.yml` documents as its F-S2107-P7-003 fix. Confirmed safe here: `release.yml`'s `validate` job runs `Mount factory artifacts` (`git worktree add .factory origin/factory-artifacts`) at step 2, well before this test step at step 5, so `.factory/` exists at workspace root and `VSDD_CORPUS_ROOT` resolves. No new blocker.

### Carried forward unchanged

N-002 (no toolchain pin / cargo cache on `validate`, NIT, adjacent to TD #70) and N-003 (observation, no action) stand as recorded at cycle 2. No new findings.

---

# Cycle 2 — verdict: APPROVE

*(Posted as a comment-type review via `gh pr review 783 --comment --body-file` — the PR author and the authenticated `gh` identity are the same account, so GitHub rejects a formal `APPROVE`/`REQUEST_CHANGES` review-state change. Note this is `gh pr review`, NOT `gh pr comment`: it creates a review record with state `COMMENTED`, confirmed via `gh pr view 783 --json reviews` → `{"author":"Zious11","state":"COMMENTED"}`. The comment-type downgrade is forced by the GitHub self-review restriction, not a choice of mechanism.)*

Independent re-review of the delta `1301ea01b..d39c62e0` (1 commit, 2 files, +54/−9), diffed directly rather than taken from the fix summary. Both cycle-1 items are genuinely closed, verified by execution rather than by reading.

### F-001 [BLOCKING → CLOSED] — `release.yml` full-suite execution restored

| Field | Value |
|-------|-------|
| Severity | blocking → **closed** |
| Category | missing / coherence |
| File | `.github/workflows/release.yml` |

The compile-only `cargo test -p factory-dispatcher --all-targets --no-run` step is replaced by a real execution step, `cargo test --workspace --all-targets`, byte-identical to `ci.yml`'s `cargo-host` command — the coverage the T-012 narrowing dropped. Placement is correct: after `Mount factory artifacts` (the ordering the corpus tests' parent-walk needs, per ci.yml's F-S2107-P7-003 note) and before `./run-all.sh`.

**No duplicate/wasted compilation — verified empirically.** The non-obvious risk was feature unification: `--workspace` unifies dependency features, so a later `-p factory-dispatcher --lib` can resolve a *narrower* set, mismatch fingerprints, and force a full rebuild of `tokio`/`wasmtime` — silently defeating the warm-up that was this step's original purpose. Measured on a clean fingerprint state:

```
$ cargo test --workspace --all-targets --no-run     # 20.6s wall
$ cargo test -p factory-dispatcher --lib -- s19_09
running 6 tests
test result: ok. 6 passed; 0 failed; 235 filtered out; finished in 0.07s
cargo test -p factory-dispatcher --lib -- s19_09  0.637 total
```

Zero `Compiling` lines. The comment's "strict superset" / "separate compile-only pre-build no longer needed" claim holds in the fingerprint sense, not just the target-set sense. (`6 passed; 235 filtered out` also independently confirms the 241-test F-001 magnitude.)

**Sibling sweep (TD-VSDD-060):** `git grep "Pre-build factory-dispatcher"` returns exactly one hit — the new comment's own historical reference. No test, doc, or workflow asserts on the old step name. **YAML validity:** `yaml.safe_load` parses the new `release.yml` clean.

### F-003 [SUGGESTION → CLOSED] — T-012 identity gate

| Field | Value |
|-------|-------|
| Severity | suggestion → **closed** |
| Category | coverage |
| File | `plugins/vsdd-factory/tests/host-abi-hygiene.bats` |

All 6 entries in `expected_tests` match a real `fn` verbatim — 5 in `crates/factory-dispatcher/src/invoke.rs`, 1 (`test_s19_09_t013_emit_plugin_completed_async_has_timestamp_field`) in `src/host/emit_event.rs`. That sixth name's different prefix shape is the one most likely to be transcribed wrong; it is correct.

**Still passes:** full `bats host-abi-hygiene.bats` at `d39c62e0` → 9/9 ok (whole file, not just T-012, so no collateral damage to the other hygiene gates — notably T-011's `_scan_bare_literals`, which scans only `.rs` and is unaffected by the new string literals).

**Load-bearing, not a paper fix (TD-VSDD-059):** mutation control — one character case-flipped in a scratch copy (`..._no_grow` → `..._no_grOW`) — made the gate fire correctly *while the count gate still passed*, i.e. exactly the wrong-tests-right-count scenario F-003 targeted:

```
not ok 1 T-012 …
#   missing: t015_s19_09_read_prefix_empty_file_returns_ok_with_zero_ptr_len_no_grOW
```

**Bash mechanics:** `${#missing_tests[@]}` on an empty array is safe under `set -u` (only bare `${arr[@]}` is not, and that expansion is guarded behind the `-gt 0` check); array `+=` is bash 3.1+, so macOS bash 3.2 is fine; `grep -qF` substring-matches libtest's `test <module>::<name> ... ok` lines; `false` terminates under bats errexit and returns non-zero as last statement either way.

### New findings introduced by the fix — 3, none blocking

**N-001 [SUGGESTION] — `release.yml`'s new step omits `cargo-host`'s env block.**

| Field | Value |
|-------|-------|
| Severity | suggestion |
| Category | coverage |
| File | `.github/workflows/release.yml` |
| Finding | `ci.yml`'s `cargo-host` runs the identical command with `CI_REQUIRE_ARTIFACTS: "1"` and `VSDD_CORPUS_ROOT: ${{ github.workspace }}/.factory`. `CI_REQUIRE_ARTIFACTS=1` converts the ~11 `validate-cross-site-correspondence` corpus tests from silently skipping to hard-failing when `.factory/` is undiscoverable. Without it the release gate could report green with those tests skipped — a smaller-magnitude instance of the same defect class as F-001 — and the comment's "parity with ci.yml's cargo-host job" is true of the command but not of the coverage semantics. |
| Suggestion | Add the two-line `env:` block to the new step. |
| Why not blocking | `.factory/` *is* mounted at workspace root in this job and the 8-level parent-walk from `CARGO_MANIFEST_DIR` finds it, so the corpus tests execute today; ci.yml's own comment calls `VSDD_CORPUS_ROOT` "belt-and-braces explicit override." Real coverage is preserved. |

**N-002 [NIT] — `validate` job has no toolchain pin and no cargo cache.**

| Field | Value |
|-------|-------|
| Severity | nit |
| Category | size / cost |
| Finding | Every Rust-executing job in `ci.yml` (`cargo-host`, `bats-full-suite`) pairs `dtolnay/rust-toolchain@1.95.0` with `Swatinem/rust-cache`. `release.yml`'s `validate` has neither, so it now compiles *and runs* the whole workspace stone-cold on every tag. |
| Suggestion | Add both actions to `validate`; adjacent to the TD #70 cargo-cache-reuse thread. |
| Why not blocking | Correctness is unaffected (`rust-toolchain.toml` governs the channel, as before this change), and release is tag-triggered so the cost is rare. Pre-existing structure that this fix amplifies rather than introduces. |

**N-003 [observation, no action] — the release gate now executes `full_stack_plugin_invocation.rs`** where it previously only compiled it (`--no-run`). Stated explicitly because it is a consequence of the fix. Assessed as correct: the PR's own diagnosis is that the flake is specific to running those tests ~2200 lines into the bats suite under load and that they pass under a dedicated `cargo test --workspace --all-targets` step — precisely the context this new step provides, at exposure identical to what `cargo-host` already carries on every PR.

### Disposition on the remaining cycle-1 findings

- **F-002 (hoist TC-4/7/9 timing constants alongside TC-5/6's)** — agreed non-blocking, **not re-escalated**. Readability/maintainability in test code, no correctness or coverage consequence, orthogonal to the flake fix. Reasonable to leave out of minimal-fix scope; pick up next time the file is touched.
- **F-004 (`timeout 120` over-provisioned)** — **WITHDRAWN, not deferred.** Cycle 1 reasoned from `release.yml`, where the warm-up makes T-012 sub-second. But T-012 also runs under `ci.yml`'s `bats-full-suite`, which pre-builds only `--release` artifacts (`cargo build --release -p factory-dispatcher` plus release-profile WASM) and never warms the **debug** lib-test target — so there T-012 may genuinely have to compile and the 120s headroom is doing real work. It is a ceiling, not a wait; nothing pays it when the target is warm. Leaving it at 120 is correct and the original NIT was wrong. Drop it from the register rather than carrying it forward.

### Cycle-2 checklist

| # | Item | Result |
|---|------|--------|
| 1 | Diff coherence | PASS — 2 files, both traceable to F-001/F-003, nothing unrelated |
| 2 | Description accuracy | PASS — comment rewrite matches actual behavior; parity claim verified (N-001 nuance) |
| 3 | Test coverage | PASS — changed test executed locally, 9/9 in-file; new gate mutation-verified |
| 4 | Demo evidence | n/a — CI-config / test-harness fix, no user-facing AC |
| 5 | Commit quality | PASS — conventional `fix(ci):`, cites PR #783 review F-001, single commit, no AI attribution |
| 6 | Diff size | PASS — +54/−9 |
| 7 | Missing changes | PASS — both required fixes present; sibling sweep clean |
| 8 | Dependency status | PASS — no upstream PR dependencies |

---

# Cycle 1 — verdict: REQUEST_CHANGES

*(Posted as a comment-type review — the PR author and the authenticated `gh` identity are the same account, so GitHub rejects a formal `APPROVE`/`REQUEST_CHANGES` review-state change. This comment is the review of record.)*

Fresh-eyes review of `579d75f3f1` → `1301ea01b`, read from the diff directly (`git diff 579d75f3...1301ea01`) rather than from the PR description.

**The timing fix itself is correct and I would approve it as-is.** One blocking finding is about a sibling site the diff did not sweep, plus an affirmative claim in the PR body that is factually wrong for one of the two workflows that consume the changed test.

---

### What I verified independently

| Check | Method | Result |
|---|---|---|
| 6-test count in the new sanity gate is real, not hallucinated | Ran `cargo test -p factory-dispatcher --lib -- s19_09` locally | PASS — `running 6 tests`, all 6 green in 0.02s, names match the comment block verbatim (`t001/t002/t002b/t003/t015_s19_09_*`, `test_s19_09_t013_*`), `235 filtered out` |
| Sanity-gate regex is not accidentally loose | Traced `grep -qE "running 6 tests?"` against `running 16 tests` / `running 60 tests` | PASS — neither matches; `--lib` guarantees exactly one libtest summary line |
| Zero-test escape hatch is closed | Read gate placement | PASS — gate runs after the exit-code check, so a silently-broken filter (`running 0 tests`, cargo exit 0) fails loudly. Good design. |
| Assertion semantics unchanged by the widening | Read TC-5 / TC-6 assert blocks at head | PASS — no assertion weakened, removed, or made conditional; only the deadline moved |
| `--lib` actually contains the 6 tests | `crates/factory-dispatcher/Cargo.toml` has `[lib] name = "factory_dispatcher"`; tests live in `src/invoke.rs` + `src/host/emit_event.rs` | PASS |
| Coverage-drop check for `ci.yml` | `ci.yml` `cargo-host` (ubuntu + macos) runs `cargo test --workspace --all-targets`; triggers are `push: [main, develop]` and `pull_request: [main, develop]` | PASS — the narrowing is a genuine de-duplication *here* |
| Coverage-drop check for `release.yml` | `.github/workflows/release.yml` | **FAIL — see F-001** |
| Diff coherence, size, commit quality | 2 files / 64+ / 12-, single commit, `fix(test):` conventional, no AI attribution, no unrelated hunks | PASS |

---

## Findings

### [BLOCKING] F-001 — `release.yml` is an unswept sibling site; the "no coverage drop" claim does not hold there

**Category:** missing / coherence
**File:** `.github/workflows/release.yml` (lines 33–45), consequence of `plugins/vsdd-factory/tests/host-abi-hygiene.bats:497`

`release.yml`'s `validate` job has a step that exists *solely* to serve T-012's old invocation:

```yaml
- name: Pre-build factory-dispatcher test binaries
  # T-012 in host-abi-hygiene.bats runs `timeout 120 cargo test -p
  # factory-dispatcher --all-targets`. On a cold runner without a Rust
  # cargo cache, compilation alone can exceed the 120-second bats timeout
  # (exit code 124, rc.23 release run 29656342082). Pre-building here
  # warms target/ so the bats test only executes the already-compiled
  # binary within its window.
  run: cargo test -p factory-dispatcher --all-targets --no-run
```

Two problems, one cosmetic and one substantive:

1. **The comment now describes a command that no longer exists.** This is exactly the TD-VSDD-060 sibling-site-sweep case named in `CLAUDE.md` — the T-012 invocation was changed without grepping its callsites. `grep -rn -- "--all-targets" .github/workflows/` finds this in one hop.

2. **The substantive one:** `release.yml` has **no `cargo test --workspace --all-targets` step**. Its only *executing* test step is `run-all.sh`. So on the release-tag path, T-012 was the thing actually running the `factory-dispatcher` test binaries — the prebuild is `--no-run` (compile only). After this PR, `release.yml`'s `validate` job goes from executing **241 lib unit tests + every `crates/factory-dispatcher/tests/*` integration binary** down to **6 unit tests**. `validate` gates the whole release chain (`build-binaries` → `commit-binaries` → `release`), so this is a real gate, not decoration.

The PR body states:

> confirmed `.github/workflows/ci.yml`'s `cargo-host` job runs `cargo test --workspace --all-targets` … narrowing bats T-012 does not silently drop CI coverage, it removes a duplicate execution path.

That is true for `ci.yml` and false for `release.yml`. The coverage-drop check examined one of the two workflows that run `run-all.sh`.

**Mitigating context (why this is a defense-in-depth loss, not an unguarded hole):** the tag is cut at `main`'s tip after the release PR merges, and `ci.yml` triggers on both `pull_request: [main]` and `push: [main]`, so the tagged tree has already had `cargo test --workspace --all-targets` run on it. I am still calling this blocking rather than a suggestion because (a) the unswept comment is a named forbidden pattern in this repo's own `CLAUDE.md`, (b) the PR makes an affirmative correctness claim that is wrong, and (c) the fix is a few lines of YAML — squarely inside Canonical Principle Rule 4 (fix in scope, don't defer).

**Suggestion — pick one:**

- *Preferred (restores the coverage):* replace the now-vestigial prebuild with a real execution step, and update the comment:
  ```yaml
  - name: cargo test (factory-dispatcher, all targets)
    # run-all.sh's T-012 was narrowed to `--lib -- s19_09` (PR #783), so the
    # factory-dispatcher integration binaries are no longer executed by the bats
    # suite. release.yml has no cargo-host equivalent, so run them explicitly here.
    run: cargo test -p factory-dispatcher --all-targets
  ```
- *Minimum:* if you deliberately accept `ci.yml`'s push-to-`main` run as sufficient, then **delete** the prebuild step (it now warms a cache nothing uses) and say so in the PR body, replacing the "does not silently drop CI coverage" sentence with the accurate version: *"drops a duplicate execution path in `ci.yml`; in `release.yml` it drops execution entirely, accepted because `ci.yml` runs `cargo test --workspace --all-targets` on push to `main` before the tag is cut."*

Either way the stale comment must go.

---

### [SUGGESTION] F-002 — The widening fixes the two tests that failed, not the class

**Category:** coherence
**File:** `crates/factory-dispatcher/tests/full_stack_plugin_invocation.rs`

TC-5 and TC-6 went 15s → 60s. Three sibling waits in the same file, on the same debug-WASM cold start, on the same runners, were left alone:

- L649 — TC-4 `plugin.invoked`, `Duration::from_secs(30)`
- L1085 — TC-7 `plugin.invoked`, `Duration::from_secs(30)`
- L1300 / L1311 — TC-9 `plugin.timeout` 30s, crash fallback `Duration::from_secs(5)` (TC-5's equivalent fallback just went to 10s)

If the diagnosis is "fixed deadlines blow out when the runner is loaded," these three sit at half the new tolerance with the same failure mode. The narrowing in F-002's sibling change does reduce their exposure (they no longer run inside `run-all.sh` at all), so this is not urgent — but the underlying smell is five different magic numbers for one physical phenomenon.

**Suggestion:** hoist a single named constant and use it at every cold-start wait:
```rust
/// Wall-clock tolerance for debug-build WASM cold start + compile on a loaded
/// CI runner. Widened 15s→60s in rc.24 after a confirmed flake; see PR #783.
const WASM_COLD_START_BOUND: Duration = Duration::from_secs(60);
/// Headroom for a terminal event that follows an already-observed lifecycle event.
const TERMINAL_EVENT_FALLBACK: Duration = Duration::from_secs(10);
```
One place to tune, and the next flake in TC-4/7/9 gets fixed by the same edit instead of a fourth round of this PR. If 30s is genuinely sufficient for those three for a reason I can't see from the diff, a one-line comment saying why would close it.

---

### [SUGGESTION] F-003 — The sanity gate asserts cardinality but not identity

**Category:** coverage
**File:** `plugins/vsdd-factory/tests/host-abi-hygiene.bats:508–517`

The gate is a genuinely good addition — it's the difference between this narrowing being safe and being a silent-coverage-loss trap. But `running 6 tests` only pins the *count*. Add one `s19_09` test and delete another and the count stays 6 while the 6 names documented in the comment block above go stale, which is the drift the gate is there to catch.

`cargo test` already prints `test <path>::<name> ... ok` for each. Same amount of code, strictly stronger, and it also removes the (currently theoretical) risk that a future `--lib` removal produces multiple `running N tests` lines where `grep -q` passes on any one of them:

```bash
for t in t001_s19_09_read_prefix_instantiates \
         t002_s19_09_read_prefix_round_trip \
         t002b_s19_09_read_prefix_head_c_bound \
         t003_s19_09_read_prefix_capability_absent \
         t015_s19_09_read_prefix_empty_file \
         test_s19_09_t013_emit_plugin_completed_async; do
  echo "$output" | grep -qE "^test .*${t}.* \.\.\. ok$" || {
    echo "FAIL: expected T-012 test '${t}' in the s19_09 selection; it was not run."
    echo "T-012: filter scope drifted — update either the filter or this list."
    false
  }
done
```

---

### [NIT] F-004 — `timeout 120` is now heavily over-provisioned

**Category:** coherence
**File:** `plugins/vsdd-factory/tests/host-abi-hygiene.bats:497`

The 120s bats timeout was sized for `--all-targets` on a cold runner. The narrowed command runs in 0.02s once compiled. Harmless as-is, so genuinely a nit — but if F-001 is resolved by *deleting* the `release.yml` prebuild, the 120s window becomes the only thing absorbing a cold-cache compile of the lib target on the release runner, and it's worth a comment saying that's deliberate rather than leftover.

---

## Checklist summary

| # | Item | Result |
|---|---|---|
| 1 | Diff coherence | PASS — both hunks serve the stated flake fix; no unrelated changes |
| 2 | Description accuracy | **FAIL** — the coverage-drop claim is wrong for `release.yml` (F-001) |
| 3 | Test coverage | PASS with caveat — no new product code; the new sanity gate is the right instinct, hardenable per F-003 |
| 4 | Demo evidence | N/A, correctly justified — no user-observable behavior changed |
| 5 | Commit quality | PASS — `fix(test):` conventional, detailed body, no AI attribution |
| 6 | Diff size | PASS — 64+/12- across 2 files |
| 7 | Missing changes | **FAIL** — `release.yml` sibling site unswept (F-001, TD-VSDD-060) |
| 8 | Dependency status | PASS — base #782 / `579d75f3` already merged to `develop` |

**Verdict: REQUEST_CHANGES on F-001 only.** F-002/F-003/F-004 are non-blocking. The timing widening and the T-012 narrowing are both sound engineering and I have no objection to either; the blocker is the one workflow the sweep missed. Re-review should be quick.

PR #817, opened 2026-09-06
chore(config): register prd-supplement artifact path pattern

Zious11:
## What Changed

Registers the `prd-supplement` artifact-path pattern in
`plugins/vsdd-factory/config/artifact-path-registry.yaml` so that
`.factory/specs/prd-supplements/*.md` is recognized as a valid write target
by the factory-dispatcher path-validation hook chain.

This is a standalone infrastructure/config fix — not a story delivery. No
code, tests, stubs, or demo evidence are involved.

## Why

The S-25.02 F2 burst needed to materialize `error-taxonomy.md` under
`.factory/specs/prd-supplements/`, but that path pattern was not yet
registered in the artifact-path registry, causing the write to be blocked
by the path-validation guard. This PR closes that gap so future writes to
`.factory/specs/prd-supplements/*.md` are permitted without requiring an
ad hoc bypass.

## Diff Summary

- `plugins/vsdd-factory/config/artifact-path-registry.yaml`: +9 lines,
  registering the `prd-supplement` pattern.

## Testing

- [x] YAML-only change; no code paths affected.
- [x] No new tests required — config registration only.
- [ ] Demo recorded — N/A (non-behavior-changing infra config chore).

## Pre-Merge Checklist

- [x] Branch based on current `origin/develop`.
- [x] Single, minimal diff (9-line addition only).
- [x] No AI attribution in commit/PR per project convention.
- [ ] CI green (fmt/clippy/test/bats).
- [ ] PR review convergence (pr-reviewer APPROVE).

https://claude.ai/code/session_01Y45N2GutCTbmvLPG8cv9xu

### DIFF
--- a/plugins/vsdd-factory/config/artifact-path-registry.yaml
+++ b/plugins/vsdd-factory/config/artifact-path-registry.yaml
@@ -143,6 +143,15 @@ artifacts:
     description: Product requirements document
     enforcement_level: block
 
+  - artifact_type: prd-supplement
+    canonical_path_pattern: ".factory/specs/prd-supplements/{slug}.md"
+    description: >
+      PRD supplement extracted from the monolithic PRD per DF-021 (interface-definitions.md,
+      error-taxonomy.md, test-vectors.md, nfr-catalog.md). Producer: product-owner. Added
+      S-25.02 F2 fix-burst (F-S2502-F2-006) closing a pre-existing gap — prd.md §5.1/§5b/§4/§3
+      have referenced this directory since before this registry entry existed.
+    enforcement_level: block
+
   - artifact_type: state
     canonical_path_pattern: ".factory/STATE.md"
     description: Pipeline state document


Zious11:
# PR Review — #817 `chore(config): register prd-supplement artifact path pattern`

**Verdict: APPROVE**
**Reviewed SHA:** `dbe4b6b67e34d3586198c616188de5969fe7cca9`
**Scope:** 1 file, +9 / -0. Config-only, declarative. Not a story PR — no stubs,
Red Gate, or demo evidence expected; checklist items 3/4 scoped accordingly.

---

### 1. Diff coherence — PASS
The entire diff is a single `artifacts:` list entry in
`plugins/vsdd-factory/config/artifact-path-registry.yaml`. No collateral edits,
no reflowing of neighbouring entries, no version bump of the `version: 1` schema
field (correctly — the schema shape is unchanged, only a new instance is added).
Placement is correct: the entry sits inside the
`── Product Requirements / State ──` band, immediately after `prd`, which is
where a PRD-derived artifact belongs.

### 2. Pattern correctness — PASS (verified against the matcher, not assumed)
The registry's documented placeholder semantics are implemented in
`crates/hook-plugins/validate-artifact-path/src/lib.rs` (`pattern_matches` /
`matches_canonical`): per BC-4.11.001 invariant 6 (v1.1), `{placeholder}` matches
**one or more characters that do not contain `/`** — a single path segment.

Applying that to `.factory/specs/prd-supplements/{slug}.md`:

- `.factory/specs/prd-supplements/error-taxonomy.md` → **matches** (placeholder
  content `error-taxonomy`, no `/`), so the stated purpose of the PR is actually
  achieved.
- The other three DF-021 supplements (`interface-definitions.md`,
  `test-vectors.md`, `nfr-catalog.md`) match by the same construction.
- `.factory/specs/prd-supplements/nested/dir.md` → **does not match**, because the
  placeholder cannot span `/`. This is the desired behaviour: the entry grants
  exactly one directory level and does not silently widen the allowlist. Good.

The `{placeholder}.md` shape is the established idiom in this file — cf.
`VP-{vp-id}.md`, `S-{story-id}-{slug}.md`, `ADR-{adr-id}-{slug}.md` — so the
matcher is exercised on this shape already.

### 3. Consistency with existing entries — PASS
Checked mechanically after parsing the branch version of the file:

- 72 entries parse cleanly under `yaml.safe_load`; **no duplicate
  `artifact_type`** and **no duplicate `canonical_path_pattern`** introduced.
- Key set `{artifact_type, canonical_path_pattern, description,
  enforcement_level}` is the uniform shape for every entry in the file. The
  optional fifth key `required_frontmatter` is used by exactly one entry
  (`proposal`); every other `.factory/specs/**` artifact omits it, so omitting it
  here is the consistent choice, not an oversight.
- `enforcement_level: block` matches all 72 entries and the file's own header
  note ("Per OQ5 resolution: all entries ship with enforcement_level set to
  block"). Correct.
- `description: >` folded scalar matches 16 existing multi-line descriptions.
  Consistent style.
- **Placeholder token choice is the right one.** `{slug}` is already used 5×
  in this registry; `{name}` appears nowhere. Picking `{slug}` keeps the
  vocabulary closed rather than introducing a synonym for the same concept.

### 4. Effect without a release — PASS (worth stating explicitly)
The plugin does **not** embed the registry via `include_str!`. It reads it at
runtime through the host callback:
`REGISTRY_PATH = "plugins/vsdd-factory/config/artifact-path-registry.yaml"`, loaded
in `lib.rs` via `(callbacks.read_file)(REGISTRY_PATH)`. So this config-only change
becomes effective on merge to `develop` without cutting a new rc — no dispatcher
rebuild or marketplace-cache refresh is required for the gap to close. That
matters for the PR's stated purpose and is a point in its favour.

### 5. Test coverage — PASS (correctly none added)
This is declarative data consumed by an already-tested matcher. There is no new
code path: `pattern_matches` is covered by `src/tests.rs` and a Kani harness
(`tests/kani_path_matching.rs`), and no test in the tree asserts a fixed registry
entry count — the one place `registry.artifacts.len()` appears
(`tests.rs`, the EC-007 perf budget test) uses it only for a diagnostic message,
not as an assertion operand, so adding a 72nd entry cannot break it. Adding a
bespoke test for one data row would be noise. Agreed with "no new tests required".

### 6. Description / commit accuracy — PASS
PR body and commit message both describe exactly a 9-line YAML addition
registering `prd-supplement`, and both correctly disclaim demo evidence as N/A for
a non-behaviour-changing config chore. Commit subject is conventional
(`chore(config): …`), imperative, and carries no AI attribution — matching project
convention. Single commit on the branch; no chain to squash-collapse.

### 7. Diff size — PASS
9 lines. Far below the 500-line flag threshold.

### 8. Missing changes / dependencies — PASS
Nothing else is required for this to be complete on its own terms. The four
supplement filenames named in the description are already referenced across
`plugins/vsdd-factory/agents/` (product-owner, security-reviewer,
performance-engineer, story-writer, e2e-tester, consistency-validator) and in
`workflows/greenfield.lobster` and `phases/phase-2-story-decomposition.lobster`,
so the directory was a live convention that the registry had simply never
enumerated. This PR is the registry catching up, and it has no upstream PR
dependency.

---

## Findings

| 1 | NIT | description | The `description` embeds burst-specific provenance (`S-25.02 F2 fix-burst (F-S2502-F2-006)`) and a section-pin list (`prd.md §5.1/§5b/§4/§3`). Section anchors are more stable than line numbers, so this is not a TD-VSDD-091 volatile-pin violation, and neighbouring entries do carry provenance ("POLICY 9 propagation target from VP-INDEX", "written by skills/create-brief"). Still, this is the longest description in the file by some margin, and the historical narrative will read as stale once the burst is closed. | Optional. If you trim on a follow-up, keep "PRD supplement extracted from the monolithic PRD per DF-021 (interface-definitions.md, error-taxonomy.md, test-vectors.md, nfr-catalog.md). Producer: product-owner." and drop the burst-history sentence — the decision log is the durable home for provenance. Not worth a round-trip on its own. |

**No BLOCKING findings. No SUGGESTION-severity findings.**

---

## One thing for the merging agent to be aware of (not a finding against this PR)

At review time the local `develop` checkout carries an **uncommitted working-tree
modification** to this same file adding a *competing* `prd-supplement` entry, with
`canonical_path_pattern: ".factory/specs/prd-supplements/{name}.md"` and a
different description. That variant is not part of this PR and is not on any
branch.

This PR's version is the better of the two — `{slug}` is the token this registry
already uses, whereas `{name}` would introduce a one-off synonym. Flagging only so
the uncommitted variant is discarded rather than committed on top after merge,
which would otherwise produce two `prd-supplement` entries with divergent
placeholder tokens.

---

## CI

No failing checks. At review time: `validate`, `SAST (Semgrep)`,
`policy-15-attestation-location`, `attestation-gate-non-vacuity-controls`,
`platforms-drift`, `deny-advisories`, `bats-wave-handoff (macos)` all green;
`cargo-host` (×2), `bats-full-suite (linux)`, and `build-dispatcher` (×5) still
running. Merge remains gated on those completing green — that gate is the
pr-manager's, not this review's.

PR #818, opened 2026-09-07
feat(S-25.02): cluster 1 — cap formula + native shard-cap trigger (BC-1.18.005)

Zious11:
# [S-25.02] Artifact Sharding Layer 2 — Cluster 1: Cap Formula + Native Size/Item-Count Trigger

**Epic:** S-25.02 — Artifact Sharding Layer 2 (BC-1.18.001..012)
**Mode:** feature (Feature Mode F4, cycle `v1.0-brownfield-backfill`)
**Convergence:** CONVERGED at BC-5.39.001 3-CLEAN (LOCAL adversary passes 10/11/12, D-1172) after 12 cluster-1 LOCAL adversarial passes total

Tests
Bats
fmt
clippy

This PR delivers cluster 1 of the incremental-by-BC-cluster delivery of S-25.02 (per D-1170): the
native (non-WASM) dispatcher PreToolUse **shard-cap GATE + TRIGGER** for **BC-1.18.005**. It adds
`crates/factory-dispatcher/src/shard_manager.rs`, computing a `shard_cap_bytes` ceiling from four
calibrated config inputs, dispatching on artifact `shape` to either a `stat()`-only flat byte-size
trigger or an item-count trigger for the `"frontmatter-changelog-array"` shape, and wiring the
match-first `ShardRegistry::load()` (structural-parse-only) + pure `validate_entry` gate into
`executor.rs`'s PreToolUse path with fail-loud block-reason surfacing for load-time config defects
(EC-009/EC-011/EC-013/EC-015/EC-016/EC-017) and graceful missing-file degradation (EC-004/EC-014).

**Cluster boundary (explicit, spec-recorded — not a gap):** this cluster owns the cap-formula and
the trigger-BOUNDARY decision only. The *observable* roll-before-write + `HookResult::Block` action
is owned by **BC-1.18.006** (cluster 2, not yet in scope). Today, a fired byte-size trigger on the
`"flat"` shape surfaces as `tracing::warn!` + `Continue` — an honest interim, not silent
swallowing (BC-1.18.005 Postcondition 3, "Ownership" text).

---

## Architecture Changes

```mermaid
graph TD
    PreToolUse["executor.rs::PreToolUse dispatch"] -->|Edit/Write/MultiEdit| ShardGate["shard_cap_precheck()"]
    ShardGate -->|config exists?| ConfigProbe["Path::exists() shard-config.toml (O(1), fuel-free)"]
    ConfigProbe -->|no| Continue1["Continue (zero-cost bypass)"]
[...]

PR #824, opened 2026-09-08
S-25.02 cluster-2: artifact-sharding roll (BC-1.18.006 v1.11)

Zious11:
# [S-25.02 cluster-2] Artifact sharding layer-2 — roll (BC-1.18.006 v1.11)

**Epic:** S-25 — Artifact Sharding
**Mode:** feature (brownfield, engine-discipline concurrent cycle)
**Cluster:** 2 of 7 BC-clusters for S-25.02 (cluster-1 "cap+trigger", BC-1.18.005, merged as PR #818)

Delivers BC-1.18.006 — the staged, crash-recoverable roll-before-write sequence
(copy-then-atomic-truncate, never rename-away) that BC-1.18.005's cap/trigger
gate hands off to once a canonical artifact exceeds its shard cap. This PR
implements and hardens the roll sequence itself: seal the pre-roll canonical
content into an immutable numbered shard, atomically truncate the canonical
to 0 bytes, and publish the shard-index entry — plus the write-once
immutability guard on sealed shards (`E-SHD-009`), its bounded 0-byte
reclaim exception, and the three sanctioned retry-instruction `Block`
message templates (unified rotate-and-retry, plus the empty-canonical
Case B1/B2 split).

---

## Spec Traceability

- **Story:** `.factory/stories/S-25.02-artifact-sharding-layer2.md` v3.3
- **Behavioral contract:** `.factory/specs/behavioral-contracts/ss-01/BC-1.18.006.md` v1.12 — "Staged, Crash-Recoverable Roll-Before-Write Sequence (Copy-Then-Atomic-Truncate, Never Rename-Away)"
- **Error taxonomy:** v1.8 (`E-SHD-008`, `E-SHD-009`)

```mermaid
flowchart LR
    BC["BC-1.18.006 v1.12<br/>roll-before-write"] --> AC006["AC-006<br/>roll: seal→truncate→index,<br/>write-once E-SHD-009,<br/>0-byte reclaim"]
    BC --> AC007["AC-007<br/>empty-canonical Block<br/>templates Case B1/B2"]
    AC006 --> T1["test_..._AC006_over_cap_write_dispatch_stages_full_roll_sequence"]
    AC006 --> T2["test_..._FC2P4_002_publish_sealed_shard_refuses_to_overwrite_existing_seq"]
    AC006 --> T3["test_..._EC025_FC2P8_002_execute_roll_reclaims_zero_byte_destination_and_succeeds"]
    AC007 --> T4["test_..._FC2P4_003_empty_canonical_block_message_matches_verbatim_template"]
    AC007 --> T5["test_..._EC026_FC2P8_004_double_fire_emits_case_b2_not_case_b1"]
    T1 --> S1["shard_manager.rs::execute_roll"]
    T2 --> S1
    T3 --> S1
    T4 --> S2["shard_manager.rs::build_empty_roll_retry_block_reason"]
    T5 --> S2
```

| Requirement | Story AC | Postconditions / Invariants / ECs | Test(s) | Status |
|---|---|---|---|---|
| Roll: seal→truncate→index, byte-level I/O, self-heal zero-byte-seal guard | AC-006 | PC1, PC2, PC8, Inv 2/3/9, EC-021/022/024/025 | `bc_1_18_006_roll_test.rs` (full-stack, dispatch-driven) + `shard_manager.rs` unit tests | PASS |
| Write-once sealed shards (non-empty destination) — `E-SHD-009` | AC-006 | PC8, EC-024 | `test_BC_1_18_006_FC2P4_002_publish_sealed_shard_refuses_to_overwrite_existing_seq` | PASS |
| Bounded 0-byte-destination reclaim (single stat→unlink→retry, never a loop) | AC-006 | PC8's 0-byte exception, EC-025/EC-046 | `test_BC_1_18_006_EC025_FC2P8_002_execute_roll_reclaims_zero_byte_destination_and_succeeds` | PASS |
| Empty-canonical Block templates — unified, Case B1 (pure), Case B2 (double-fire) | AC-007 | PC2, Inv 4, EC-021/026/047 | `test_BC_1_18_006_FC2P4_003_...verbatim_template` + `test_BC_1_18_006_EC026_FC2P8_004_double_fire_emits_case_b2_not_case_b1` | PASS |
| Fired shard-gate verdict must survive a `build_engine()` failure on the empty-tiers path (post-hoc MAJOR-3 sweep finding) | AC-006 | PC1/PC8 (verdict-integrity corollary) | `test_MINORN1_fired_shard_gate_verdict_on_empty_tiers_survives_build_engine_failure` | PASS |

---

## Architecture Changes

```mermaid
graph TD
    ExecuteTiers["executor.rs::execute_tiers"] -->|shard_cap_precheck| ShardCapGate["shard_manager.rs::shard_cap_gate_check"]
    ShardCapGate -->|trigger fires| ExecuteRoll["shard_manager.rs::execute_roll"]
    ExecuteRoll -->|step b| PublishSealed["shard_manager.rs::publish_sealed_shard<br/>write_exclusive, E-SHD-009 write-once"]
    ExecuteRoll -->|step c| TruncateCanonical["atomic truncate-in-place<br/>never rename-away"]
    ExecuteRoll -->|step d| PublishIndex["shard-index TOML publish"]
    ExecuteRoll -.->|Ok(None) short-circuit| BlockTemplate["build_empty_roll_retry_block_reason<br/>Case A / B1 / B2"]
    ShardCapGate -->|fired verdict, empty tiers| VerdictOutcome["executor::shard_gate_verdict_outcomes<br/>short-circuits exit code BEFORE build_engine()"]
    style PublishSealed fill:#90EE90
    style BlockTemplate fill:#90EE90
    style VerdictOutcome fill:#90EE90
```

`crates/factory-dispatcher/src/shard_manager.rs` carries the bulk of this
change (staged roll sequence, write-once guard + 0-byte reclaim, three-case
Block template dispatch). `executor.rs`, `invoke.rs`, and `main.rs` carry the
wiring for self-heal-before-roll sequencing (catch point (i)) and the
Write-arm backstop probe (catch point (ii)) established across the LOCAL
adversarial cascade below, plus the PreToolUse byte-cap gate hoist to
`main::run` per ADR-051 §Decision 17 (cycle-3 MAJOR-3 remediation, see
PR Review Findings — Cycle 3 below), and the cycle-4 MINOR-N1 fix
(`54d6c84b`) that extracts `executor::shard_gate_verdict_outcomes` so a
fired verdict's exit code is resolved before `build_engine()` is ever
called (see PR Review Findings — Cycle 4 below).

---

## Test Evidence

| Metric | Value |
|---|---|
| Full workspace test suite | Locally re-verified at tip `27bbc6f8` by cycle-4 fresh-context review: `shard_manager::` lib tests 162/162 pass, 0 failed; cluster-1 + cluster-2 integration suites (`bc_1_18_005_shard_cap_trigger_test`, `bc_1_18_006_roll_test`) 0 failed. Follow-up commit `54d6c84b` (MINOR-N1 fix) adds `test_MINORN1_fired_shard_gate_verdict_on_empty_tiers_survives_build_engine_failure` (asserts exit 2, proven red-without-the-fix); local gate re-verified green including bats. Full `cargo test --workspace --all-targets` is CI's job (`cargo-host`, both runners) — pending, see Pre-Merge Checklist |
| `cargo fmt --check --all` | clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| New/extended test file | `crates/factory-dispatcher/tests/bc_1_18_006_roll_test.rs` (2,108+ lines added this cluster) |
| Story spec Red Gate obligations (AC-006/AC-007) | all landed and green |

Coverage/mutation-kill-rate percentages are not independently re-measured
for this PR body — the workspace-green result above (tip `54d6c84b`) is the
authoritative, reproducible gate (`cargo test --workspace --all-targets`);
CI (step 6 of this PR's lifecycle) re-runs it from a clean checkout.

### Demo Evidence

`docs/demo-evidence/S-25.02/cluster-2-roll/` — 5 real VHS terminal recordings
(`vhs 0.11.0`), one per acceptance-criterion behavior, each running the real
`cargo test <name> -- --exact --nocapture` against the actual roll logic (no
manufactured or hand-typed output):

| Clip | AC / EC | Test |
|---|---|---|
| `AC-006-roll-seal-truncate.{gif,webm}` | AC-006, PC1 steps a-d | `test_BC_1_18_006_AC006_over_cap_write_dispatch_stages_full_roll_sequence` |
| `AC-006-write-once-e-shd-009.{gif,webm}` | AC-006, PC8 / EC-024 | `test_BC_1_18_006_FC2P4_002_publish_sealed_shard_refuses_to_overwrite_existing_seq` |
| `AC-006-0-byte-reclaim.{gif,webm}` | AC-006, EC-025 | `test_BC_1_18_006_EC025_FC2P8_002_execute_roll_reclaims_zero_byte_destination_and_succeeds` |
| `AC-007-empty-canonical-block.{gif,webm}` | AC-007, PC2 / Inv 4, Case B1 | `test_BC_1_18_006_FC2P4_003_empty_canonical_block_message_matches_verbatim_template` |
| `AC-007-double-fire-case-b2.{gif,webm}` | AC-007, EC-026, Case B2 | `test_BC_1_18_006_EC026_FC2P8_004_double_fire_emits_case_b2_not_case_b1` |

Scope note (from the demo README): AC-008/AC-009 (shard+index atomicity,
shard-index schema) and the self-heal/catch-point recovery paths are
exercised by this branch's test suite (all green) but not separately
recorded in this pass — they belong to a later cluster-2 demo pass if
dispatched.

---

## Convergence Provenance — stated honestly

This is **not** a literal BC-5.39.001 3/3 clean cascade. The LOCAL
adversarial cascade for this cluster ran **10 fresh-context passes**:

| Pass | Result |
|---|---|
| P1–P6 | findings at MAJOR/MEDIUM severity, fixed each pass |
| P7 | MAJOR finding, fixed |
| P8 | MEDIUM findings (F-C2-P8-001/002/004), fixed — BC-1.18.006 v1.8→v1.9 |
| P9 | MINOR-only (F-C2-P9-001/002/003) — correctness surface clean |
| P10 | MINOR-only — correctness surface clean, second consecutive clean-on-correctness pass |

Closed via **human-authorized asymptotic acceptance (D-386 Option C)**, not
a literal 3/3 streak — distinct from cluster-1's literal 3/3 convergence.
This PR-level review (pr-reviewer, dispatched next in this lifecycle) is the
next independent review layer on top of that LOCAL cascade, not a
replacement for it. That PR-level review is itself now on its 5th cycle —
see PR Review Findings sections below.

---

## Security Review

Two independent fresh-context security-reviewer passes were run against the PR diff — each found a genuinely different, non-overlapping subset of the findings below (the second pass caught the HIGH finding the first pass missed)
(`crates/factory-dispatcher/src/shard_manager.rs` primarily). No CRITICAL findings.
All HIGH/MEDIUM/LOW findings below were fixed in-scope on this branch before merge,
per this repo's production-grade default (fix in-scope rather than defer).

| Finding | Severity | CWE | Fix commit | Status |
|---|---|---|---|---|
| `write_exclusive`'s temp file (`.{basename}.tmp-{pid}`) created via `File::create` (no `O_EXCL`), symlink-followable, could defeat write-once seal immutability | HIGH | CWE-59, CWE-367, CWE-377 | `0f56530d` | Fixed — `OpenOptions::create_new` (O_EXCL) |
| 0-byte seal reclaim used `stat()` (follows symlinks) instead of `lstat()` to decide reclaimability | MEDIUM | CWE-61, CWE-367 | `e67eb7ad` | Fixed — `symlink_metadata`, refuses symlinks |
| Config-supplied `artifact_stem` not validated against path-traversal characters before building filesystem paths | MEDIUM | CWE-22 | `002962ce` | Fixed — new `InvalidArtifactStem` fail-loud check |
| TOCTOU in 0-byte reclaim: `stat()`→`unlink()` didn't re-verify identity immediately before unlink | MEDIUM | CWE-367 | `0ea79c2c` | Fixed — open-handle re-check before unlink |
| Symlink-following reads on `canonical_path` (4 sites) could exfiltrate arbitrary file content into a durably-sealed, committable shard | MEDIUM | CWE-59, CWE-200 | `8d17ffc4` | Fixed — new `E-SHD-010` fail-loud guard at all 4 sites |
| No rejection of `..` (`ParentDir`) components in the tool-call-supplied path comparison | LOW | CWE-22 | `5e025366` | Fixed — rejects `ParentDir` components |

**Known follow-up (documentary, not a code gap):** the new `E-SHD-010` error variant
(symlink-on-canonical-path guard) is not yet documented in `error-taxonomy.md`.
Per this repo's Agent Routing Table, `error-taxonomy.md` is product-owner's domain,
not implementer's or pr-manager's — flagged for a follow-up documentary pass rather
than edited here. **Update:** this row has since been added by product-owner on
`factory-artifacts` (commit `3bf113c1`) — see the PR Review Findings table below.

---

## PR Review Findings — Cycle 1 (Resolved)

Fresh-context `pr-reviewer` reviewed this PR at an earlier head SHA and returned
**REQUEST_CHANGES** (1 BLOCKING/external, 1 MAJOR, 5 MINOR, 2 NIT). Full detail:
`.factory/code-delivery/S-25.02/pr-review.md`. All 10 findings have since been
addressed on this branch (current head `54d6c84b`):

| # | Severity | Finding | Resolution | Ref |
|---|---|---|---|---|
| 1 | BLOCKING | `cargo-host` red on both runners — a `.factory/STATE.md` banner line-count assertion (`validate-state-structure`), unrelated to this diff | External to this PR's code; not fixable on this branch. Routed to `state-manager` for a STATE.md banner reconcile on `factory-artifacts`. Tracked by the orchestrator, not gating this report. | — |
| 2 | MAJOR | `write_exclusive` temp-path collision misattributed as a real `E-SHD-009` sealed-shard collision; a failed op could unlink a reclaimable 0-byte destination | Fixed — stage-then-publish with a non-deterministic temp path and a typed distinction between temp-path-occupied and destination-exists | `ef6ca3b4` |
| 3 | MAJOR | `E-SHD-010` symlink guard absent and untested on the Edit (~L1678) and MultiEdit (~L1748) arms (Write-only backstop before this fix) | Fixed — guard added to Edit + MultiEdit arms, with new Edit/MultiEdit test variants. Independently **CONFIRMED-COMPLETE** by security-reviewer. One residual **pre-existing LOW CWE-367 TOCTOU** was flagged separately by security-reviewer as accepted/LOW; that is a distinct, human-decision-pending item, not part of this finding's scope. | `ef6ca3b4` |
| 4 | MINOR | `reclaim_identity_still_safe`'s `File::open` could block indefinitely on a FIFO | Fixed — non-blocking, no-follow open semantics (`O_NONBLOCK \| O_NOFOLLOW`) | `c3c87bd6` |
| 5 | MINOR | PR body's "Deferred to F6" section cited commit `6cccf3a3` as this PR's FIX-MED-1 commit; that SHA is **not an ancestor of HEAD** (orphaned artifact of a discarded branch state) | Fixed by this edit — corrected to `0ea79c2c` (the in-HEAD FIX-MED-1 commit; parent `0f56530d`), matching the SHA already cited correctly in the Security Review table above | this edit |
| 6 | MINOR | Demo README's 0-byte-reclaim prose (EC-025 row) described the pre-SEC-001/pre-FIX-MED-1 `stat()`-once mechanism | Fixed — README updated to describe the current `lstat` + open-handle re-check behavior | `ac56f244` |
| 7 | MINOR | `E-SHD-010` shipped without an `error-taxonomy.md` entry or a named deferral anchor | Closed — taxonomy row added by product-owner on `factory-artifacts` | `3bf113c1` |
| 8 | MINOR | FIX-MED-1's TOCTOU re-check was tested only at the `reclaim_identity_still_safe` helper level, never through `publish_sealed_shard` itself | Fixed — new integration-level test exercises the real call site | `27d2f5ec` |
| 9 | NIT | `next_seal_seq`'s `max + 1` could overflow `u32` | Fixed — `checked_add(1)` with a fail-loud error | `c3c87bd6` |
| 10 | NIT | Diff size (8,847/2,253) far over the 500-line guideline | Acknowledged (see PR comment) — ≈3:1 test-to-production ratio on a data-integrity path is the correct trade-off; no split warranted | see PR comment |

A cycle-2 fresh-context `pr-reviewer` re-review was performed against head
`27d2f5ec`; its findings and verdict are tracked in
`.factory/code-delivery/S-25.02/pr-review-cycle-2.md`. A cycle-3 fresh-context
re-review followed (3 MAJOR + 12 MINOR/NIT findings, all resolved — see the
Cycle 3 section below), tracked in
`.factory/code-delivery/S-25.02/pr-review-cycle-3.md`. A cycle-4
fresh-context re-review is tracked in
`.factory/code-delivery/S-25.02/pr-review-cycle-4.md`, and a cycle-5
fresh-context re-review (confirming the cycle-4 MINOR-N1 fix) is tracked in
`.factory/code-delivery/S-25.02/pr-review-cycle-5.md`.

---

## PR Review Findings — Cycle 3 (Resolved)

Fresh-context `pr-reviewer` cycle-3 re-review independently confirmed all 5
cycle-2 claimed-fixed items as genuinely fixed (mutation-tested), then
surfaced 3 new MAJOR findings and 12 MINOR/NIT findings — none overlapping
with the cycle-1/cycle-2 sets above. All resolved on this branch prior to
this cycle-4 dispatch:

| # | Severity | Finding | Resolution | Commit |
|---|---|---|---|---|
| MAJOR-1 | MAJOR | Finding #2's fix (`ef6ca3b4`) silently widened FIX-MED-1's TOCTOU re-check window by inserting a full staged write (incl. `fsync`) between the re-check and the unlink — the exact data-loss scenario FIX-MED-1 exists to prevent | Fixed — re-check restored genuinely adjacent to `remove_file`; all 4 stale doc/comment sites corrected; ordering-pinning regression test added | `e9941b4b` |
| MAJOR-2 | MAJOR | New catch-point-(i) PostToolUse leg reached the destructive `execute_roll` without `validate_entry`, bypassing EC-022 (and the EC-010/011/013/015/017 cap-sanity checks) | Fixed — `validate_entry(&entry)` now called in `reconcile_replace_all_overcap_if_qualifying` immediately after `find_matching_entry`; fail-open-but-never-silent `tracing::warn!` added on `Err` | `0d91a4c3` |
| MAJOR-3 | MAJOR | `main.rs` placement rationale asserted a PreToolUse gate guarantee that did not exist — the byte-cap gate was reachable only *after* the empty-plugin-set early return, unlike the new catch-point-(i) it claimed parity with | Fixed — PreToolUse byte-cap gate hoisted to `main::run`, before the early return, per **ADR-051 §Decision 17**; both stale comment sites (`main.rs`, `invoke.rs`) corrected to state the (now-closed) parity accurately | `6d3d96e7` |
| MINOR-1..6, NIT-1..5 | MINOR/NIT | Swallowed `DuplicateArtifactStem` telemetry, silently-`None` `file_path` handling, false "unconditional/independent of registry" claim, false "exercised end-to-end" claim, unconsidered PostToolUse-ordering side effect, factually-wrong dependency-justification comment, inverted `O_NOFOLLOW`/`O_NOCTTY` constants comment, doc-overstated symlink acceptance, ancestor-probe doc gap, retracted-framing reintroduction, dependency-direction cleanup | All fixed — warns added where "never silent" was violated, doc/comment corrections, dependency-direction note | `878e6ab1`, `a4afa013`, `c0a57d75`, `603eface`, `9d749546`, `02dfd02f`, `361fa507` |
| MINOR-7 | MINOR | PR description stale against the reviewed head (`27d2f5ec` cited as "current head" in two places; hedged/ambiguous cycle-2 tracking pointer) | Fixed — this edit: current-head references refreshed to `27bbc6f8`; cycle-2 pointer resolved unambiguously to `pr-review-cycle-2.md`; Test Evidence section refreshed to current tip | `27bbc6f8` + this edit |

**Spec closure (factory-artifacts, commit `7c71b193`):** ADR-051 §Decision 17
(PreToolUse byte-cap gate placement rationale), BC-1.18.005 v1.15,
BC-1.18.006 v1.12, BC-INDEX/ARCH-INDEX v1.11.

**Known CI flake (unrelated to this PR):** `bats-full-suite`'s
`precompact-routing.bats` TC-AC004/TC-EC001 is a known pre-existing flake;
a failure limited to that specific test pattern is not a regression
introduced by this PR and should not block this PR's merge gate on its own.

Full detail: `.factory/code-delivery/S-25.02/pr-review-cycle-3.md`.

---

## PR Review Findings — Cycle 4

Fresh-context `pr-reviewer` cycle-4 re-review (head `27bbc6f8`, confirmed
current via `gh pr view 824 --json headRefOid`) independently re-verified
all three cycle-3 MAJOR remediations by deliberate mutation testing
(reverting each fix reproduces exactly the defect it closes) and confirmed
the cycle-3 MINOR/NIT set as addressed, with the two exceptions below.

**VERDICT: APPROVE** — zero MAJOR/blocking findings.

| # | Severity | Finding | Status |
|---|---|---|---|
| MINOR-N1 | MINOR (new) | MAJOR-3's gate hoist places the destructive `execute_roll` before a fallible (effectively-never-fails) `build_engine()`; on a `build_engine()` failure a fired shard-gate verdict is silently downgraded to exit 0 even though the roll already mutated the canonical | **Resolved by commit `54d6c84b`** — extracted `executor::shard_gate_verdict_outcomes`, which resolves a fired shard-gate verdict to its exit code directly and returns it BEFORE `build_engine()` is invoked, closing the roll-happened/verdict-dropped divergence entirely rather than narrowing it. Backed by a test-only fault-injection seam (`VSDD_FORCE_ENGINE_BUILD_FAILURE`, `#[cfg]`-gated out of shipped/release binaries) and a new load-bearing test `test_MINORN1_fired_shard_gate_verdict_on_empty_tiers_survives_build_engine_failure` (asserts exit 2; proven red-without-the-fix). Local gate (fmt/clippy/cargo test/bats) re-verified green. Independent cycle-5 fresh-context re-verification tracked in `.factory/code-delivery/S-25.02/pr-review-cycle-5.md`. |
| MINOR-N2 (= cycle-3 MINOR-7) | MINOR | PR body still stale (none of the 8 cycle-3 remediation commits touch the body) | Resolved by this edit (and this cycle-5 follow-up edit, which refreshes head references to `54d6c84b`) |

CI at `27bbc6f8`: `validate`, `policy-15-attestation-location`,
`platforms-drift`, `deny-advisories`, `attestation-gate-non-vacuity-controls`,
and SAST/Semgrep pass; `build-dispatcher` (all 5 platforms, including the
merge-critical `windows-x64` — the only job exercising the `#[cfg(windows)]`
N-2 fix tests), `cargo-host` (both runners), `bats-full-suite`, and
`bats-wave-handoff` are still pending. Must be green before merge.

Full detail: `.factory/code-delivery/S-25.02/pr-review-cycle-4.md`.

---

## Deferred to F6 targeted-hardening

Both items below are **human-authorized** deferrals, anchored to Phase F6
(targeted hardening) per the story's own Phase F6 scope, not ad hoc AI
deferrals. Item 1's underlying TOCTOU sub-window is now substantially
narrowed by this PR's own `FIX-MED-1` commit (`0ea79c2c`, open-file-handle
identity re-verification immediately before `unlink()`), but full
concurrent-race *test* coverage for both the original EC-025 retry path and
the new open-handle re-check still needs the injection seam described below.

1. **O_EXCL re-create-then-swap hardening of the 0-byte reclaim `stat()`→`unlink()` sub-window.** Narrowed (not fully closed) by `FIX-MED-1` in this PR — the reclaim now re-verifies identity via an open file handle immediately before unlinking, closing most of the window, but a genuinely atomic `O_EXCL`-based re-create-then-swap primitive would close it completely rather than narrowing it. Scoped to F6 formal hardening.
2. **Concurrent-race retry-collision fault-injection test (EC-025 + the new open-handle re-check).** Needs a `#[cfg(test)]` injection seam (to deterministically force a collision between the reclaim's re-check/`unlink()` and a concurrent writer) that does not yet exist in `shard_manager.rs` — the same gap already documented in-file for the pre-existing EC-025 unlink-failure test. The race-variant behavior is specified and partially covered (deterministic before/after-state tests, not a true concurrent-thread fixture), not yet exercised by a fault-injected concurrent fixture. Scoped to F6.

---

## Risk Assessment

- **Systems affected:** `factory-dispatcher` PreToolUse gate chain — `shard_manager.rs`, `executor.rs`, `invoke.rs`, `main.rs`. Native (non-WASM) dispatcher code path; no hook-registry or WASM plugin changes.
- **User impact on failure:** a regression here could either fail to roll an over-cap canonical (data-loss risk on the un-sealed pre-roll content) or wrongly refuse a legitimate write (availability risk). Both classes are covered by the Red Gate test suite and the LOCAL adversarial cascade above.
- **Blast radius:** scoped to the roll/seal/truncate/index code path this BC owns; does not touch the WASM hook-plugin tier or other subsystems.
- **Risk level:** MEDIUM (native dispatcher code, but heavily fenced by TDD + 10-pass adversarial cascade + demo evidence + 5 cycles of PR-level fresh-eyes review).

---

## Pre-Merge Checklist

- [ ] CI green (`fmt --check`, `clippy -D warnings`, `cargo test --workspace --all-targets`, bats) — fresh run in progress on `54d6c84b`
- [x] pr-reviewer fresh-eyes review — cycle-1 REQUEST_CHANGES findings all addressed; cycle-2 REQUEST_CHANGES findings (N-1..N-7) addressed; cycle-3 REQUEST_CHANGES findings (3 MAJOR + 12 MINOR/NIT) addressed; cycle-4 fresh-context re-review **APPROVE** (zero MAJOR/blocking; MINOR-N1 fixed by `54d6c84b`, MINOR-N2 resolved by PR-body edits — see Cycle 4 section above); cycle-5 confirming re-review tracked in `.factory/code-delivery/S-25.02/pr-review-cycle-5.md`
- [x] Security review — 1 HIGH + 4 MEDIUM + 1 LOW found across 2 independent passes, all fixed in-scope (6 commits, `e67eb7ad`..`8d17ffc4`); one residual pre-existing LOW CWE-367 TOCTOU noted separately, human-decision-pending
- [x] Demo evidence — 5/5 clips present, 1+ per AC
- [x] Dependency PR (cluster-1, BC-1.18.005, PR #818) already merged to develop


Zious11 (commented):
# pr-reviewer — fresh-eyes review of PR #824 (S-25.02 cluster-2 "roll", BC-1.18.006 v1.11)

**Reviewed at:** HEAD `8d17ffc4`, merge-base `fff5e4cc`, targeting `develop`.
**Scope:** the diff as it stands now, including the 6 post-PR security-fix commits (`e67eb7ad`..`8d17ffc4`).
**Verdict: REQUEST_CHANGES** — one MAJOR data-integrity/availability defect in the write-once seal path (reproduced empirically, below), plus an unmet CI-green merge gate.

This is a high-quality PR. The staged roll sequence is coherent, the fail-loud discipline is
consistent, the error taxonomy is well-factored, and the test-to-production ratio is roughly 3:1
(~2,000 lines of new production code against ~6,400 lines of tests). Every finding below is
specific; none of them is a style preference.

---

## Summary

| # | Severity | Category | Finding |
|---|---|---|---|
| 1 | BLOCKING | dependency / CI | `cargo-host` red on both runners — merge gate unmet (root cause is outside this diff) |
| 2 | MAJOR | correctness | `write_exclusive` temp-path collision misreported as `E-SHD-009`, and destroys a reclaimable 0-byte destination |
| 3 | MINOR | coverage | E-SHD-010 symlink guard misses the Edit/MultiEdit arm's own symlink-following `stat()` |
| 4 | MINOR | correctness | `reclaim_identity_still_safe`'s `File::open` can block indefinitely on a FIFO |
| 5 | MINOR | description | PR body cites an orphaned commit SHA (`6cccf3a3`) that is not an ancestor of HEAD |
| 6 | MINOR | description | Demo README's 0-byte-reclaim description is stale after SEC-001 + FIX-MED-1 |
| 7 | MINOR | missing | `E-SHD-010` ships without an `error-taxonomy.md` entry and without a story anchor for the deferral |
| 8 | NIT | correctness | `next_seal_seq`'s `max + 1` can overflow `u32` |
| 9 | NIT | size | 8,847 / 2,253 diff, far over the 500-line guideline |

---

## 1. [BLOCKING] CI is red on both runners — the PR's own "CI green" gate is unmet

`gh pr checks 824` reports `cargo-host (ubuntu-latest)` **fail** and `cargo-host (macos-latest)`
**fail**. Both fail in the same step (`cargo test (workspace, all targets)`) on the same two tests:

```
---- tests::test_BC_5_39_005_f_p1_001_real_state_md_banner_wc_passes stdout ----
panicked at crates/hook-plugins/validate-state-structure/src/lib.rs:2557:9:
assertion `left == right` failed: real STATE.md banner claims 386 lines but actual count is 395
  left: 386
 right: 395

---- tests::test_BC_5_39_005_full_validation_against_real_state_md stdout ----
validate_banner_wc must return None for real STATE.md; got: Some("banner claims 386 lines but
actual line count is 395 — reconcile banner wc-l ...")
```

**This is not caused by this PR's diff.** I verified:

- Locally on this branch, `cargo test --workspace --all-targets` is **3091 passed / 0 failed**
  (the `.factory/` worktree these two tests read is not mounted in my checkout, so they do not run).
- `develop`'s own CI has been red for the **last 5 consecutive runs** (2026-09-05 → 2026-09-07),
  including at this PR's merge-base `fff5e4cc`.

So the failure is a pre-existing, branch-wide condition: `validate-state-structure` asserts against
the live `.factory/STATE.md` mounted in CI, and that file's banner `wc -l` is stale (386 vs 395).

I am flagging it as BLOCKING because the PR's own Pre-Merge Checklist lists "CI green" as a gate and
that gate is objectively unmet — not because the author introduced it. Two things follow:

1. **Routing:** the fix belongs to `state-manager` (STATE.md banner reconcile), not to this PR.
2. **Worth escalating on its own:** `develop` has been merging with a red `cargo test` gate for
   3+ days. Every PR merged in that window had a non-functional workspace-test gate. Separately,
   coupling PR CI to a mutable external branch's content makes the PR signal non-hermetic — a
   change no PR author can control can turn every PR red.

---

## 2. [MAJOR] `write_exclusive`'s temp-path collision is indistinguishable from a real sealed-shard collision

**File:** `crates/factory-dispatcher/src/shard_manager.rs`, `write_exclusive` (~L2435-2467) and
`publish_sealed_shard` (~L2495-2600).

`write_exclusive` creates its temp file at a fully deterministic path,
`.{basename}.tmp-{pid}`, with `create_new(true)`. FIX-HIGH-1 correctly made that `O_EXCL` so a
pre-planted symlink cannot be written through — and there is a good regression test for exactly
that (`test_FIXHIGH1_write_exclusive_refuses_to_follow_preplanted_symlink_at_temp_path`, which
asserts the returned `io::ErrorKind` is `AlreadyExists`).

The problem is what the **caller** does with that `AlreadyExists`. `write_exclusive` returns a bare
`io::Result<()>`, so `publish_sealed_shard` cannot tell "the temp path was occupied" from "the
sealed destination was occupied" — and it assumes the latter:

```rust
if first_err.kind() != io::ErrorKind::AlreadyExists { /* SealWriteFailed */ }
// ...otherwise fall into the destination-collision / 0-byte-reclaim path
```

The doc comment on `write_exclusive` states this is intentional — *"The resulting `AlreadyExists`
is handled by `publish_sealed_shard`'s EXISTING collision-handling path exactly like any other
exclusive-create collision — no new fail-loud branch is needed here."* That reasoning conflates two
different conditions, and the consequences are observable. I added two temporary probes to the
crate and ran them (then reverted; the working tree is clean):

**Probe A — no sealed shard exists at all, only a stale/planted temp file:**

```
PROBE-A err = E-SHD-009: refusing to overwrite an already-sealed shard at
  '.../decision-log.0001.md' for artifact_stem "decision-log" — sealed shards are
  write-once/immutable; this seq already has durable content on disk
PROBE-A sealed exists = false
```

The error is factually false: no sealed shard exists and no durable content is on disk. Worse, the
condition is **self-perpetuating** — nothing ever removes that temp file, so every subsequent roll
of that artifact from a process with that pid fails identically.

**Probe B — a legitimately reclaimable 0-byte destination plus a stale temp file:**

```
PROBE-B err = E-SHD-009: refusing to overwrite an already-sealed shard at '.../burst-log.0001.md' ...
PROBE-B destination still exists = false
```

This one is the real defect. The 0-byte reclaim path ran against the *wrong* path's collision: it
lstat'd the destination, passed the `reclaim_identity_still_safe` re-check, **unlinked the
destination**, retried `write_exclusive`, hit the same stale temp file again, and failed. Net
effect: a failed operation that nonetheless deleted a file, contradicting `publish_sealed_shard`'s
own documented contract ("refuse loud, leave it byte-identical and untouched").

**How the temp file gets there without an attacker.** `write_exclusive` leaks it on any failure
after creation:

```rust
{
    let mut file = OpenOptions::new().write(true).create_new(true).open(&tmp_path)?;
    file.write_all(content)?;   // <-- `?` returns here; cleanup below is never reached
    file.sync_all()?;           // <-- same
}
let link_result = std::fs::hard_link(&tmp_path, path);
let _ = std::fs::remove_file(&tmp_path);   // only reached if the block above succeeded
```

A single `ENOSPC` or `EIO` during `write_all` leaves the temp file behind permanently. With an
attacker in the picture, the same doc comment already assumes a co-resident process that can read
the pid and write into the shard directory — that process does not need a symlink at all; a plain
regular file at `.{basename}.tmp-{pid}` yields a permanent, falsely-attributed roll refusal.

**Suggested fix** — either is sufficient, and the two compose well:

1. Make the temp path non-deterministic (pid **plus** a random nonce). With `O_EXCL` already in
   place this eliminates both the collision and the conflation, and matches the standard
   `tempfile`-style discipline. The crate already depends on `tempfile` for tests.
2. Have `write_exclusive` distinguish the two failures so the caller cannot misroute them, e.g.:

```rust
enum ExclusiveWriteError {
    TempPathOccupied(io::Error),   // never a destination collision
    DestinationExists,             // the real write-once signal
    Io(io::Error),
}
```

and additionally scope-guard the temp file so it is removed on every early return, not only on the
success path.

---

## 3. [MINOR] The E-SHD-010 symlink guard misses the Edit/MultiEdit arm

**File:** `crates/factory-dispatcher/src/shard_manager.rs`, `shard_cap_gate_check`.

The PR body says the FIX-MED-2 guard was applied "at all 4 sites," and I confirmed four
`reject_canonical_symlink` call sites (Write arm ~L1645, `execute_roll` ~L2731,
`self_heal_resume_from_truncate` ~L3061, `reconcile_post_write_replace_all_overcap` ~L3325).

But the **Edit arm** calls `current_shard_bytes_flat(target_path)` directly (~L1678) with no
preceding guard, and `current_shard_bytes_flat` uses `std::fs::metadata` — which follows symlinks:

```rust
pub fn current_shard_bytes_flat(shard_path: &Path) -> io::Result<u64> {
    match std::fs::metadata(shard_path) {   // stat(2), dereferences symlinks
```

The Write arm was given the guard immediately before its own `current_shard_bytes_flat` call; the
Edit arm was not. `read_changelog_item_count`'s `File::open(target_path)` (~L1244) is likewise
unguarded and reads content, not just metadata.

**Impact is bounded** — this is a stat, and if the trigger then fires, `execute_roll`'s own guard
refuses before anything is sealed, so the exfiltration channel FIX-MED-2 targets stays closed. What
leaks is a size oracle on the symlink target. I am raising it as MINOR rather than dropping it
because it is exactly the missed-sibling-callsite pattern TD-VSDD-060 exists to catch, and because
the asymmetry between the two arms will read as an oversight to the next person in this file.

**Suggestion:** hoist `reject_canonical_symlink` above the `match tool_kind` so all three arms
inherit it, rather than repeating it per-arm.

---

## 4. [MINOR] `reclaim_identity_still_safe` can block indefinitely on a FIFO

**File:** `crates/factory-dispatcher/src/shard_manager.rs` (~L2633).

```rust
fn reclaim_identity_still_safe(path: &Path) -> bool {
    std::fs::File::open(path)
```

The function's doc comment carefully reasons about symlink substitution in the race window, but not
about non-regular files. `open(2)` on a FIFO with no writer **blocks forever**. Reaching it requires
swapping a FIFO in between the `symlink_metadata` probe and this call, so the window is narrow — but
the consequence is a hung `PreToolUse` gate, which stalls the whole Claude Code dispatch rather than
failing it.

**Suggestion:** on Unix, open with `O_NONBLOCK | O_NOFOLLOW` via
`std::os::unix::fs::OpenOptionsExt::custom_flags` — that closes both this hang and the residual
symlink case the doc comment currently documents as unclosable without a new dependency. No new
crate needed; `std::os::unix` is enough.

---

## 5. [MINOR] The PR body cites a commit that is not in this PR

The "Deferred to F6 targeted-hardening" section says Item 1 is narrowed by *"this PR's own
`FIX-MED-1` commit (`6cccf3a3`)"*. That commit is **not an ancestor of HEAD**:

```
6cccf3a3 in-HEAD: NO (dangling/orphaned)   parent f874cd1e
0ea79c2c in-HEAD: YES                      parent 0f56530d
```

Both have the identical subject line, so `6cccf3a3` is an orphaned artifact of a discarded branch
state. The Security Review table above it cites the correct SHA (`0ea79c2c`). Given this repo treats
SHA drift as a first-class blocker class, the deferrals section should cite `0ea79c2c` too.

---

## 6. [MINOR] Demo README describes pre-security-fix behavior

`docs/demo-evidence/S-25.02/cluster-2-roll/README.md`, the EC-025 row, says the reclaim
*"`stat()`s once, unlinks the 0-byte file, retries `write_exclusive` exactly ONCE."*

As of SEC-001 the probe is `lstat` (`symlink_metadata`, explicitly non-dereferencing), and as of
FIX-MED-1 there is an additional open-handle identity re-check before the unlink. The recording
itself is still valid evidence — the test it runs is unchanged and still passes — but the prose
describes the pre-fix mechanism. Worth a one-line update so the demo README does not contradict the
security-fix rationale sitting in the same PR.

---

## 7. [MINOR] `E-SHD-010` ships without a taxonomy entry or a deferral anchor

The PR body acknowledges this and routes it to product-owner, which is correct per the Agent Routing
Table. The gap is that it is recorded as *"flagged for a follow-up documentary pass"* with **no
story or wave anchor**. This repo's Canonical Principle Rule 3 requires a deferral to name the
specific future story where it will be resolved, precisely so it cannot get lost. Please attach a
real story ID (or have product-owner add the entry in-scope, which is a single table row).

---

## 8. [NIT] `next_seal_seq` arithmetic can overflow

```rust
Some(index) => Ok(index.shards.iter().map(|s| s.seq).max().unwrap_or(0) + 1),
```

`seq: u32` at `u32::MAX` panics in debug / wraps in release. Practically unreachable, but
`checked_add(1)` returning a fail-loud error costs one line and matches the fail-loud discipline
everywhere else in this module.

---

## 9. [NIT] Diff size

8,847 additions / 2,253 deletions is far over the 500-line review guideline. The breakdown is
mitigating and worth stating: ~1,693 added lines of production code in `shard_manager.rs` plus ~336
across `invoke.rs` / `main.rs` / `executor.rs`, against ~4,300 lines of new unit tests and 2,108
lines of new integration tests, with the rest being demo binaries. A ~3:1 test-to-production ratio
on a data-integrity code path is the right trade. No action requested — noting it so the size is not
mistaken for scope creep.

---

## What I verified and found clean

Stating this explicitly so the approval surface is not mistaken for a rubber stamp:

- **Diff coherence:** every changed file is within `crates/factory-dispatcher/`,
  `docs/demo-evidence/S-25.02/`, or `Cargo.lock`. No unrelated changes.
- **Commit quality:** all 42 commits use Conventional Commits format, all carry the story ID, and
  there is **no AI attribution** in any commit body (this repo forbids it).
- **Forbidden patterns:** no `println!`/`eprintln!`/`dbg!` added anywhere in production code; no
  `unwrap()`, `expect()`, `panic!`, `todo!`, or `unimplemented!` remaining in the non-test half of
  `shard_manager.rs` (the only matches are inside comments).
- **Dependencies:** the single new dependency, `last-amended-migrate`, is workspace-internal,
  path-pinned, and justified inline against the ADR-051 §8 acyclicity rule. `Cargo.lock` gains
  exactly one line and no new third-party packages.
- **`main.rs` refactor:** `resolve_project_cwd` is a faithful extract-to-function of the pre-existing
  `base_host_ctx.cwd` logic — same env var, same empty-filter, same canonicalize-with-fallback, same
  `current_dir` fallback. The documented `SEC-004 TOCTOU ACCEPTED` rationale survives the move (and is
  now referenced from two places rather than one). No behavior change for the other hook plugins.
- **Path resolution:** `invoke.rs` uses raw `file_path` for the target while joining `cwd` for the
  config. I checked this against `executor.rs`, which does the same, and against the in-repo
  invariant comment that the harness always emits absolute `file_path` values. Consistent, not a bug.
- **Catch-point (i) error disposition:** `reconcile_replace_all_overcap_if_qualifying` only
  `tracing::warn!`s on failure. That is correct rather than a swallowed error — it is a PostToolUse
  janitor that cannot block, and catch point (ii)'s leading-probe backstop *does* fail loud on the
  next dispatch, so the condition is recoverable rather than lost.
- **Demo evidence:** 5 `.gif` + 5 `.webm` + 5 `.tape` + README, covering AC-006 (3 clips) and AC-007
  (2 clips), so ≥1 per AC. Both success paths (roll succeeds, 0-byte reclaim succeeds) and error
  paths (E-SHD-009 refusal, Case B1/B2 Block templates) are recorded. I confirmed each `.tape`
  invokes a real `cargo test ... -- --exact --nocapture`, and that all five named tests exist in the
  tree. These are genuine recordings, not `.txt` placeholders.
- **Dependency PR:** cluster-1 (BC-1.18.005, PR #818) is merged to develop, as the PR body claims.

---

## Verdict

**REQUEST_CHANGES.**

- Finding **#2** is the substantive one: a reproducible defect on the write-once seal path that
  reports a factually false `E-SHD-009` and can delete a reclaimable destination during a failed
  operation. It should be fixed before merge.
- Finding **#1** blocks merge mechanically but is not this PR's fault; it needs a `state-manager`
  STATE.md banner reconcile on `factory-artifacts`, independent of this branch.
- Findings #3-#7 are worth closing in-scope per this repo's production-grade default; #8-#9 are
  optional.

Zious11 (commented):
# pr-reviewer addendum — test-coverage pass (sharpens Finding #3)

A second independent review pass (test-coverage focus) confirms and extends **Finding #3** from the
main review. Recording the deltas here:

## Finding #3 escalated in scope — MultiEdit arm also unguarded, and the gap is untested

The main review named the **Edit** arm (`current_shard_bytes_flat` at ~L1678). The **MultiEdit** arm
(~L1748) has the identical unguarded `current_shard_bytes_flat(target_path)` call — so **two** of the
three mutation-tool arms follow a symlinked canonical through `stat()`, while only the Write arm
(~L1645) carries the `reject_canonical_symlink` guard.

Crucially, this path has **zero test coverage**: `test_FIXMED2_..._fails_loud_e_shd_010` (~L7846)
exercises only the Write-arm backstop. There is no Edit or MultiEdit variant. So the guard asymmetry
between the arms is not caught by any existing test and would not regress-fail if left as-is.

- Impact is still bounded exactly as the main review stated (if the leaked size trips the trigger,
  `execute_roll`'s own guard at ~L2731 refuses loud, so no symlink-target bytes are ever sealed) —
  residual is info-exposure + guard asymmetry, not data loss. The coverage gap is what raises this
  from a MINOR nit toward MAJOR: it is precisely the missed-sibling-callsite pattern TD-VSDD-060
  exists to catch, shipped without a failing test to pin it.
- **Route:** implementer adds `reject_canonical_symlink` at ~L1678 and ~L1748 (or hoist it above the
  `match tool_kind` so all three arms inherit it, per the main review's suggestion); test-writer adds
  Edit + MultiEdit variants of the ~L7846 test.

## Additional coverage notes (MINOR / for the record)

- **FIX-MED-1 (TOCTOU re-check) is tested only at helper level.** Both tests (~L7973, ~L8008) call
  `reclaim_identity_still_safe` directly; no test drives the
  probe → concurrent-write → re-check → unlink path through `publish_sealed_shard` itself. The
  integration-level guarantee currently rests on inspection of the callsite (~L2564), not a test.
  (This also intersects Finding #2 in the main review — an integration-level test through
  `publish_sealed_shard` would likely have surfaced the temp-path-collision misattribution.)
- **Debug-format coupling (NIT).** `bc_1_18_006_roll_test.rs` ~L309 and ~L346 assert `.contains(...)`
  against `format!("{:?}", summary.per_plugin_results)` — fragile coupling to `Debug` rendering,
  compensated by the verbatim `exact_block_reason` assertion at ~L383. Acceptable; noted.

## Confirmed well-covered (no rubber-stamp)

The other five security fixes have dedicated, non-tautological tests against the real filesystem and
the named error variants: FIX-HIGH-1 → ~L7909, SEC-001 → ~L6645, SEC-002 → ~L3949/~L3982,
SEC-003 → ~L3809/~L3827, and FIX-MED-2's other four *guarded* sites → ~L7846/~L8026/~L8070/~L8104.
No `#[ignore]`, no `should_panic`, no tautological or over-mocked tests found.

Verdict is unchanged: **REQUEST_CHANGES**.

Zious11:
## Finding #10 (NIT) — diff size acknowledgment

Diff size (8,847 / 2,253 lines) is acknowledged as far over the repo's 500-line PR guideline. No split is warranted: this cluster implements a single indivisible behavioral contract (BC-1.18.006's staged roll-before-write sequence, its write-once guard, and the 0-byte reclaim exception) plus the security hardening layered onto it across two independent review passes. The size is driven by an approximately 3:1 test-to-production ratio (2,108 of the added lines are `bc_1_18_006_roll_test.rs`) on a data-integrity code path, which is the correct trade-off per this repo's production-grade default rather than scope creep. No code change from this finding.

Zious11 (commented):
# PR #824 — Fresh-Eyes Final Review, CYCLE 3

**PR:** `S-25.02 cluster-2: artifact-sharding roll (BC-1.18.006 v1.11)`
**Head reviewed:** `68796531` (`feature/S-25.02-roll` → `develop`, merge-base `fff5e4cc`)
**Diff:** 9,933 additions / 2,122 deletions across 23 files

## VERDICT: REQUEST_CHANGES

The five claimed-fixed items from the prior cycle **all verify as genuinely fixed** — confirmed empirically, including by mutation-testing the code to prove the new tests are load-bearing rather than decorative. That work is solid.

However, this pass independently surfaced **three MAJOR findings** not addressed by the claimed remedies. One is a silent regression of this PR's *own* security fix caused by the composition of two fixes landed in different commits; two are in the `invoke.rs` / `main.rs` wiring, which the prior cycle's findings did not reach.

---

## Verification Performed

| Check | Command / method | Result |
|---|---|---|
| Full crate suite | `cargo test -p factory-dispatcher` | 454 lib + 24 integration + others, **0 failed** |
| Targeted shard tests | `cargo test -p factory-dispatcher --lib shard_manager::` | 161 passed, 0 failed |
| Formatting | `cargo fmt --check --all` | clean (exit 0) |
| Lint | `cargo clippy -p factory-dispatcher --all-targets -- -D warnings` | clean |
| **Mutation A** — reverted the `TempPathOccupied` typed distinction so it falls through to reclaim | rebuild + `cargo test test_N1` | `test_N1_..._temp_path_collision_never_misattributed` **FAILED** with exactly `SealedShardAlreadyExists` — the misattribution the fix exists to prevent |
| **Mutation B** — restored pre-`ef6ca3b4` unlink-then-stage ordering | rebuild + `cargo test test_N1` | `test_N1_..._failed_retry_staging_never_destroys_reclaimable_destination` **FAILED** with `NotFound` — destination destroyed |
| **Mutation C** — replaced ancestor walk with naive `parent().metadata()` | rebuild + `cargo test test_N2` | `test_N2_closest_existing_ancestor_propagates_when_blocked_by_a_file` **FAILED** |
| **Ordering probe** — moved `reclaim_identity_still_safe` to immediately before `remove_file` | full suite | **all pass** → proves no test pins the ordering, and proves the recommended fix is safe |
| Test seam scoping | grep for `test-support` gating, integration-test references | seam is `#[cfg(test)]`-only, not feature-gated, not reachable from `tests/` |
| Windows constant `0x0100` | `test_N4_..._rejects_symlink_via_o_nofollow` on darwin | passes → empirically proves `0x0100` is `O_NOFOLLOW` on **macOS**, contradicting the N-7 comment |
| Demo evidence | `ls` + `file` on `docs/demo-evidence/S-25.02/cluster-2-roll/` | 5 real GIF (1200×700) + 5 real WebM + 5 tapes + README — no `.txt` placeholders |
| Commit hygiene | `git log` scan | 56 commits, all Conventional, all story-scoped, **no `Co-Authored-By`**; `Claude-Session:` trailer matches established `develop` convention |
| CI | `gh pr checks 824` | **NOT green** — see gate note below |

Worktree was restored to pristine `68796531` after every mutation; `git status --porcelain` empty.

---

## Claimed-Fixed Items — Independent Confirmation

**1. N-1 call-site regression coverage for `publish_sealed_shard` — CONFIRMED FIXED.**
Both tests exist in `shard_manager::bc_1_18_006_roll_tests`, both call the real `pub fn publish_sealed_shard` (not `stage_temp_file_with_nonce`), and both are proven load-bearing by Mutations A and B. They assert on the *specific* behavior the fix changed — the error variant *and* its `E-SHD-001` Display text for misattribution, and post-condition on-disk state for destructive ordering — not merely "it doesn't crash".

**2. Windows `ERROR_PATH_NOT_FOUND` over-correction — CONFIRMED FIXED, better than the minimum asked for.**
The fix is not a `parent().metadata()` probe; it is `closest_existing_ancestor_is_directory_or_absent`, walking the full ancestor chain to the first extant node. That correctly handles the multi-level case a bare parent check would misjudge, which Mutation C confirms the suite pins. Genuine traversal-through-a-file still propagates (`test_FINDING1_..._false_for_path_traversal_through_non_directory`, `test_N2_..._propagates_when_blocked_by_a_file`). Compiling the helper under `#[cfg(any(windows, test))]` for portable unit-testing is a good call. See NIT-3 for the residual probe wrinkle.

**3. `FORCE_STAGE_FAILURE` test seam — CONFIRMED LEGITIMATE.**
`#[cfg(test)]` on the `thread_local!`, the enum, the guard, the setter, *and* the branch inside `stage_temp_file`. Not gated on `test-support` (which CI enables on the `build-dispatcher` matrix) — verified `test-support` remains `[]`, non-default, and touches only `VSDD_ASYNC_DRAIN_WINDOW_MS`. Not referenced from `tests/bc_1_18_006_roll_test.rs`. Pure early-return that fires only when a test arms it, with an RAII guard clearing state on drop including on unwind. Does not weaken the real path.

**4. Doc corrections (`O_NOFOLLOW` contradiction, FIFO `O_NONBLOCK` mechanism) — SUBSTANTIALLY FIXED.**
The N-5 correction is technically right and non-obvious: `O_RDONLY | O_NONBLOCK` on a writerless FIFO *succeeds* per POSIX (only `O_WRONLY` fails with `ENXIO`), and rejection genuinely comes from `is_file()` being false. The "must not remove `O_NONBLOCK` believing it is load-bearing for the open FAILING" note is good defensive documentation. See NIT-1 / NIT-2 for two statements not corrected.

**5. README + `publish_sealed_shard` doc vs. actual stage-before-unlink order — CONFIRMED ACCURATE.**
README line 46 and the `publish_sealed_shard` header both describe stage → unlink → publish, matching the code. (But see MAJOR-1: a *different* claim in those same paragraphs is now stale.)

**6. fmt / clippy / commit hygiene — CLEAN.**

---

## Findings

### MAJOR-1 — FIX-MED-1's TOCTOU re-check is no longer "immediately before the unlink"; Finding #2's fix silently widened the window it exists to close

`crates/factory-dispatcher/src/shard_manager.rs`, `publish_sealed_shard`

Actual execution order in the 0-byte-reclaim branch:

1. `reclaim_identity_still_safe(sealed_path)` — the FIX-MED-1 re-check
2. `stage_temp_file(sealed_path, content)` — **create + `write_all` + `sync_all`**
3. `std::fs::remove_file(sealed_path)` — the unlink
4. `publish_staged_temp_file(...)`

FIX-MED-1 (`0ea79c2c`) was introduced specifically so the identity re-check sits *adjacent* to the unlink. Finding #2's fix (`ef6ca3b4`) then inserted a full staged write — **including an `fsync`** — between them. The CWE-367 window FIX-MED-1 narrowed from "two adjacent syscalls" is now "one durable write wide". Concrete consequence: the exact data-loss scenario FIX-MED-1 was written to prevent — a legitimate concurrent writer replacing the 0-byte placeholder with real sealed content during step 2 has its content silently unlinked at step 3.

Four sites still assert the property that no longer holds:

- `shard_manager.rs:2989` — `// re-verify identity IMMEDIATELY BEFORE unlinking`
- `shard_manager.rs:3065-3068` — `reclaim_identity_still_safe`'s doc: *"called by `publish_sealed_shard`'s 0-byte reclaim path immediately before the unlink… closing the window as tightly as `std::fs` allows"*
- `shard_manager.rs:8994` — the Finding #8 test doc: *"which calls `reclaim_identity_still_safe` immediately before its `remove_file` unlink"*
- `docs/demo-evidence/S-25.02/cluster-2-roll/README.md:46` — *"IMMEDIATELY BEFORE unlinking"*

The PR body likewise still says FIX-MED-1 *"clos[es] most of the window"*.

No test catches this. The Finding #8 FIFO test gates on the re-check *existing*, not on where it sits, so it passes under either ordering. Moving the re-check to immediately before `remove_file` and running the full suite yields **all 454+ tests passing** — confirming both that nothing pins the current (worse) order and that the fix is safe.

**Suggested fix** — preserves *both* fixes fully (staging still precedes the unlink; the re-check becomes genuinely adjacent to it):

```rust
    let staged_tmp_path = match stage_temp_file(sealed_path, content) { /* ... unchanged ... */ };

    // FIX-MED-1: re-verify identity IMMEDIATELY before the unlink — after
    // staging, so no durable write widens the window between the two.
    if !reclaim_identity_still_safe(sealed_path) {
        let _ = std::fs::remove_file(&staged_tmp_path);
        return Err(already_exists_err());
    }

    if std::fs::remove_file(sealed_path).is_err() { /* ... unchanged ... */ }
```

Note the added `remove_file(&staged_tmp_path)` cleanup on the new early-return — without it the staged temp file leaks, the same leak class `stage_temp_file` already guards internally. Also add a test pinning the ordering (e.g. arm `FORCE_STAGE_FAILURE` for call #1 against a FIFO destination and assert the re-check's abort, not the staging failure, is what surfaces), and correct all four doc sites.

### MAJOR-2 — New catch-point-(i) leg reaches the destructive `execute_roll` without `validate_entry`, bypassing EC-022

`crates/factory-dispatcher/src/invoke.rs:2413-2431` (new in this PR)

`validate_entry` has exactly **one** production call site in the crate: `shard_manager.rs:1634`, inside `shard_cap_gate_check` (every other hit is a test or comment). The new path — `reconcile_replace_all_overcap_if_qualifying` → `find_matching_entry` → `reconcile_post_write_replace_all_overcap` → `execute_roll(entry, canonical_path, true)` at `shard_manager.rs:3899` — never validates the matched entry. `find_matching_entry` only compares `artifact_stem` against `file_stem()` and calls `path_falls_under_or_equals`; it validates nothing.

This matters because `execute_roll` is *destructive*: it seals and then truncates the canonical to 0 bytes. An entry with `artifact_path = "."` or `"./"` — which `validate_entry` rejects under EC-022, with dedicated tests — degenerates to an empty registered-component vector in `path_falls_under_or_equals`, whose suffix test then matches **any** path whose `file_stem()` equals `artifact_stem`. `shard_manager.rs:~4569` says so in its own words: *"This is precisely the raw behavior EC-022's `validate_entry` check exists to make unreachable via config."* On the PreToolUse leg such an entry is a loud `HookResult::Error`; on this new PostToolUse leg it silently seals-and-empties a file the config never legitimately governed. The same bypass covers the EC-010/011/013/015/017 cap-sanity checks — `reconcile_post_write_replace_all_overcap` compares `actual_size <= entry.shard_cap_bytes` using a cap the gate itself would refuse to operate under.

**Suggested fix:** call `validate_entry(&entry)` in `reconcile_replace_all_overcap_if_qualifying` immediately after `find_matching_entry` returns; on `Err`, emit a `tracing::warn!` and return `None` — consistent with this leg's documented fail-open-but-never-silent contract (Decision 15 point 2) and with the F-C2-P1-005 warn 25 lines above.

### MAJOR-3 — `main.rs` placement rationale asserts a PreToolUse guarantee that does not exist; both cited precedents are also wrong

`crates/factory-dispatcher/src/main.rs:291-311`

The new 30-line comment justifies placing the catch-point-(i) call before the `sync_tiers.is_empty() && partition.async_group.is_empty()` early return by invoking *"the exact 'silently stop firing if the plugin set changes' failure mode Decision 1's placement rule already exists to prevent **for the PreToolUse leg**."*

Traced: `shard_cap_precheck` is called at `executor.rs:532`, **inside** `execute_tiers`; `execute_tiers` is invoked at `main.rs:501` — *after* the early return at `main.rs:322`. So the PreToolUse byte-cap gate does exactly what the comment claims is already prevented: it silently stops firing on any configuration where no plugin matches. Both cited precedents fail the same way — `inject_git_context_if_qualifying` is at `main.rs:420` (behind the guard), and `write_indeterminate_marker` lives inside `execute_tiers` (per-plugin). The new call is in fact the *only* occupant of that slot.

Not merely a stale comment: it tells a future maintainer a safety property holds when it does not, and obscures a real asymmetry between the two legs of the same BC. Per CLAUDE.md's own rule (doc comment claiming a capability with no capability check → either implement the gate or remove the docs), either hoist the PreToolUse gate to match, or correct the comment (and its twin at `invoke.rs:2348-2352`) to state the asymmetry honestly and route the hoist decision to the architect.

### MINOR-1 — `find_matching_entry`'s `Err(DuplicateArtifactStem)` silently swallowed

`invoke.rs:2413-2416`: `find_matching_entry(&registry, &target_path).ok().flatten()?`. That variant is documented as a normal, fail-loud return callers MUST handle ("NOT a `panic!`-only-on-bug escape hatch"), and the PreToolUse path converts it to `HookResult::Error`. Here an ambiguous `[[shard]]` config produces zero telemetry — directly inconsistent with the F-C2-P1-005 fix 25 lines above, which added a `tracing::warn!` to the registry-load failure with the explicit rationale *"fail-open, never propagated as an error to the caller — but never silent."* Add the matching warn.

### MINOR-2 — Missing / non-string `tool_input.file_path` silently yields `None`

`invoke.rs:2408-2412`. The sibling PreToolUse path (`executor.rs:398-411`) treats the identical condition as fail-loud, with the stated rationale *"MUST fail loud, never silently resolve to an empty `PathBuf`."* Same "never silent" argument; a warn is the minimum.

### MINOR-3 — `main.rs:321` "UNCONDITIONAL native call, independent of the registry" is false

Two registry paths return before it: `resolve_registry_path()?` at `main.rs:152`, and the fail-open `return Ok(exit_code)` on `Registry::load` failure (~`main.rs:238`). A broken or unparseable `hooks-registry.toml` therefore silently disables catch point (i) — an over-cap canonical left unreconciled with no telemetry — while the doc asserts registry independence.

### MINOR-4 — "exercised end-to-end by `bc_1_18_006_roll_test.rs`'s AC-024 integration test" is not true of this call site

`main.rs:319-320`. That test calls `reconcile_replace_all_overcap_if_qualifying(&payload, dir.path())` directly (`tests/bc_1_18_006_roll_test.rs:603`, 1365, 1550); it never goes through `main::run`. No bats coverage either. The wiring itself — the placement, and the fact that the *canonicalized* `project_cwd` rather than raw `$CLAUDE_PROJECT_DIR` is joined with `.factory/shard-config.toml` — is untested. Either add coverage or correct the claim.

### MINOR-5 — Ordering side effect on PostToolUse plugins not considered

Because the catch-point-(i) call precedes the tier loop, PostToolUse WASM plugins registered for `Edit`/`MultiEdit` now execute against a canonical that may already have been truncated to 0 bytes. A PostToolUse validator that reads the artifact it was invoked for would see an empty file instead of the content just written — a plausible false-positive advisory/block source on precisely the dispatch where the roll fires. The 30-line placement rationale does not mention this trade-off; it should, and the risk should be assessed.

### MINOR-6 — New dependency's justification comment is factually wrong

`crates/factory-dispatcher/Cargo.toml:59-66` claims steps (b)-(d) *"all reuse `write_atomic`'s existing temp-file-then-rename primitive verbatim… no new atomic-write primitive is introduced."* Steps (c) and (d) do. Step (b) does **not**: `publish_sealed_shard` uses a purpose-built `write_exclusive` / `stage_temp_file` + `publish_staged_temp_file` pair over `OpenOptions::create_new` + `hard_link` + a local `random_nonce()` — deliberately *not* `rename(2)`, because rename clobbers and this path must be write-once. The dependency is still justified by (c)/(d); the stated scope is not. (The dependency itself is fine: workspace-internal path dep, `Cargo.lock` gains exactly one line, no new external crates, genuinely used in non-test code.)

### MINOR-7 — PR description is stale against the head being reviewed

The body twice cites `27d2f5ec` as "current head" and cites `39369cc6` for the "3072 tests / 0 failed" evidence; actual head is `68796531`, eight commits later. The cycle-2 tracking pointer is also hedged (*"…or a sibling `pr-review-cycle2.md`"*). Refresh before merge.

### NIT-1 — The N-7 constants comment is inverted and self-contradictory

`shard_manager.rs:3136-3137`: *"(`0x0100` is `O_NOFOLLOW` on Linux/Android but `O_NOCTTY` on Linux — the wrong flag entirely)"*. Backwards *and* internally contradictory (names Linux on both sides). The truth is the reverse: `0x0100` is `O_NOFOLLOW` on **macOS/BSD**, and `O_NOCTTY` on **Linux/Android** (Linux `O_NOFOLLOW` is `0o400000`). The **code is correct** — proven empirically: `test_N4_reclaim_identity_still_safe_rejects_symlink_via_o_nofollow` passes on a macOS host, which it could only do if `0x0100` really is `O_NOFOLLOW` on darwin. Flagged specifically because this comment is the entire payload of commit `e4c5d332`, whose stated purpose was getting these constants right — the classic "reworded, still wrong" shape.

### NIT-2 — `reclaim_identity_still_safe` doc overstates what it accepts

`shard_manager.rs:3094-3095`: *"Returns `false`… for a symlink whose target is non-empty."* On Unix, `O_NOFOLLOW` rejects **every** symlink regardless of target — as `test_N4` (which uses a symlink to a genuinely 0-byte target) demonstrates. The phrasing implies a symlink-to-empty-file would be accepted, true only of the `#[cfg(not(unix))]` fallback. Suggest: *"on Unix, for any symlink at all (`O_NOFOLLOW`); on other platforms, for a symlink whose target is non-empty."*

### NIT-3 — Ancestor probe conflates "absent" with "unstattable"

`closest_existing_ancestor_is_directory_or_absent`'s `Err(_) => cur = candidate.parent()` treats a permission or I/O failure identically to non-existence, walking further up until it finds a directory and relieving. Narrow in practice — an `EACCES` ancestor surfaces as `ERROR_ACCESS_DENIED` (5), which never reaches this helper via `is_genuinely_missing`'s `_ => false` arm — so no behavior change requested, only that the doc say the probe treats any inconclusive `metadata` as "absent".

### NIT-4 — `invoke.rs:2352-2354` reintroduces a retracted framing, and orders the discriminators backwards

*"Every check here is cheap and structural"* — `ShardRegistry::load` is a filesystem read plus a full TOML parse per qualifying dispatch. `find_matching_entry`'s own doc already retracts exactly this framing for the PreToolUse leg ("PR #818 fix-burst finding B4"). Relatedly, `shard_config_path.exists()` + `ShardRegistry::load` both run *before* `file_path` is extracted; extracting `file_path` first is free and skips the parse on a malformed payload.

### NIT-5 — Dependency direction

`last-amended-migrate` is a `clap`-based operator CLI crate; taking it as a normal (non-dev) dependency of the shipped dispatcher pulls `clap` into the dispatcher's build graph for one ~40-line `write_atomic`. Lock/CI impact is nil (`clap` is already a workspace dep, dead code is linker-dropped), but hoisting `atomic_write` into a leaf crate both could depend on would be the cleaner edge.

---

## Gate note — CI is not green at `68796531`

At review time: `validate`, `semgrep`, `deny-advisories`, `platforms-drift`, `policy-15`, `attestation-gate`, and `build-dispatcher (linux-arm64)` pass; **`cargo-host` (both runners), `bats-full-suite`, `bats-wave-handoff`, and `build-dispatcher` (darwin-arm64, darwin-x64, linux-x64, windows-x64) are all still pending.**

Specifically: `build-dispatcher (windows-x64)` is the **only** job in the matrix that executes the `#[cfg(windows)]` tests — including `test_N2_is_genuinely_missing_windows_path_not_found_relieved_for_missing_parent_dir`, this cycle's headline fix. `cargo-host` runs only ubuntu + macos. So the Windows arm of `is_genuinely_missing` has **zero executed coverage until that job goes green**. It must be green before merge, not merely "expected to pass".

---

## Checklist Summary

| # | Item | Result |
|---|---|---|
| 1 | Diff coherence | PASS — all changes scoped to `factory-dispatcher` + this story's demo evidence |
| 2 | Description accuracy | MINOR-7 (stale head SHAs) |
| 3 | Test coverage | Strong on `shard_manager.rs` (mutation-verified); gaps at MAJOR-1 (ordering), MINOR-4 (`main.rs` wiring) |
| 4 | Demo evidence | PASS — 5 real GIF+WebM, 1+ per AC, success and error paths, README accurate |
| 5 | Commit quality | PASS — Conventional, story-scoped, no AI co-author attribution |
| 6 | Diff size | 9,933 lines, far over the 500 guideline; ~3:1 test-to-production on a data-integrity path — prior disposition not to split is reasonable and concurred |
| 7 | Missing changes | MAJOR-2 (`validate_entry` not wired into the new leg) |
| 8 | Dependency status | PASS — cluster-1 (PR #818) merged; no new external crates |

